PR #31, opened 2021-01-21
display version number during root and secondary server start

kalluriramkumar:


murali-shris:
you have to add version to pubspec ?

murali-shris (approved):

PR #67, opened 2021-03-04
Cache public keys

kalluriramkumar:


at_secondary/at_secondary_server/lib/src/connection/inbound/dummy_inbound_connection.dart
```diff
@@ -22,12 +23,15 @@ class DummyInboundConnection implements InboundConnection {
...
+    if (connection is DummyInboundConnection) {
+      return true;
+    }
+    return false;
   }
 
   @override
   AtConnectionMetaData getMetaData() {
-    return null;
+    return InboundConnectionMetadata()..fromAtSign = null;
```
murali-shris, line 34: InboundConnectionMetadata() should be created once for one dummy inbound connection.in this case what will happen is, on every getMetaData() new InboundConnectionMetadata will be created

kalluriramkumar: Removed the getMetadata() method, the fromAtSign is being set by 'from:@atsign' command in fromVerbHandler. Though we set fromAtSign to null here, it is being set to atsign in verb handler. Hence reverted the change. Verified the refresh job and is working fine.

murali-shris (approved):

PR #93, opened 2021-04-19
Stream sync response

sitaram-kalluri:


at_secondary/at_secondary_server/test/sync_verb_test.dart
```diff
@@ -81,114 +76,7 @@ void main() async {
       expect(paramsMap['regex'], 'me');
     });
   });
-
-  group('A group of sync verb handler tests', () {
```
murali-shris, line 85: why are these tests commented ?

kalluriramkumar: Earlier, sync response is added to the Response object. Now that we are directly writing to the socket, removed the test cases.

VJag (approved):

VJag (approved):

cpswan (approved):
LGTM, though it's not entirely clear to me how the tests cover this additional functionality. Do we need a performance test to show how much faster things are after this change?

PR #121, opened 2021-05-12
template based config parameters

sitaram-kalluri:
**- What I did** Template based config parameters

**- How I did it** generate_config.yaml will generate the config.yaml for given environments. defaults to production.

**- How to verify it** Run the generate_config.yaml and a config.yaml should be generated with all the configurations.

**- Description for the changelog** Override the default configurations in config-base.yaml for different environments.


cpswan (commented):
I'm not sure we need the linter in the pipeline for a single script.

When I asked about linting this morning it was more of a one off thing.

This is probably an example of something that seems like a good idea now, but that will slow down the commit cycle from now until eternity, with little gain after the initial run.

The rest LGTM

cpswan (approved):
LGTM :)

at_secondary/at_secondary_server/config/generate_config.py
```diff
@@ -0,0 +1,49 @@
...
+import yaml
+from jproperties import Properties
+
+# Initialize parser
+parser = argparse.ArgumentParser()
+# Adding optional argument
+parser.add_argument("-e", "--env", default='production')
+
+# Read arguments from command line
+environment = parser.parse_args().env
```
VJag, line 22: It would help to have some console.writes on for what environment we are building the config

at_secondary/at_secondary_server/config/generate_config.py
```diff
@@ -0,0 +1,49 @@
...
+        fields = key.split('.')
+        temp = yamlMap
+        i = 0
+        # Iterate into the map until the key is fetched.
+        while i < len(fields):
+            # Condition is to check if the key in the map is fetched.
+            if i == len(fields) - 1:
+                temp[fields[i]] = configs.properties[key]
+                break
+            temp = yamlMap[fields[i]]
```
VJag: I did not get this. Can you please explain what are we doing here for my understanding?

VJag (commented):
Hi Sita. Please check my comments.

at_secondary/at_secondary_server/config/generate_config.py
sitaram-kalluri: Sure Jagan, Will add a print statement.

at_secondary/at_secondary_server/config/generate_config.py
sitaram-kalluri: For a property in configuration, drill down to a specific key in the map and assign the value to the key.
For example:
For root_server.url, loop until url in the map and assign the root server url.

VJag (approved):

PR #132, opened 2021-05-24
Change dart2native to dart compile

### LINKED ISSUE #131: Replace `dart2native` with `dart compile`
**Is your feature request related to a problem? Please describe.**
`dart2native` is scheduled for deprecation in Dart 2.14, and will be discontinued in Dart 2.15 (per [Deprecation and discontinuation of standalone Dart tools](https://github.com/dart-lang/sdk/issues/46100))

**Describe the solution you'd like**
Replace `dart2native` with `dart compile`

**Additional context**
We presently use `dart2native` in two workflows:
* https://github.com/atsign-foundation/at_server/blob/7091ed1ddb8c4d3bf9c689743b550620e8bd0d7d/.github/workflows/at_server_trunk.yaml
* https://github.com/atsign-foundation/at_server/blob/9addc8a3ef54512b0b257088f11044a6e0b4da70/.github/workflows/at_server_release.yaml

sitaram-kalluri:
**- What I did** Change dart2native to dart compile

**- How I did it** Change dart2native in GitHub actions to dart compile

**- How to verify it** Github actions should build root and secondary binaries successfully.

### DIFF
--- a/.github/workflows/at_server_trunk.yaml
+++ b/.github/workflows/at_server_trunk.yaml
@@ -223,7 +223,7 @@ jobs:
         with:
           push: true
           context: at_functional_test/lib/
-          tags: atsigncompany/virtualenv:GHA${{ github.run_number }}
+          tags: atsigncompany/virtualenv:dev_env
 
       - name: Image digest of at_virtual_env
         run: echo ${{ steps.docker_build_trunk.outputs.digest }}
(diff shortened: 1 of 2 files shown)


.github/workflows/at_server_trunk.yaml
```diff
@@ -223,13 +223,13 @@ jobs:
         with:
           push: true
           context: at_functional_test/lib/
-          tags: atsigncompany/virtualenv:GHA${{ github.run_number }}
```
cpswan, line 226: I'd like to keep automation tags, though just GHAx maybe isn't sufficient as we have different action types that will clock through their own increments, and we don't want them stepping on each other

cpswan (approved):
One suggestion to keep the automation tags, otherwise LGTM

PR #139, opened 2021-05-27
Monitor connection does not timeout

sitaram-kalluri:
**- What I did** Make monitor connection stay alive on server side

**- How I did it** When isMonitor is set to true on connection, ignore connection timeout.

**- How to verify it** When the connection is monitor, the connection should stay alive even after the connection exceeds the time out


cpswan (approved):
There's a lot of dependency override stuff riding in on this PR, and I would have preferred to have seen it split into 2 PRs, but I appreciate that sometimes it can be difficult to get a clean sequence of commits.

PR #155, opened 2021-06-01
Fix server start time

chkarthik1729:
**- What I did**
The server, on startup will require SSL certificates to start a SecureSocket. It will retry for certificates if not found for every 10 seconds (sleeps for 10 seconds) for retryCount times.

I have fixed an issue where the server goes to sleep once even if the certificates are available.

**- How I did it**
Moved the break condition in the retry loop to appropriate place.

**- How to verify it**
You can start the server, and check the startup time. It will not take more than 5 seconds now.


murali-shris:
@kalluriramkumar please check this PR

sitaram-kalluri (approved):

PR #177, opened 2021-06-08
Increment root and secondary versions in pubspec.yaml

sitaram-kalluri:
**- What I did** Increment version in pubspec.yaml in root and secondary servers

**- How I did it**

**- How to verify it**

### DIFF
--- a/at_root/at_root_server/pubspec.yaml
+++ b/at_root/at_root_server/pubspec.yaml
@@ -1,6 +1,6 @@
 name: at_root_server
 description: Root server implementation.
-version: 1.0.0+1
+version: 1.0.0
 repository: https://github.com/atsign-foundation/at_server
 homepage: https://atsign.dev
 publish_to: none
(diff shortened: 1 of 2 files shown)


at_root/at_root_server/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: at_root_server
 description: Root server implementation.
-version: 1.0.0+1
+version: 1.0.0
```
cpswan: If we were at 1.0.0+1 then why are we going back to 1.0.0?

sitaram-kalluri: Can i have it has 1.0.1?

cpswan: Yes - that would be fine, and also makes sense if we're going to have to release the whole of this repo at 1.0.1

cpswan (approved):

PR #245, opened 2021-07-23
Functional tests

purnimavenkatasubbu:
**- What I did**
* Added functional tests for all the verbs
* Also added end-to-end tests for all the verbs

* Functional tests for the verbs
* End-to-end tests for the verbs
* [gkc]
  * Made update_verb_test faster by reusing the socket handlers across tests, thus saving on sending 'from:' and 'pkam:' each time. Same approach can be applied to other tests
  * Removed a "logger.severe" from the main catch bracket of AtSecondaryServerImpl and moved the responsibility for determining what level to log at to the global exception handler. Accordingly, modified global exception handler to have different log levels for different exceptions - INFO, WARNING, SEVERE or SHOUT. There is no change in this commit to the behaviour with regards to closing connections for various exceptions


cpswan:
@purnimavenkatasubbu thanks for rebasing this. It looks like a bunch of the functional tests are failing for various reasons, so there's still some polish needed before these can be merged.

cpswan:
@purnimavenkatasubbu can you please fix the merge conflicts on this (@kalluriramkumar please help if needed)

@gkc added you as a reviewer. I'd like to get this merged, as there's a lot of work here that's taken many months of effort.

gkc (approved):
Looks good! Will be great to have these in our cicd pipeline

cpswan:
@purnimavenkatasubbu it looks like 42662aa has a corrupted fullchain.pem in it that's got merge conflict dividers still present, which is stopping the virtual environment from coming up for functional tests.

I'll grab the latest certs from at_virtual_environment and try to fix it.

cpswan:
@purnimavenkatasubbu please take a look at [test run #464](https://github.com/atsign-foundation/at_server/runs/4381202415?check_suite_focus=true) as it seems a bunch of the functional tests are failing inside of GitHub Actions.

We may also need to do some refactoring (to run tests in parallel) if tests are going to take significantly longer than before, as I'd like to keep the commit loop where it was ~4m. But for now working tests are the priority, we can work on speedy later.

cpswan:
@purnimavenkatasubbu lots of `NoSuchMethodError: The method 'trim' was called on null.`, which I think we seen before when dependencies have been tangled.

I also see a bunch of `@cicd1` stuff in logs. Is there stuff that's supposed to be in the end to end tests that's found its way back to the at_functional_tests directory? I fear that this might be introducing a dependency on the cicd host VMs that shouldn't be there for functional tests.

gkc (approved):

sitaram-kalluri (approved):

PR #246, opened 2021-07-23
Add dess armv7 build to at_server workflow

### LINKED ISSUE #168: automation of multi architecture build of secondary for arm7

cpswan:
Closes #168 and progresses [dess #10](https://github.com/atsign-foundation/dess/issues/10)

**- What I did**

Modified push_prod_secondary_image step to generate platform specific labels for x64 and arm64, build an arm version on a Pi, then bring the three versions together into a multi-arch manifest

**- How I did it**

x64 and manifest generation takes place within GitHub Actions.
arm64 uses the cicd1 VM
arm uses a Mythic Beasts Pi

**- How to verify it**

Confirm that the atsigncompany/secondary:dess image contains digests for all three OS/ARCH e.g.

image

Add dess armv7 build to at_server workflow


xavierchanth:
Looks good to me, I will verify from the pi

cpswan:
@kalluriramkumar once again the end2end tests are failing. It turns out that this is [by design to prevent leakage of secrets](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/).

It shouldn't be a problem for PRs from branches. We should probably add conditionals so that the end2end tests don't run on PRs from forks.

xavierchanth:
Built the armv7 image on a pi using getdess with no issues.

sitaram-kalluri (approved):

PR #271, opened 2021-08-13
Observable secondary Dockerfile

cpswan:
**- What I did**

Added Dockerfile to create an observable secondary.

**- How I did it**

* Single stage build based on our usual build image with the Dart SDK
* No need to compile an AOT binary
* Fixes permissions so that non root user can run dart and access root .pub-cache
* Entrypoint is running:

`/usr/lib/dart/bin/dart run --observe=8181/0.0.0.0 /app/at_secondary_server/bin/main.dart`

NB binding to 0.0.0.0 so that Docker compose can map ports into it

**- How to verify it**

Build the image:

```bash
sudo docker build . -f Dockerfile.observe -t atsigncompany/secondary_obs
```

Modify a dess docker-compose.yaml like this:

```yaml
image: atsigncompany/secondary_obs
    networks:
      second: {}
    ports:
    - published: 6767
      target: 6767
    - published: 8181
      target: 8181
```

Then (re)create a service from that docker-compose.yaml

The start of the service logs will contain the authentication URL for the observatory.

Observable secondary Dockerfile

### DIFF
--- a/at_secondary/Dockerfile.observe
+++ b/at_secondary/Dockerfile.observe
@@ -1,0 +1,28 @@
+FROM atsigncompany/buildimage
+ENV HOMEDIR=/atsign
+ENV USER_ID=1024
+ENV GROUP_ID=1024
+WORKDIR /app
+COPY . .
+RUN \
+  cd at_persistence_secondary_server ; \
+  dart pub get ; \
+  dart pub update ; \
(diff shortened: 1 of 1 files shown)


cconstab:
Taking a look today thanks Chris

PR #365, opened 2021-10-26
Add latestCommitId to at_commit_log.dart

sitaram-kalluri:
**- What I did** Return latest commitId.

**- How I did it**

**- How to verify it**

### DIFF
--- a/at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
+++ b/at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
@@ -16,6 +16,7 @@ class CommitLogKeyStore implements LogKeyStore<int, CommitEntry?> {
   final _currentAtSign;
   late String _boxName;
   final _commitLogCacheMap = <String, CommitEntry>{};
+  int latestCommitId = -1;
 
   CommitLogKeyStore(this._currentAtSign);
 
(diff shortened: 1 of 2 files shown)


at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
```diff
@@ -71,6 +76,10 @@ class CommitLogKeyStore implements LogKeyStore<int, CommitEntry?> {
         await _getBox().put(internalKey, commitEntry);
         // update the commitId in cache commitMap.
         _updateCacheLog(commitEntry.atKey!, commitEntry);
+        if (commitEntry.commitId != null &&
```
VJag, line 81: I see the remove method too. Who calls it? Do we have to adjust the latest commit id when an entry is deleted?

Future remove(int commitId) async

VJag: I see the method "Iterator getEntries(int commitId, {String regex = '.*'}) " Is this the method we use to get commit Id's after a given id's in sync? If yes, it's good and we are able to use the cache, however, is there a need to the cache map up to date in remove and update methods.

VJag: Future<List> _getValues() async  method is o(N) why do we need that?

sitaram-kalluri: For update and delete operation, commit method in at_commit_log.dart is invoked. Inside commit method we call add method in commit_log_keystore.dart

VJag (approved):

PR #366, opened 2021-10-26
lazy box and normal box support

murali-shris:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/at_secondary/at_persistence_secondary_server/lib/src/log/accesslog/access_log_keystore.dart
+++ b/at_secondary/at_persistence_secondary_server/lib/src/log/accesslog/access_log_keystore.dart
@@ -2,6 +2,7 @@ import 'dart:collection';
 import 'dart:io';
 
 import 'package:at_persistence_secondary_server/at_persistence_secondary_server.dart';
+import 'package:at_persistence_secondary_server/src/keystore/hive_base.dart';
 import 'package:at_persistence_secondary_server/src/log/accesslog/access_entry.dart';
 import 'package:at_utils/at_logger.dart';
 import 'package:at_utils/at_utils.dart';
@@ -9,7 +10,7 @@ import 'package:hive/hive.dart';
 
 export 'package:at_persistence_spec/at_persistence_spec.dart';
 
-class AccessLogKeyStore implements LogKeyStore<int, AccessLogEntry?> {
+class AccessLogKeyStore with HiveBase implements LogKeyStore<int, AccessLogEntry?> {
   var logger = AtSignLogger('AccessLogKeyStore');
   String? storagePath;
   final _currentAtSign;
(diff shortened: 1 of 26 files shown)


at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
```diff
@@ -340,7 +351,9 @@ class CommitLogKeyStore implements LogKeyStore<int, CommitEntry?> {
     var keys = _getBox().keys;
     var value;
     await Future.forEach(keys, (key) async {
-      value = await _getBox().get(key);
+      value = await (_isLazy
+          ? (_getBox() as LazyBox).get(key)
```
VJag: Do we have to cast to call a get()? it should be polymorphic.

at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
```diff
@@ -49,7 +55,9 @@ class CommitLogKeyStore implements LogKeyStore<int, CommitEntry?> {
   @override
   Future<CommitEntry?> get(int commitId) async {
     try {
-      var commitEntry = await _getBox().get(commitId);
+      var commitEntry = await (_isLazy
```
VJag: Do we have to cast to call a get()? it should be polymorphic.

at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
```diff
@@ -31,7 +33,11 @@ class CommitLogKeyStore implements LogKeyStore<int, CommitEntry?> {
     }
 
     this.storagePath = storagePath;
-    await Hive.openBox(_boxName);
+    if (_isLazy) {
```
VJag: Let's have a openBox() ideally in the base class for all stores that take isLazy and other params that initializes properly. Similarly, getBox() also needs to be in a base class.

Another option is, like java if Dart supports default methods have openBox and getBox() in an interface and make these classes implement or extend it.

at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
murali-shris: get method is not available in the base class(BoxBase). it is available only in the sub class (Box and LazyBox)

at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
murali-shris: sure..will make this change

at_secondary/at_persistence_secondary_server/lib/src/log/accesslog/access_log_keystore.dart
```diff
@@ -48,7 +51,9 @@ class AccessLogKeyStore implements LogKeyStore<int, AccessLogEntry?> {
   @override
   Future<AccessLogEntry?> get(int key) async {
     try {
-      var accessLogEntry = await _getBox().get(key);
+      var accessLogEntry = await (super.isLazy
```
VJag: Should this be super.getBox).get()?

VJag: get(key) is available on both the boxes. Do we need a cast()?

VJag (approved):

PR #419, opened 2021-12-03
Handle sync key does not exist in keystore.

sitaram-kalluri:
**- What I did** Handle KeyNotFoundException in SyncProgressiveVerbHandler when getting key value in for sync response

**- How I did it** Verify if key exists prior to getting value of a key.

**- How to verify it** Sync should complete successfully even if key does not exist. However, when key does not exist, log a message secondary server log.

### DIFF
--- a/at_secondary/at_secondary_server/lib/src/verb/handler/sync_progressive_verb_handler.dart
+++ b/at_secondary/at_secondary_server/lib/src/verb/handler/sync_progressive_verb_handler.dart
@@ -52,7 +52,7 @@ class SyncProgressiveVerbHandler extends AbstractVerbHandler {
         // If commitOperation is update (or) update_all (or) update_meta and key does not
         // exist in keystore, skip the key to sync and continue.
         if (!keyStore!.isKeyExists(itr.current.key)) {
-          logger.info(
+          logger.finer(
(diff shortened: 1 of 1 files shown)


at_secondary/at_secondary_server/lib/src/verb/handler/sync_progressive_verb_handler.dart
```diff
@@ -49,14 +49,20 @@ class SyncProgressiveVerbHandler extends AbstractVerbHandler {
       keyStoreEntry.commitId = itr.current.value.commitId;
       keyStoreEntry.operation = itr.current.value.operation;
       if (itr.current.value.operation != CommitOp.DELETE) {
+        // If commitOperation is update (or) update_all (or) update_meta and key does not
+        // exist in keystore, skip the key to sync and continue.
+        if (!keyStore!.isKeyExists(itr.current.key)) {
+          logger.info(
```
murali-shris: can we change level to finer

sitaram-kalluri: Modified the log level to finer.

murali-shris (approved):

PR #426, opened 2021-12-06
Don't run e2e tests for dependabot PRs

cpswan:
**- What I did**

Modified parameters for running e2e tests so that they shouldn't run for Dependabot PRs (which will fail for lack of secrets)

**- How to verify it**

Check that e2e tests don't run for next Dependabot PR

Don't run e2e tests for dependabot PRs


gkc (approved):
LGTM

PR #442, opened 2021-12-16
Notification expiry

srieteja:
**- What I did**
Assigned values to notifications with no expiry time for backward compatibility and also that other methods can perform further operations on these notifications.
**- How I did it**
Created a method called checkExpiry that ensures that notifications that were created without the "expiresAt" parameter are assigned with "notificationDateTime" as their expiry time.

**- How to verify it**

Notifications that were created with expiry time null now have assigned expiry time.


at_secondary/at_persistence_secondary_server/lib/src/notification/at_notification_keystore.dart
```diff
@@ -134,6 +135,39 @@ class AtNotificationKeystore
     return expiredKeys;
   }
 
+  Future<void> checkExpiry() async {
```
murali-shris: * move the logic to getExpiredKeys..
* default the number of hrs to 72..move this value to config.yaml

srieteja: Changes have been made as suggested.

murali-shris (approved):

sitaram-kalluri (approved):

Issue #472, opened 2022-01-17
Change default log level for prod secondaries to INFO


gkc:
**Describe the bug**
Logging is very verbose by default, and this verbosity has not proven useful in resolving issues in the past

**Expected behavior**
Logging level should be set to WARN by default in prod secondaries

See also #433 for broader context

gkc:
@VJag @murali-shris @kalluriramkumar As part of this work I notice that currently, in the secondary config, we have 'debug' and 'trace' options - only one of which is actually used, and neither of which are used as levels by the dart logger package (which uses FINEST, FINER, INFO, WARN etc)

I propose removing both, and adding a LogLevel param to the config instead, setting it to WARN by default, and modifying logging and config-setting code accordingly.

Does this sound OK to you?

VJag:
@gkc I agree.

cpswan:
As I just noted in #445 I think INFO should be the default.

gkc:
Should be able to close when changes in #476 are rolled out

cpswan:
@gkc how about a Canary release?

gkc:
@cpswan SGTM. Good exercise for me to do it. Can you ping me on chat to let me know the tasks required?

murali-shris:
closing this issue since server code is released to canary

PR #476, opened 2022-01-20
Change secondary config to use Dart logging's Levels, and set default to INFO

gkc:
See #472 for context

* at_secondary_server/lib/src/server/at_secondary_config.dart
  * Removed debugLog and traceLog from AtSecondaryConfig
  * traceLog wasn't used anywhere. Usages of debugLog in main.dart and in secondary_config_test.dart - see comments below
  * Replaced with 'logLevel' which will come from env var 'logLevel' or from YAML log.level and defaults to INFO. AtSecondaryConfig.logLevel can never be null
  * Added a TODO to make everything else in AtSecondaryConfig null safe, since everything has default values
* at_secondary_server/bin/main.dart
  * Removed reference to `AtSecondaryConfig.debugLog`; now uses `AtSecondaryConfig.logLevel`
* secondary_config_test.dart
  * Removed check that AtSecondaryConfig.debugLog is a boolean
  * Added a test to ensure that log level will be initialized correctly
* at_secondary_server/config/config_production.properties
  * **Added `log.level=WARNING` as we want production secondaries log level to be 'WARNING', but the default in config-base.yaml remains as FINEST, equivalent to existing settings of debug=true**
    * @cpswan I believe this will have the desired effect but need some advice on how best to test it
* at_secondary_server/config/config-base.yaml
  * removed log.debug and log.trace, added log.level as 'FINEST' since log.debug and log.trace were previously true
* at_secondary_server/config/config.yaml
  * removed log.debug and log.trace, added log.level as 'FINEST' since log.debug and log.trace were previously true
* at_functional_test/lib/secondary/base/config/config.yaml
  * removed log.debug and log.trace, added log.level as FINEST since log.debug and log.trace were previously true


cpswan:
@gkc we changed the way that we build and configure in staging a little while ago, so we no longer build a separate image carrying a separate config, but rather use ENV variables.

We also default to prod, so if we want WARNING in prod then that should be the default. (Though I think it should actually be INFO so that at least we get the version number into the logs at startup).

If we want the staging environment to log at FINEST then we should add that ENV variable to the Docker services there (and make it part of the template for future secondaries deployed there). Quite possibly we actually want staging to be as same as prod as possible, and we can selectively subject chosen secondaries to tailored logging levels.

gkc:
Thanks @cpswan 


> @gkc we changed the way that we build and configure in staging a little while ago, so we no longer build a separate image carrying a separate config, but rather use ENV variables.

Does this mean that config-production.properties and generate-config.py are obsolete?

> We also default to prod, so if we want WARNING in prod then that should be the default.

Yeah I went this way because my assumption was that config-production.properties was being used for prod, so I therefore changed config-production.properties but tried to leave everything else as unchanged as possible. I'm a bit wary of changing default log levels for everyone's secondaries in their dev / testing 

> (Though I think it should actually be INFO so that at least we get the version number into the logs at startup).

OK

> If we want the staging environment to log at FINEST then we should add that ENV variable to the Docker services there (and make it part of the template for future secondaries deployed there).

OK - can you point me at the repo & file(s) where I can add that ENV variable?

> Quite possibly we actually want staging to be as same as prod as possible, and we can selectively subject chosen secondaries to tailored logging levels.

sure

cpswan:
@gkc yes, there's a bunch of vestigial files. I'll use your branch and this PR to clear them out.

https://github.com/atsign-company/secondaries-scripts/blob/master/base/docker-compose.yml shows how the environment is set for the staging secondaries.

at_functional_test/lib/secondary/base/config/config.yaml
```diff
@@ -9,8 +9,7 @@ root_server:
 
 # The @Protocol application log configurations
 log:
-  debug: true
-  trace: true
+  level: FINEST
```
cpswan, line 12: We might want to keep FINEST here for when we run tests.

at_secondary/at_secondary_server/config/config.yaml
```diff
@@ -9,8 +9,7 @@ root_server:
 
 # The @Protocol application log configurations
 log:
-  debug: true
-  trace: true
+  level: FINEST
```
cpswan: INFO here I think

at_secondary/at_secondary_server/lib/src/server/at_secondary_config.dart
```diff
@@ -59,9 +59,8 @@ class AtSecondaryConfig {
   static final int? _stats_top_keys = 5;
   static final int? _stats_top_visits = 5;
 
-  //log configurations
-  static final bool _debugLog = true;
-  static final bool _traceLog = true;
+  //log level configuration. Value should match the name of one of dart logging package's Level.LEVELS
+  static final String _defaultLogLevel = 'FINEST';
```
cpswan: INFO here too

cpswan (changes requested):

at_functional_test/lib/secondary/base/config/config.yaml
gkc: yes I think so

at_secondary/at_secondary_server/config/config.yaml
gkc: Done

at_secondary/at_secondary_server/lib/src/server/at_secondary_config.dart
gkc: Done

cpswan (approved):
LGTM

Issue #501, opened 2022-02-03
A failed lookup hangs the TCP session and this builds up to a DoS on the secondary


cconstab:
Lead: @cconstab 

**Describe the bug**
When using the at_find_api to lookup a value if that value exists everything works. But if that value does not exist sessions to the secondary do not close and eventually the secondary starts rejecting connections (too many sessions)

**To Reproduce**
1. make a call to an @sign via the at_find_api to lookup a 'key' that does not exist
 `https://wavi.ng/api?atp=snackbarsforever.fourballcorporate9@ai6bh' - snackbarsforever.fourballcorporate9 does not exist !
2. AFter 20 + requests the secondary becomes unavailable
3. Connecting to the secondary using openssl the message that too many sessions 
```
read R BLOCK
error:AT0012-Inbound connection limit exceeded
read:errno=0
```

**Expected behavior**
If the value being looked up  this should not hang a TCP session perhaps that session should get dropped.

This is likely an at_lookup issue but could be in the at_find_api perhaps ?

gkc:
Moved to this repo as the problems diagnosed to date as part of investigating this issue are in the at_secondary_server codebase. There may be an issue with how at_lookup handles `error:` responses, if so then will create another issue in the at_libraries repo

gkc:
Remaining work for PR #504 : Write e2e tests (positive plookip, negative plookup) to verify behaviour. Write unit test to verify timeout behaviour. Estimate 2 SP

Work for other issues uncovered via this ticket will be in #503

PR #504, opened 2022-02-06
Fixes for issue #501

gkc:
Changes to address https://github.com/atsign-foundation/at_server/issues/501#issuecomment-1030029839

* outbound_message_listener.dart - old version would not handle error: responses but timeout after 5 seconds returning empty string
  * Modified read() so
    * it handles `error:...` responses by throwing an AtConnectException but leaves the connection open, as 'error:' is a valid response
    * handles malformed responses by throwing an AtConnectException and **closing** the connection (malformed responses should never occur)
    * changed default timeout from 5 seconds to 3 seconds as per comment from @cconstab 
    * throws a TimeoutException upon timing out, and closes the connection as timeouts are generally bad news
  * Modified _closeClient() so it doesn't check `if (!outboundClient.isInValid())` before calling outboundClient.close() - why? See [comments in code](https://github.com/atsign-foundation/at_server/pull/504/files#diff-1256148a521ddac147864fcc4cfd28c50b9974dfcbde0aa3531d60637233acf2R108)
  * Tidied up method comments for accuracy
  * For readability purposes, renamed instance variable client to outboundClient, and renamed method _closeClient to _closeOutboundClient

* outbound_connection_impl.dart
  * call socket.destroy() instead of socket.close()
  * We already call socket.destroy() in the at_lookup library see https://github.com/atsign-foundation/at_libraries/blob/trunk/at_lookup/lib/src/connection/base_connection.dart#L24 so this is normal behaviour when connecting to secondary servers
    * Originally changed base_connection.dart as well but reverted that just to prevent any unexpected side effects on **inbound_connection_impl** from this PR

* global_exception_handler.dart
  * added handling for AtConnectException and TimeoutException which are now being thrown by outbound_message_listener.dart

* proxy_lookup_verb_handler.dart
  * Added only an explanatory comment, about timeouts, 'error:..' responses and malformed responses being handled downstream in outbound_message_listener.dart

* Modified outbound_client.dart so that outbound_message_listener.dart log messages can include the remote atSign, host and port

Changes to address https://github.com/atsign-foundation/at_server/issues/501#issuecomment-1030029839

### DIFF
--- a/at_end2end_test/pubspec.yaml
+++ b/at_end2end_test/pubspec.yaml
@@ -11,12 +11,7 @@ dependencies:
       url: https://github.com/atsign-foundation/at_demos.git
       path: at_demo_data
       ref: master
-  at_lookup:
-    git:
-      url: https://github.com/atsign-foundation/at_libraries.git
-      path: at_lookup
-      ref: trunk
-      version: ^2.0.4
+  at_lookup: ^3.0.7
 
 dev_dependencies:
   lints: ^1.0.1
--- a/at_end2end_test/test/commons.dart
+++ b/at_end2end_test/test/commons.dart
@@ -26,16 +26,16 @@ void clear() {
 }
 
 ///Secure Socket Connection
-Future<SecureSocket> secure_socket_connection(host, port) async {
+Future<SecureSocket> secure_socket_connection(var host, var port) async {
   var socket;
-  while (true) {
+  while (retryCount > maxRetryCount) {
     try {
       socket = await SecureSocket.connect(host, port);
-      if (socket != null || retryCount > maxRetryCount) {
+      if (socket != null) {
         break;
       }
     } on Exception {
-      print('retrying for connection.. $retryCount');
+      print('retrying "$host:$port" for connection.. $retryCount');
       await Future.delayed(Duration(seconds: 5));
       retryCount++;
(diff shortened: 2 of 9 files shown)


gkc:
Something weird is happening with e2e tests not being able to open sockets. Debugging.

gkc:
pubspec.yaml currently has dependency override to pick up at_commons from trunk; this is because of the new exception that has been added, TimeoutException

Plan is to publish another version of at_commons and then remove the dependency override from here

gkc:
Need to add e2e test to verify expected behaviour

sitaram-kalluri (approved):

cconstab (approved):
My only comment is that 5 seconds is an eternity to wait for a response.. Esp if you have multiple timeouts.. Perhaps needs some testing but my gut says 1 second is plenty and if the secondary has not responded in a second then it is pretty unwell..
A conservative figure I thought I used 2 years back was 3 seconds, so not sure where 5 came from..


  Future<String?> read({int maxWaitMilliSeconds = 5000}) async {

gkc:
Agreed. All of the timeouts need to be reviewed but in the meanwhile I’ll
set this to 2 seconds

VJag (approved):

PR #507, opened 2022-02-09
add if key exists validation.

### LINKED ISSUE #387: plookup for a non existent key fails with format exception
**Describe the bug**
Doing a plookup of a non-existent key throws format exception

**To Reproduce**
1. First start the server 
2. Then from say bob's atsign do a plookup for a non- existen key
bob@plookup:random@ alice

**Expected behavior**
It should throw an error as Key not found exception or show as null

**Screenshots**

Screenshot from 2021-11-11 17-20-08
Screenshot from 2021-11-11 17-20-17

**Additional context**
branch - exception_null_value

sitaram-kalluri:
**- What I did** Fix plookup fails for non existent keys.

**- How I did it** Verify if key exists in keystore before calling keystore.get.

**- How to verify it** When looked up for non existent keys, server should return data:null instead of error

**-Dev Testing**
@sitaram@plookup:country@murali
data:India
@sitaram@plookup:city@murali
data:null


sitaram-kalluri:
The issue is fixed in [PR-504](https://github.com/atsign-foundation/at_server/pull/504)

PR #512, opened 2022-02-11
Temporarily skip a notify test flaking due to a timing issue

gkc:
**- What I did**
* One of the notify tests is flaking due to a timing issue, so I've marked it as *skip* so that other PRs aren't held up while I'm working on a fix


cconstab (approved):
Make a mental note to come back to this one !

PR #517, opened 2022-02-15
Update the expired certs

sitaram-kalluri:
**- What I did** Update the expired certs.


murali-shris (approved):

sitaram-kalluri:
Pulled the latest certs from [here](https://github.com/atsign-foundation/at_virtual_environment/blob/trunk/secondary/base/certs).

With the new certs, when secondary run through the code, it works fine. However when running on the latest docker image, facing certs expiry issue.

Following is the log snippet:

`
SEVERE|2022-02-16 09:10:00.538972|AtSecondaryServer|HandshakeException: Handshake error in server (OS Error: 
	CERTIFICATE_VERIFY_FAILED: certificate has expired(handshake.cc:393)) `

Following is the docker image used in local: [atsigncompany/virtualenv:dev_env](https://hub.docker.com/layers/atsigncompany/virtualenv/dev_env/images/sha256-41c50dd687afa28da9266cb57e08f7352be18931da604888f287e20c24874b2d?context=explore)

Issue #518, opened 2022-02-15
**Blocker**Certificate expiry with latest atsigncompany/virtualenv:dev_env image


sitaram-kalluri:
**Describe the bug**
Update the certs in `at_server/at_functional_test/lib/secondary/base/certs` with the latest certs and still seeing the certs expiry issue. This causes the failure of functional tests in at_client_sdk.

[PR](https://github.com/atsign-foundation/at_server/pull/517/files) with which the certs are updated.

**To Reproduce**
1. Pull the latest docker `atsigncompany/virtualenv:dev_env` image and run the container.
2. Run a plookup or lookup verb.
3. The server logs show the certificate expiry issue.
Snippet from server logs:
`SEVERE|2022-02-15 14:27:07.904971|AtSecondaryServer|HandshakeException: Handshake error in server (OS Error: 
	CERTIFICATE_VERIFY_FAILED: certificate has expired(handshake.cc:393)) `

**Expected behavior**
Certificate expiry should not occur.

**Additional context**
There is a change in verified By field between the old and new certs(Not sure if this is the cause; just an observation).

![image](https://user-images.githubusercontent.com/16645698/154093575-bd07dc12-f4e9-4187-a276-4f14ccd20a8b.png)

gkc:
All certs are as they should be for this to work

Most likely a problem with verifying CA not being in the chain that the docker container sees

@cconstab any ideas on how to verify that (& fix it if so)?

gkc:
The certs in the virtual environment image are the right ones, the image was pushed by https://github.com/atsign-foundation/at_server/actions/runs/1845269611

gkc:
What I don't understand is why the functional tests are failing in at_client_sdk when they ran successfully in at_server, as the GitHub action builds the virtualenv docker image, and the tests are run against that docker image

gkc:
OK I've downloaded the LetsEncrypt R3 CA cert
Before adding it to cacert.pem:
```
gkc2019:certs gary$ openssl verify -CAfile cacert.pem cert.pem 
cert.pem: CN = vip.ve.atsign.zone
error 20 at 0 depth lookup:unable to get local issuer certificate

gkc2019:certs gary$ openssl verify -CAfile cacert.pem fullchain.pem 
cert.pem: CN = vip.ve.atsign.zone
error 20 at 0 depth lookup:unable to get local issuer certificate
```

After adding it to cacert.pem:
```
[...]

PR #519, opened 2022-02-15
Adding LetsEncrypt root R3 cert to cacert.pem

gkc:
See https://github.com/atsign-foundation/at_server/issues/518 for details of the problem, the investigation and why this PR might hopefully fix it

Summary of this PR - added the LetsEncrypt R3 CA root cert to the three cacert.pem files in this repo


cconstab (approved):
If this is the problem it must of been hanging around for a while..

gkc:
Last certs had verisign as root ca, not letsencrypt

cpswan:
I've just backed out this change, as we shouldn't have intermediate certs in the cacert.pem files.

This was a flare up of the issue we hit last year when DST X3 was retired, and we need to make sure that that CA isn't in cacert.pem

It only bit us when we switched to LetsEncrypt certs for the virtual environment, having previously used ZeroSSL.

For the time being I've generated new ZeroSSL certs for VE, but going forward I think having DST X3 out of cacert.pem should let us use LetsEncrypt certs without issues (as we do for all the prod and staging secondaries).

PR #522, opened 2022-02-16
Fix for at_client_sdk-issue#384

yahu1031:
**- What I did**
- Prevented the cached keys getting updated with empty value when refresh job runs.

**- How I did it**
- Add IF condition to check if value is empty. If empty, do not update the cached key value.

**- How to verify it**
Scenario - 1
1. Notify a key along with value with TTR set to -1
2. Reduce the corn job timer to 1 minute for Refresh Job.
3. When the cron job runs, the cached key created in step-1 should not be picked up.

Scenario - 2
1. Notify a key along with value with TTR set to 10 seconds
2. Reduce the corn job timer to 1 minute for Refresh Job.
3. When the cron job runs, the cached key value should not be updated to empty value.

**Additional context**
Fix for https://github.com/atsign-foundation/at_client_sdk/issues/384


at_secondary/at_secondary_server/lib/src/refresh/at_refresh_job.dart
```diff
@@ -21,32 +22,37 @@ class AtRefreshJob {
   final logger = AtSignLogger('AtRefreshJob');
 
   /// Returns the list of cached keys
-  Future<List<dynamic>?> _getCachedKeys() async {
-    var keysList = await keyStore.getKeys(regex: CACHED);
+  Future<List<String>?> _getCachedKeys() async {
+    List<String>? keysList = keyStore?.getKeys(regex: CACHED);
```
murali-shris: keystore can never be null on server.
getKeys() returns List<String>. List<String>? not needed

yahu1031: @murali-shris I added this because `secondaryPersistenceStore.getSecondaryKeyStore()` returns a nullable `HiveKeystore`. So the return will be nullable. As the return was nullable, the `keyList` can be null. I can add a condition to check `keyStore` is null.

murali-shris (approved):

PR #556, opened 2022-03-14
Root Dockerfile refactored to be like Secondary

### LINKED ISSUE #467: Refactor root dockerfile to be similar to secondary
**Is your feature request related to a problem? Please describe.**
The root dockerfile is presently using 'FROM ubuntu:latest`, which increases size and attack surface area.

**Describe the solution you'd like**
Move to a two stage Docker build that utilises `FROM dart:version` then `FROM scratch`

**Describe alternatives you've considered**
If the root needs more of an OS underneath it then we could use Alpine or a slim Debian build.

cpswan:
fixes #467 

**- What I did**

Modified at_root_server Dockerfile to use two stage build similar to secondary Dockerfile

Also some tidying of secondary Dockerfile, and added root Dockerfile to Dependabot.

**- How to verify it**

Should be tested locally with appropriate Redis environment

Root Dockerfile refactored to be like Secondary


cconstab (approved):
much neater thanks

PR #563, opened 2022-03-15
Info and noop verbs

gkc:
**- What I did**
New verbs 'info' and 'noop'

**- How I did it**
* at_commons change and [PR](https://github.com/atsign-foundation/at_tools/pull/163)
* at_server_spec change and [PR](https://github.com/atsign-foundation/at_server/pull/561)
* This PR:
  * Added noOp and info to VerbEnum in at_server_spec/lib/verbs.dart so they are available via the package (doh!)
  * Implementations of Info and NoOp verbs. Yes they are both very very simple.
  * Made a symbolic link from at_functional_test/lib/secondary/base/pubpsec.yaml to at_secondary/at_secondary_server/pubspec.yaml - so that I can run my secondary server locally in working directory at_functional_test/lib/secondary/base and it will still pick up the version from the pubspec.yaml


gkc:
Hi @nickelskevin @cconstab Please can you do a quick review of this PR (it's pretty safe & unexciting) ... Sita & Jagan also want to use the 'features' thing supported by the new 'info' command to have graceful forwards and backwards compatibility for the new notification stuff they're working on, so I spent time today putting together pull requests to make that available to them tomorrow

@VJag if Colin & Kevin don't get to this, please review and merge in your morning if it looks good to you

cconstab (commented):
noop works fine.. 

```
@noop:5000
data:ok
@noop:4000
data:ok
@noop:5000
data:ok
@info
data:{"version":null,"uptime":"8 minutes 6 seconds","features":[{"name":"No-Op verb","status":"Beta","description":"NoOp simply does nothing for the requested number of milliseconds. The requested number of milliseconds may not be greater than 5000. Upon completion, the noop verb sends 'ok' as a response to the client."}]}
@
```

But what worries me is that commands sent whilst in the noop also get executed during the noop.. I predict things getting confused because of that

I would have expected noop not to accept any commands  in the network stream until it has finnished noop'ting..

e.g. sending a noop:5000 whilst noop is waiting results in this..

```
@noop:5000
noop:5000
data:ok
@data:ok
@
```

video is easier to see

https://asciinema.org/a/X3T6QDw2k225Ytw5uojl8EsL1

even better a scan works although there was a noop

https://asciinema.org/a/9IWxce8gWIX4UqG9ZxH8CK2gx


It takes two to tango, so the client side might well wait for the @ but on the other hand something else might not...

gkc:
Very interesting. Yes I would have expected it to block also. As there is
nothing special going on with this verb, this must be an underlying feature
of the current server impl

Can you approve this PR please, and let’s have a separate bug for “requests
on a given connection must be fully handled before subsequent inbound
request is handled” ?

cconstab:
Approving aasap

cconstab (approved):
after discussion with @gkc the issue I found/have is something that needs work outside of this PR

at_secondary/at_secondary_server/lib/src/verb/handler/noop_verb_handler.dart
```diff
@@ -0,0 +1,32 @@
...
+  Future<void> processVerb(Response response, HashMap<String, String?> verbParams, InboundConnection atConnection) async {
+    String? delayMillisParam = verbParams['delayMillis']; // as per NoOp documentation, delayMillis may not be more than 5000
+    if (delayMillisParam == null) {
+      throw IllegalArgumentException(noOpVerb.usage() + " where the duration maximum is 5000 milliseconds");
+    }
+    var delayMillis = int.parse(delayMillisParam);
+    if (delayMillis > 5000) {
+      throw IllegalArgumentException(noOpVerb.usage() + " where the duration maximum is 5000 milliseconds");
+    }
+    await Future.delayed(Duration(milliseconds:delayMillis));
```
VJag, line 29: Gary, should we use "sleep(Duration(milliseconds:delayMillis))" for a blocking wait?

VJag (approved):

gkc:
Re “sleep” - no, because that will block the entire isolate as I understand
it - see
https://api.dartlang.org/stable/dart-io/sleep.html

PR #573, opened 2022-03-17
Don't try to sudo echo

cpswan:
**- What I did**

Another try at getting permissions right inside of GitHub Actions


gkc (approved):
Do you still want to sudo python3?

cpswan:
I should only need sudo for the `mkdir`

Issue #503, opened 2022-02-05
Reduce likelihood of server reaching capacity on inbound connections


gkc:
Lead: @gkc 

**Describe the bug**
* There is currently some reaping of inbound connections in the at_secondary_server, as follows:
  * When the listener asks InboundConnectionManager to create an inbound connection, the manager checks its 'hasCapacity()' method
  * hasCapacity() will attempt to close all 'inValid' inbound connections - i.e. connections where `_isIdle() || getMetaData().isClosed || getMetaData().isStale;`
  * connections are deemed idle once the configurable duration `inboundIdleTimeMillis` has passed without activity. This currently defaults to 600000 i.e. 600 seconds i.e. 10 minutes
* However we have no real defense against using-up-all-available-inbound-connections DoS attacks on individual secondaries whether deliberate or as an inadvertent result of badly-written clients
* See #502 for some additional context

**Expected behavior**
Needs further analysis. Some initial thoughts:
* As with #502 for outbound clients, the current hard limit on number of inbound clients should probably become a low-water-mark.
* Until low-water-mark is reached, we can continue with the current approach
  * We can also 'prefer' to reap connections in this order: unauthenticated, authenticated but from some other atSign, authenticated and from this secondary's atSign
* While above low-water-mark:
  * If no inbound clients can be reaped because they are all active, then we can add additional inbound clients up to a high-water-mark
  * as we approach the high-water-mark, we can start to reduce `inboundIdleTimeMillis` from its configured value down to an absolute minimum of, say, 1 second
  * and also logging WARNING or SEVERE log messages as we approach the high-water-mark, so that some appropriate action can be taken whether by human or automation (e.g. increase the limits for this secondary, block the originating IP addresses, etc)

As we implement what is required for this and for #502, we will need to also invest in client-side handling of outbound connections. There is some discussion of what this looks like in #502 - basically it means if we are keeping outbound connections open from the clients, then we need to have graceful retries for when the server end has had to kill them

**Additional context**
As the network of connections between atSigns grows, we need to have policies which will optimise for the "normal" number of inter-server connections but will gracefully handle much larger numbers

gkc:
@cconstab @cpswan @VJag @murali-shris @kalluriramkumar @nickelskevin Interested in any initial comments or thoughts you have on this

gkc:
Marked this as P1 since we are seeing issues related to handling of connections in the secondary servers.

VJag:
[...]

PR #583, opened 2022-03-18
Better inbound connection management

gkc:
**- What I did**
Made changes to inbound connection management as first wave of fixing problems described in https://github.com/atsign-foundation/at_server/issues/503

This PR should largely eliminate issues under normal operations. It will not prevent a determined DOS attacker; more changes are required for that, eg temporarily banning IP addresses of abusive connections, faster auth handshaking, and so on. 

**- How I did it**
* config.yaml : Added some more explanation on the inbound_idle_time_millis config parameter
* inbound_connection_impl.dart
  * added optional construction parameter, owningPool which, when provided, allows the inbound connection impl to do some fancier logic to do with idle time and thus allow the connection pool to do some fancier connection pool maintenance
  * added an override for the close() method, as previously done for (outbound connection impl), to destroy the underlying socket rather than just closing it
* inbound_connection_manager.dart
  * pass the owningPool parameter when constructing InboundConnectionImpls
  * some linting
* inbound_connection_pool.dart
  * log WARNING when reaches 85% full, SEVERE when reaches 95% full
  * log INFO when drops to < 95% full, INFO when drops to < 85% full
  * some linting
* at_secondary_impl.dart : one tiny code format change
* server_context.dart
  * bumped inboundConnectionLimit to same as latest in config.yaml (50)
  * bumped outboundConnectionLimit to same as latest in config.yaml (50)
  * added a few more global config parameters which we can later add to config.yaml if we need them to be tunable:
    * unauthenticatedMinAllowableIdleTimeMillis - the minimum amount of inactive time that needs to pass before a connection can be cleaned up as idle
    * inboundConnectionLowWaterMarkRatio - beyond this proportion of the inboundConnectionLimit, we get progressively more aggressive when determining acceptable idle times for connection cleanup
    * progressivelyReduceAllowableInboundIdleTime - a switch to disable all of this new logic if we want to
* inbound_connection_pool_test.dart : added some tests for the above changes

**- How to verify it**
Check that the added tests are sensible and that all the tests pass

Reduce likelihood of server reaching capacity on inbound connections #503

### DIFF
--- a/at_secondary/at_secondary_server/config/config.yaml
+++ b/at_secondary/at_secondary_server/config/config.yaml
@@ -39,9 +39,11 @@ hive:
 
 # The @Protocol connection configurations.
 connection:
-  # The maximum time in milli seconds for an inbound connection to expire.
+  # The maximum time in milliseconds for an inbound connection to expire.
+  # This should NEVER be set to less than 60 seconds. Internally, the effective
+  # value used is reduced progressively as the number of connections approaches the max limit
   inbound_idle_time_millis: 600000
-  # The maximum time in milli seconds for an outbound connection to expire.
+  # The maximum time in milliseconds for an outbound connection to expire.
   outbound_idle_time_millis: 600000
   # Creates an inbound connection and adds the connection to inbound connection pool. At any point, at most [inbound_max_limit] connections can only be created.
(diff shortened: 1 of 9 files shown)


cpswan (approved):
@gkc it would be useful to do a walk through of this PR in an eng stand to serve as an example of what good looks like.

PR #596, opened 2022-03-24
shared key and public key checksum support for notify

murali-shris:
**- What I did**
- added shared key and public key checksum support in notify
**- How I did it**
- modified notify verb handler to set the metadata when sharedkey and publickey checksum are passed in notify verb
**- How to verify it**
- run the unit test in notify_verb_test.dart "A group of tests to verify public key checksum and shared key on metadata "
- Notify verb handler changes for shared key and public key checksum in metadata

### DIFF
--- a/at_end2end_test/pubspec.yaml
+++ b/at_end2end_test/pubspec.yaml
@@ -12,8 +12,8 @@ dependencies:
       url: https://github.com/atsign-foundation/at_demos.git
       path: at_demo_data
       ref: master
-  at_lookup: ^3.0.7 
-  at_client: ^3.0.9
+  at_lookup: ^3.0.14
+  at_client: ^3.0.13
 
 dev_dependencies:
   lints: ^1.0.1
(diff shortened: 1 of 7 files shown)


gkc (commented):
LGTM - only one question - are there functional or e2e tests for this? If not, we should add them.

(If there already are functional/e2e tests in the at_client_sdk repo for example, then that's fine of course)

murali-shris:
> LGTM - only one question - are there functional or e2e tests for this? If not, we should add them.
> 
> (If there already are functional/e2e tests in the at_client_sdk repo for example, then that's fine of course)

For the at_client PR..I have added an end2end test. I will add a test in server as well..

sitaram-kalluri (approved):

murali-shris:
> > LGTM - only one question - are there functional or e2e tests for this? If not, we should add them.
> > (If there already are functional/e2e tests in the at_client_sdk repo for example, then that's fine of course)
> 
> For the at_client PR..I have added an end2end test. I will add a test in server as well..

@gkc The test I added is failing because server in prod/trunk will be on old version(which doesn't understand new notify syntax).

gkc:
Ah! Ok, cool. We can add the test later, when this is in prod.

PR #615, opened 2022-03-31
Fixing 'invalid connection' problems in e2e test env

gkc:
**- What I did**
* Set inbound_max_limit and outbound_max_limit to 50 in e2e test server yaml files
* Ensure that InboundConnection's 'lastAccessed' metadata is set not just when a response is written to the connection, but also when an incoming request is being sent for handling. This will ensure that servers do not close connections for active requests even when they are under super-heavy load and at configured connection count limits.

**- How to verify it**
* The e2e tests should all pass

### DIFF
--- a/at_end2end_test/config/config12.yaml
+++ b/at_end2end_test/config/config12.yaml
@@ -5,6 +5,10 @@ root_server:
   # The url to connect to root server
   url: 'root.atsign.wtf'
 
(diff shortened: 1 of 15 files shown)


cpswan (approved):

gkc:
Issue 617 was fixed by pull 616

PR #616, opened 2022-04-01
Remove singleton nature of DummyInboundConnection

### LINKED ISSUE #617: Erroneous closing of outbound notification connections
**Describe the bug**
If the Notification ResourceManager encounters an exception while it attempts to connect an outbound client, it sets isClosed on its inboundConnection. Because inboundConnection for notifications is a fake object, DummyInboundConnection, which is a singleton, this means isClosed is set on that connection for all existing and future notification OutboundClients.

Whenever the Notification ResourceManager starts to send notifications to another secondary, there is a check of the pool of existing outbound clients to see if there is capacity. That check works by asking every outbound client if it is valid. The OutboundClient's isInvalid method will return true if its associated InboundConnection is inValid - which will now always be the case, because we've marked the singleton DummyInboundConnection as closed. The pool manager then calls close() on all inValid OutboundClients, which in turn leads to all of those clients' outbound *connections* being closed.

The Notification ResourceManager has a scheduled job which periodically wakes up and sends any queued notifications. This job has a queue of at-signs, each of which has a list of notifications to be sent to that at-sign's secondary. When there is more than one at-sign which has notifications queued for sending to another secondary, the outbound connections are pre-emptively being closed for all but the last at-sign in the queue, because the last at-sign's connections will not be closed until the next time the scheduled job runs.

All of this results in notifications frequently being stuck in 'Queued' state for a very long time, because the outbound connections will either not have even sent the notification to the remote secondary, or will have sent the notification but been pre-emptively closed before the response is received.

**To Reproduce**
Run the e2e tests with any server version prior to the future 3.0.14 where this will be fixed

**Expected behavior**
Outbound connections should not be closed unnecessarily. Notifications should consistently be delivered to remote secondary in less than 5 seconds if the remote secondary is available.

**Additional context**
Estimating effort retrospectively at 8SP - multiple days of investigation by multiple people, plus the fix effort

Note that identifying the issue was only possible through detailed analysis of the logs of the cicd[1-6] servers

sitaram-kalluri:
**- What I did**
Fixed #617 

**- How I did it**
Removed the singleton nature of DummyInboundConnection

**- How to verify it**
e2e tests should pass

Fixed bug which could cause premature erroneous closing of outbound connections while sending notifications to other secondaries

### DIFF
--- a/at_end2end_test/test/notify_verb_test.dart
+++ b/at_end2end_test/test/notify_verb_test.dart
@@ -384,7 +384,7 @@ void main() {
         returnWhenStatusIn: ['expired'], timeOutMillis: 1000);
     print('notify status response : $response');
     expect(response, contains('data:expired'));
-  });
+  },skip: 'Non existent atSign. Skipping the test for now to avoid connection issue');
 
   test('notify verb with notification expiry with messageType text', () async {
     //   /// NOTIFY VERB
--- a/at_end2end_test/test/update_verb_test.dart
+++ b/at_end2end_test/test/update_verb_test.dart
@@ -226,7 +226,7 @@ void main() {
     response = await sh1.read();
     print('llookup verb response : $response');
     expect(response, contains('data:unicorn'));
-  });
+  }, skip: 'Non existent atSign, skipping the test to avoid connection issue');
 
   test('update-llookup for ttl ', () async {
     ///UPDATE VERB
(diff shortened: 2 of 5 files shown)


gkc (approved):

gkc:
@kalluriramkumar can you file a ticket please, to un-skip these tests once we've released version 3.0.14 to prod

PR #632, opened 2022-04-11
Fixes #621

### LINKED ISSUE #621: Edge case means some notifications can be lost
**Describe the bug**s
In [this method](https://github.com/atsign-foundation/at_server/blob/883623b080b5247f5b6528edad88fe112d8639ef/at_secondary/at_secondary_server/lib/src/notification/resource_manager.dart#L90)

* if an exception happens while sending a notification
* and the iterator has other notifications not yet processed
* then those other notifications are not added to the errorList
* and therefore they are not re-enqueued

Note that this issue is significantly less likely to occur, now that [this other bug](https://github.com/atsign-foundation/at_server/issues/617) has been fixed.

**To Reproduce**
Very difficult. We need a unit test which can simulate the issue; in turn this means we will need to restructure the Notification ResourceManager so we can inject its dependencies, use mocktail to create mock dependencies, and configure those mock dependencies to behave in a way which supports the test case. [This (still draft) PR](https://github.com/atsign-foundation/at_client_sdk/pull/463) has an example of how to do this - but note that the server's Notification ResourceManager is much more complex than the client's Monitor

**Expected behavior**
The 'other notifications' mentioned above should be re-enqueued. The code fix here is simple. But writing the unit test which reproduces the problem prior to the fix being applied, and passes after the fix is applied, is a significant piece of work; I estimate at least 5SP

gkc:
In the _sendNotifications method, which is called for a single atSign:
* if an exception happens while sending a notification
* and the iterator has other notifications not yet processed
* then those other notifications are not added to the errorList
* and therefore they are not re-enqueued

What we want:
* The 'other notifications' mentioned above should be re-enqueued.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/at_secondary/at_secondary_server/lib/src/notification/resource_manager.dart
+++ b/at_secondary/at_secondary_server/lib/src/notification/resource_manager.dart
@@ -106,6 +106,9 @@ class ResourceManager {
       logger.severe(
           'Exception in processing the notification ${atNotification.id} : ${e.toString()}');
       errorList.add(atNotification);
+      while (iterator.moveNext()) {
+        errorList.add(iterator.current);
+      }
     } finally {
       //1. Adds errored notifications back to queue.
       await _enqueueErrorList(errorList);


VJag (approved):

gkc:
As this is an interesting part of the code base, I'm going to request that @murali-shris also review, please; I don't want to merge until another independent pair of eyes has verified

murali-shris (approved):

PR #640, opened 2022-04-14
Meta data cache

### LINKED ISSUE #600: Scan returns expired keys
**Describe the bug**
The scan verb returns the expired keys in the server and SDK.

**To Reproduce**
1. Create a key with TTL.
2. Once the key is expired (TTL met), run scan verb
3. The expired key is also returned.

**Expected behavior**
Scan should not return the expired keys.

**Additional context**
@VJag and myself had a discussion on this and below are the approaches:

Approach  - 1
------------------
1. Return a map from the key store.
2. Iterate on the map; For each key, on its metadata verify if key is expired.

Approach - 2
------------------
1. Maintain an internal data structure to cache the key and its expired time.

Approach - 3
------------------
1. Add a plugin to hive keystore which will not return the keys that are expired.

srieteja:
closes https://github.com/atsign-foundation/at_server/issues/600

**- What I did**
compare key expiry time with present time to decide if exired or not. Using the metadata cache to make this faster

**- How I did it**
Introduced a metaDataCache and updating it along with the keystore
**- How to verify it**


at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
```diff
@@ -255,7 +273,7 @@ class HiveKeystore implements SecondaryKeyStore<String, AtData?, AtMetaData?> {
   /// @param - regex : Optional parameter to filter keys on regular expression.
   /// @return - List<String> : List of keys from secondary storage.
   @override
-  List<String> getKeys({String? regex}) {
+  List<String> getKeys({String? regex, bool? removeExpired}) {
```
VJag: ignoreExired

at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
```diff
@@ -381,4 +407,8 @@ class HiveKeystore implements SecondaryKeyStore<String, AtData?, AtMetaData?> {
       await persistenceManager.openVault(persistenceManager.atsign!);
     }
   }
+
+  bool isExpired(key) {
+    return _metaDataCache[key]!.expiresAt!.isBefore(DateTime.now().toUtc());
```
VJag, line 392: - Check where expiresAt is being set
- If it is set on server, and if the key happens to be created and exired when the device is offline then expiredAt would have not been set. We have to come up with a solution for this.

at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
```diff
@@ -288,19 +308,23 @@ class HiveKeystore implements SecondaryKeyStore<String, AtData?, AtMetaData?> {
...
+    //   logger.severe('HiveKeystore getMeta exception: $exception');
+    //   throw DataStoreException('exception in getMeta: ${exception.toString()}');
+    // } on HiveError catch (error) {
+    //   await _restartHiveBox(error);
+    //   logger.severe('HiveKeystore getMeta error: $error');
+    //   throw DataStoreException(error.message);
+    // }
+    // return null;
+    if (_metaDataCache.containsKey(key)) {
+      return _metaDataCache[key];
```
VJag, line 304: Do we need a fallback to get from hive? Donno. Let's discuss this with @gkc.

gkc: Not yet. Commit log key-to-commit-id map is entirely in memory; let's do the same here for now. It will be simple later to switch to an LRU cache which has a configurable maximum size and reads from hive on cache hits

at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
srieteja: only returning unexpired keys now

at_secondary/at_secondary_server/lib/src/server/at_secondary_impl.dart
```diff
@@ -421,6 +421,8 @@ class AtSecondaryServerImpl implements AtSecondaryServer {
     hiveKeyStore.commitLog = atCommitLog;
     _commitLog = atCommitLog;
     keyStoreManager.keyStore = hiveKeyStore;
+    // Initialize the hive metadata cache
+    hiveKeyStore.initMetaDataCache();
```
VJag: This need not be called explicitly. Can you check how commit log cache is initialised?

at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
```diff
@@ -7,23 +9,38 @@ import 'package:at_utils/at_logger.dart';
...
   }
 
+  Future<void> initMetaDataCache() async {
+    if (persistenceManager == null || !persistenceManager.getBox().isOpen) {
+      logger.severe(
+          'persistence manager not initialized. skipping metadata caching');
+      return;
+    }
+    var keys = persistenceManager.getBox().keys;
+    await Future.forEach(
```
VJag, line 37: Add some logs for before and after. Logs help us to see how the cache init took or if it is in progress.

at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
```diff
@@ -381,4 +378,26 @@ class HiveKeystore implements SecondaryKeyStore<String, AtData?, AtMetaData?> {
...
   }
+
+  bool _isExpired(key) {
+    if (_metaDataCache[key]?.expiresAt == null) {
+      return false;
+    }
+    return _metaDataCache[key]!.expiresAt!.isBefore(DateTime.now().toUtc());
+  }
+
+  bool _isBorn(key) {
```
VJag, line 395: Actually, the right method name should _isNotBorn. If "availableAt" is before "DateTime.now().toUtc()" then it is not born. Born is when availableAt is after now.

VJag (changes requested):
Overall the changes are good. Minor comments.

at_secondary/at_secondary_server/lib/src/server/at_secondary_impl.dart
srieteja: The commit log Keystore already had an initialize method prior to introducing the cache. The cache is initialized through that method.

at_secondary/at_secondary_server/lib/src/server/at_secondary_impl.dart
```diff
@@ -421,6 +422,8 @@ class AtSecondaryServerImpl implements AtSecondaryServer {
     hiveKeyStore.commitLog = atCommitLog;
     _commitLog = atCommitLog;
     keyStoreManager.keyStore = hiveKeyStore;
+    // Initialize the hive metadata cache
```
VJag: Pls change this to initialize the hive store

VJag (approved):

VJag (approved):

PR #649, opened 2022-04-20
Remove last accessed check in stats_notification_service.dart

### LINKED ISSUE #646: stats not being written by server to monitor connections with frequent heartbeats
**Describe the bug**
If the client-side monitor heartbeat interval is too short, stats are no longer being delivered to the client, because the server-side stats notification service decides to write the stats to the monitor only if it is at least 15 seconds since the connection was last written to by the server

There is a workaround available which is to set the monitor heartbeat interval to a larger value on the client. Current default is 10 seconds; changing it to a much higher value (e.g. 60 seconds) will limit the impact of this bug while still retaining the value of the monitor heartbeat.

**Expected behavior**
Instead of checking when the monitor connection was last accessed (currently this means the last time it was written to by the server), the stats_notification_service should instead keep track of when the stats were last sent, and only send again if it's been more than 15 seconds since the last time the stats were sent

sitaram-kalluri:
**- What I did**
Remove the condition which check if lastAccessedTime duration on connection metadata is less than 15 seconds to enable write latest server commit id for every 15 seconds
**- How I did it**

**- How to verify it**

### DIFF
--- a/at_secondary/at_secondary_server/lib/src/notification/stats_notification_service.dart
+++ b/at_secondary/at_secondary_server/lib/src/notification/stats_notification_service.dart
@@ -41,7 +41,8 @@ class StatsNotificationService {
   final _logger = AtSignLogger('StatsNotificationService');
   final String _currentAtSign =
       AtSecondaryServerImpl.getInstance().currentAtSign;
-  var _atCommitLog;
+  AtCommitLog? _atCommitLog;
+  DateTime _lastSentTime = DateTime.now().toUtc();
 
   /// Starts the [StatsNotificationService] and notifies the latest commitID
   /// to the active monitor connections.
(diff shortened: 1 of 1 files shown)


at_secondary/at_secondary_server/lib/src/notification/stats_notification_service.dart
```diff
@@ -79,15 +80,7 @@ class StatsNotificationService {
       // Iterates on the list of active connections.
       await Future.forEach(connectionsList,
           (InboundConnection connection) async {
-        // If a monitor connection is stale for 15 seconds,
```
gkc, line 82: Removing this block entirely will mean that stats are written not only by the scheduled send which runs every `AtSecondaryConfig.statsNotificationJobTimeInterval` seconds, but also every time there is a Change operation (see ChangeVerbHandler) or a notify that results in a keyStore change operation (see NotifyVerbHandler).

Is that what we want, or do we just want it to run periodically?

gkc (changes requested):

at_secondary/at_secondary_server/lib/src/notification/stats_notification_service.dart
sitaram-kalluri: @gkc : Added _lastSentTime to StatsNotificationService which will prevent notification on every update and notify.

gkc: OK so that will match current behaviour pretty much, I think this is good enough for now given we will have streaming sync in the near future

gkc (approved):

PR #664, opened 2022-05-01
Straight through notify

### LINKED ISSUE #665: Eliminate all notification latency where possible, end-to-end
**Describe the bug**
Significant notification latency caused by
1. Server attempts to forward notifications periodically - currently every 5 seconds
2. Server closes outbound notification connections to other servers after sending them notifications. Thus every notification carries the full overhead of server-to-server authentication
3. Server persists data before queueing the notification for delivery by the notifications ResourceManager 

**Expected behavior**
1. Keep the periodic job (necessary for retries) but nudge it into activity every time there is a new notification queued for delivery
2. (a) Keep outbound notification connections open (b) make the notification connection pool into a Least-Recently-Used pool so that when we reach our configured limit on number of outbound connections, the LRU connection is closed
3. Queue the notification for delivery before persisting it

**Additional context**
Notifications are a critical component of the platform and should be delivered as close to instantaneously as possible.

gkc:
### NOTE!
This branch includes changes to at_persistence_secondary_server, so at_secondary_server's pubspec.yaml in this branch has a dependency override to get at_persistence_secondary_server from this same branch rather than from pub.dev.

**When this PR is merged**, it should immediately be followed by
- [x] Publish new version of at_persistence_secondary_server (its pupspec.yaml and CHANGELOG.md are ready to go)
- [x] Create a new PR which removes the dependency override for at_persistence_secondary_server from at_secondary_server's pubspec.yaml

### Update, May 12
This is why automated test packs are among the most valuable, if not **_the_** most valuable, of a platform's assets
* I un-skipped the e2e tests for notifications to non-existent destination atSigns, which had previously been marked 'skip' because the tests after them were flaking at the time (for a problem which was subsequently fixed)
* However the un-skipping now caused some flakiness (a different flakiness, caused because I have significantly changed various server config settings - number of retries, quarantine interval, etc)
* Stared at the cicd1 secondary logs on the cicd1 VM and eventually figured out that when connect failed for a given atSign, _processNotificationQueue would not process any other atSign in its current run
* Fixed this so that when connect fails during _processNotificationQueue, then the notifications are re-enqueued for that atSign but the while loop moves on to processing the notifications for the next atSign, if any.

### Main PR description follows
**- What I did**
Removed all avoidable server-side latency under normal conditions when delivering notifications from one atSign to another. (Normal conditions means that both secondary servers are available, connectable and functioning normally)
Lower bound of latency inside the sending secondary server is now approximately 2.5 milliseconds. With regards to total end-to-end latency, for short encrypted messages, the combination of Client-1 encryption time, client-1-to-server-1 communication time, server-2-to-client-2 communication time and client-2 decryption ( excluding other network latency caused by e.g. speed-of-light), adds about another 2.5 milliseconds, giving an overall lower bound for end-to-end latency of approximately 5 milliseconds. 

**- How I did it**
_(All of the 'how I did it' info here comes from this PR's commits' messages. As this PR also contains some logging and de-linting changes, reviewers will likely find it easier to review the changed files one commit at a time)_

In the _at_persistence_secondary_server_ package:
  * in AtNotificationKeystore, publish the new notification event before persisting

In the _at_secondary_server_ package:
* notify_connection_pool.dart
  * added pool.init(size) method. init(size) is called by AtSecondaryImpl when it is starting.
  * pool.get(atSign): Removed redundant second call to _hasCapacity() on the underlying OutboundClientPool
* at_secondary_impl.dart
  * Removed some dead code
  * Moved initialization of the notifications ResourceManager to _after_ the initialization of the OutboundClientManager
  * Uses new init(size) method on the ResourceManager, which in turn calls the new init(size) method on NotifyConnectionPool
* notification_manager_impl.dart
  * NotificationManager_storeNotificationInQueue (called by the public NotificationManager.notify method) now
    1. calls queueManager.enqueue(notification) before persisting it
    2. nudges the notifications ResourceManager to let it know there is a new notification ready to send
* update_verb_handler.dart
  * queues the notification for delivery (via NotificationManager.notify()) before persisting it
  * de-linted
* resource_manager.dart
  * Enhanced the queue processing so that it still periodically runs on a timer, but also can be 'nudged' into action at any time (e.g. NotificationManager and UpdateVerbHandler as described above)
  * Added boolean guard _isProcessingQueue to ensure that the main queue processing loop, in the _processNotificationQueue() function, is not re-entrant
  * Added boolean _nudged so that when _processNotificationQueue() completes, it knows that it should immediately run again
  * Added new init(outboundConnectionsLimit) method (called by AtSecondaryImpl while **it** is starting) which in turns initializes the NotifyConnectionsPool with size _outboundConnectionsLimit
  * Removed this line of code from the main processing for a given atSign within _processNotificationQueue() : `AtNotificationMap.getInstance().removeWaitTimeEntry(atSign);` and moved it into _sendNotifications in order to fix a race-condition bug where
      * if a new notification for a given atSign was enqueued by the NotificationManager at the same time as _processNotificationQueue() was processing previously-queued notifications for **that same atSign** (the window for this to occur is provided by `var outboundClient = await _connect(atSign);` which can be very long)
      * then the next time _processNotificationQueue() would run, with no other notification having been enqueued for that atSign in the interim, it would find nothing to process because the atSign had been removed from the waitTimeMap
  * Changed _sendNotifications so it no longer closes the outbound connection after delivering all of the notifications for a given atSign. Keeping the connection open means that we avoid the (significant) latency added by creating and authenticating a new connection each time.
* at_notification_map.dart : Bug-fix: Made the empty list strongly typed in AtNotificationMap.remove() which was previously untyped and therefore being returned as a ListIterator\<dynamic\>, which was in turn causing problems elsewhere when casting to Iterator\<AtNotification\>
* Made OutboundClientPool usable as an LRU pool
  * Added lastUsed DateTime field to OutboundClient class; field is updated to DateTime.now() every time a method is called
  * Added removeLeastRecentlyUsed() method to OutboundClientPool. Sorts the clients in the pool by lastUsed, ascending
  * Modified NotifyConnectionPool to call removeLeastRecentlyUsed() on its OutboundClientPool if it is at capacity
  * Modified OutboundClientManager to call removeLeastRecentlyUsed() on its OutboundClientPool if it is at capacity
  * Added some tests for direct use of removeLeastRecentlyUsed() method
  * Added some tests on NotifyConnectionsPool to verify LRU behaviour of its underlying OutboundClientPool

**- How to verify it**
All existing and new (in this PR) tests should pass.

Removed all avoidable server-side latency under normal conditions when delivering notifications from one atSign to another.

### DIFF
--- a/at_secondary/at_secondary_server/lib/src/connection/base_connection.dart
+++ b/at_secondary/at_secondary_server/lib/src/connection/base_connection.dart
@@ -47,11 +47,23 @@ abstract class BaseConnection extends AtConnection {
       throw ConnectionInvalidException('Connection is invalid');
     }
     try {
+      logger.info('SENT: [${getMetaData().sessionID}] ${BaseConnection.truncateForLogging(data)}');
       getSocket().write(data);
       getMetaData().lastAccessed = DateTime.now().toUtc();
     } on Exception catch (e) {
       getMetaData().isStale = true;
       throw AtIOException(e.toString());
     }
+  }
+
+  static String truncateForLogging(String toLog, {int cutOffAfter = 1000}) {
+    if (toLog.length > cutOffAfter) {
+      toLog = '${toLog.substring(0, cutOffAfter)} [truncated, ${toLog.length - cutOffAfter} more chars]';
+    }
+    var lastNewLinePos = toLog.lastIndexOf("\n");
+    if (lastNewLinePos > -1) {
+      toLog = toLog.substring(0, lastNewLinePos);
+    }
+    return toLog;
   }
 }
--- a/at_secondary/at_secondary_server/lib/src/notification/at_notification_map.dart
+++ b/at_secondary/at_secondary_server/lib/src/notification/at_notification_map.dart
@@ -32,12 +32,12 @@ class AtNotificationMap {
     // If map is empty, or map doesn't contain the atSign, return an iterator for an empty list
     if (_notificationMap.isEmpty || !_notificationMap.containsKey(atSign)) {
       return 0;
-    } else {
-      Map<String, NotificationStrategy>? tempMap = _notificationMap[atSign]!; // can't be null, we've just checked containsKey
-      LatestNotifications latestList = tempMap['latest'] as LatestNotifications;
-      AllNotifications allList = tempMap['all'] as AllNotifications;
-      return latestList.length + allList.length;
     }
+
+    Map<String, NotificationStrategy>? tempMap = _notificationMap[atSign]!; // can't be null, we've just checked containsKey
+    LatestNotifications latestList = tempMap['latest'] as LatestNotifications;
+    AllNotifications allList = tempMap['all'] as AllNotifications;
+    return latestList.length + allList.length;
   }
 
   /// Returns the map of first N entries.
--- a/at_secondary/at_secondary_server/lib/src/notification/at_notification_map.dart
+++ b/at_secondary/at_secondary_server/lib/src/notification/at_notification_map.dart
@@ -13,15 +13,11 @@ class AtNotificationMap {
     return _singleton;
   }
 
-  final _notificationMap = <String?, Map<String, NotificationStrategy>>{};
-  final _waitTimeMap = <String?, NotificationWaitTime>{};
-  var _quarantineMap = <String?, DateTime>{};
+  final Map<String?, Map<String, NotificationStrategy>> _notificationMap = <String?, Map<String, NotificationStrategy>>{};
+  final Map<String?, NotificationWaitTime> _waitTimeMap = <String?, NotificationWaitTime>{};
+  final Map<String?, DateTime> _quarantineMap = <String?, DateTime>{};
 
   Map<String?, DateTime> get quarantineMap => _quarantineMap;
-
-  set quarantineMap(value) {
-    _quarantineMap = value;
-  }
 
   /// Adds the notifications to map where key is [AtNotification.toAtSign] and value is classes implementing [NotificationStrategy]
   void add(AtNotification atNotification) {
@@ -30,6 +26,18 @@ class AtNotificationMap {
     var notificationsMap = _notificationMap[atNotification.toAtSign]!;
     notificationsMap[atNotification.strategy!]!.add(atNotification);
     _computeWaitTime(atNotification);
+  }
+
+  int numQueued(String atSign) {
+    // If map is empty, or map doesn't contain the atSign, return an iterator for an empty list
+    if (_notificationMap.isEmpty || !_notificationMap.containsKey(atSign)) {
+      return 0;
+    } else {
+      Map<String, NotificationStrategy>? tempMap = _notificationMap[atSign]!; // can't be null, we've just checked containsKey
+      LatestNotifications latestList = tempMap['latest'] as LatestNotifications;
+      AllNotifications allList = tempMap['all'] as AllNotifications;
+      return latestList.length + allList.length;
+    }
   }
 
   /// Returns the map of first N entries.
--- a/at_secondary/at_secondary_server/lib/src/notification/at_notification_map.dart
+++ b/at_secondary/at_secondary_server/lib/src/notification/at_notification_map.dart
@@ -33,21 +33,19 @@ class AtNotificationMap {
   }
 
   /// Returns the map of first N entries.
-  Iterator<AtNotification> remove(String? atsign) {
-    // If map is empty, return empty map
-    if (_notificationMap.isEmpty) {
-      return [].iterator as Iterator<AtNotification>;
+  Iterator<AtNotification> remove(String? atSign) {
+    List<AtNotification> returnList;
+    // If map is empty, or map doesn't contain the atSign, return an iterator for an empty list
+    if (_notificationMap.isEmpty || !_notificationMap.containsKey(atSign)) {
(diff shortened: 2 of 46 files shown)


gkc:
@cconstab tagging you in your role as bug-finder general to find where this breaks ... branch name is straight_through_notify

gkc:
Original fix to the waitTimeMap race condition bug introduced a different (albeit very minor) bug. Latest commit fixes that bug also.

gkc:
Noticed a potential bug in ResourceManager._enqueueErrorList(). I think that the waitTimeEntry and quarantineEntry for a given atSign should only be cleared when *all* pending notifications have reached maximum retries, but right now it happens if *any* of them have reached maximum retries :
```dart
      if (atNotification.retryCount == maxRetries) {
        AtNotificationMap.getInstance()
            .removeWaitTimeEntry(atNotification.toAtSign);
        AtNotificationMap.getInstance()
            .removeQuarantineEntry(atNotification.toAtSign);
        logger.warning(
            'Failed to notify ${atNotification.id} from ${atNotification.fromAtSign} to ${atNotification.toAtSign}. Maximum retries ($maxRetries) reached');
        continue;
      }
```

Also - for long should we retry with notifications? I've changed the default settings in config.yaml and AtSecondaryConfig so that it's now ~300 seconds (10-second quarantine duration, maxRetries 30) i.e. ~5 minutes rather than ~1 minute as it was previously. But how long should we wait for an unavailable target secondary to become available again?

sitaram-kalluri (approved):

at_secondary/at_secondary_server/lib/src/connection/base_connection.dart
```diff
@@ -47,11 +47,23 @@ abstract class BaseConnection extends AtConnection {
       throw ConnectionInvalidException('Connection is invalid');
     }
     try {
+      logger.info('SENT: [${getMetaData().sessionID}] ${BaseConnection.truncateForLogging(data)}');
```
murali-shris, line 50: @gkc this may cause lot of noise in prod since we log on every socket write. e.g image data will cause lot of encoded chars to be logged

at_secondary/at_secondary_server/lib/src/notification/at_notification_map.dart
```diff
@@ -32,22 +28,32 @@ class AtNotificationMap {
...
   }
 
-  /// Returns the map of first N entries.
-  Iterator<AtNotification> remove(String? atsign) {
-    // If map is empty, return empty map
-    if (_notificationMap.isEmpty) {
-      return [].iterator as Iterator<AtNotification>;
+  int numQueued(String atSign) {
+    // If map is empty, or map doesn't contain the atSign, return an iterator for an empty list
+    if (_notificationMap.isEmpty || !_notificationMap.containsKey(atSign)) {
```
murali-shris, line 35: since we are returning from if block can we remove else block
e.g
if(_notificationMap.isEmpty || !_notificationMap.containsKey(atSign)) {
return 0;
}
// rest of the code

at_secondary/at_secondary_server/lib/src/notification/resource_manager.dart
```diff
@@ -11,57 +10,87 @@ import 'package:at_utils/at_logger.dart';
...
   ResourceManager._internal();
 
   factory ResourceManager.getInstance() {
     return _singleton;
   }
 
+  void init(int? outboundConnectionLimit) {
+    NotifyConnectionsPool.getInstance().init(outboundConnectionLimit);
+    _isStarted = true;
+    Future.delayed(Duration(milliseconds: 0)).then((value) {
```
murali-shris, line 31: if milliseconds is 0..is there a difference in calling _schedule directly vs calling inside Future.delayed(...) ?

at_secondary/at_secondary_server/lib/src/notification/at_notification_map.dart
gkc: Sounds good, will do. Makes it a little more readable

at_secondary/at_secondary_server/lib/src/notification/resource_manager.dart
gkc: Anecdotally, I believe I've observed a difference but I don't have hard proof. I think it's different in that the code as I've written it creates a new event for the Isolate's event queue

gkc:
@murali-shris commented on this pull request.

> In
> at_secondary/at_secondary_server/lib/src/connection/base_connection.dart
> <https://github.com/atsign-foundation/at_server/pull/664#discussion_r864571563>
> :
>
> > @@ -47,11 +47,23 @@ abstract class BaseConnection extends AtConnection {
>        throw ConnectionInvalidException('Connection is invalid');
>      }
>      try {
> +      logger.info('SENT: [${getMetaData().sessionID}] ${BaseConnection.truncateForLogging(data)}');
>
> @gkc <https://github.com/gkc> this may cause lot of noise in prod since
> we log on every socket write. e.g image data will cause lot of encoded
> chars to be logged

Yes, so I added the 'truncateForLogging" method to cut down on the amount of
noise. Having used this configuration for over a week now, I am convinced
that logging all @ protocol interactions (and very little else) at INFO
level makes the logs far more useful.

gkc:
Hi @murali-shris Before I merge to trunk, please can you do another review please, of the commits since May 5th?

PR #676, opened 2022-05-10
Throw specific exceptions from outbound_message_listener.dart

sitaram-kalluri:
**- What I did**
Throw Specific Exception from outbound_message_listener.dart instead of AtConnectException.

**- How I did it**
From the errorCode from the response get the appropriate instance of sub-class of AtException and return the exception.

NOTE: Un-commented dependency overrides section in at_secondary_server, at_functional_test and at_end2end_tests packages to enable the unit, functional tests and e2e tests to run successfully. Once the at_commons is merged and published, will replace the dependency_overrides with hosted version.


at_secondary/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
```diff
@@ -80,17 +81,22 @@ class OutboundMessageListener {
           // Right now, all callers of this method only expect there ever to be a 'data:' response.
           // So right now, the right thing to do here is to throw an exception.
           // We can leave the connection open since an 'error:' response indicates normal functioning on the other end
-          throw AtConnectException("Request to remote secondary ${outboundClient.toAtSign} at ${outboundClient.toHost}:${outboundClient.toPort} received error response '$result'");
+          result = result.toString().replaceAll('error:', '');
+          var errorCode = result.toString().split('-')[0];
+          var errorDescription = result.toString().split('-')[1];
+          throw AtExceptionUtils.get(errorCode)..message = errorDescription;
```
murali-shris: since we are assigning error description after getting the exact exception object.. we can throw
throw AtExceptionUtils.get(errorCode, errorDescription)..inside AtExceptionUtils create exception instance by passing errorDescription to constructor

at_secondary/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
```diff
@@ -80,17 +81,22 @@ class OutboundMessageListener {
           // Right now, all callers of this method only expect there ever to be a 'data:' response.
           // So right now, the right thing to do here is to throw an exception.
           // We can leave the connection open since an 'error:' response indicates normal functioning on the other end
-          throw AtConnectException("Request to remote secondary ${outboundClient.toAtSign} at ${outboundClient.toHost}:${outboundClient.toPort} received error response '$result'");
+          result = result.toString().replaceAll('error:', '');
```
murali-shris, line 86: I guess it is better to change the server error response from string type to json type..these lines to parse strings will be bug prone

at_secondary/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
sitaram-kalluri: Modified the get method to take errorCode and errorDescription.

at_secondary/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
sitaram-kalluri: Modified code to send error responses in JSON format

at_functional_test/test/commons.dart
```diff
@@ -103,8 +103,8 @@ Future<String> read({int maxWaitMilliSeconds = 5000}) async {
       result = _queue.removeFirst();
       // result from another secondary is either data or a @<atSign>@ denoting complete
       // of the handshake
-      if (result.startsWith('data:') ||
-          (result.startsWith('error:')) ||
+      if (result.contains('data') ||
+          (result.contains('error')) ||
```
gkc: Why do they no longer have the colons?

at_secondary/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
```diff
@@ -76,21 +77,25 @@ class OutboundMessageListener {
         if (result.startsWith('data:') ||
             (result.startsWith('@') && result.endsWith('@'))) {
           return result;
-        } else if (result.startsWith('error:')) {
+        } else if (result.contains('errorCode')) {
```
gkc: Are you sure there will never be any other situation other than the one you've thought about, in which a response can contain the string 'errorCode'? You could tighten this up by making the match more specific

at_functional_test/test/commons.dart
sitaram-kalluri: Changed and now it is again data:/error:

at_secondary/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
sitaram-kalluri: Added error: before the JSON response.

at_end2end_test/pubspec.yaml
```diff
@@ -15,7 +15,12 @@ dependencies:
   at_lookup: ^3.0.22
   at_client: ^3.0.20
 
-
+dependency_overrides:
+  at_commons:
+    git:
+      url: https://github.com/atsign-foundation/at_tools.git
+      path: at_commons
+      ref: exception_handling_changes
```
gkc, line 23: Please make a note at the top of your PR noting the dependency override and the need, once the new at_commons is available, to remove this dependency override.

at_functional_test/pubspec.yaml
```diff
@@ -17,7 +17,12 @@ dependencies:
   at_client: ^3.0.9
   
 
-# dependency_overrides:
+dependency_overrides:
+  at_commons:
+    git:
+      url: https://github.com/atsign-foundation/at_tools.git
+      path: at_commons
+      ref: exception_handling_changes
```
gkc, line 25: same comment as earlier

at_functional_test/test/commons.dart
```diff
@@ -103,8 +103,8 @@ Future<String> read({int maxWaitMilliSeconds = 5000}) async {
       result = _queue.removeFirst();
       // result from another secondary is either data or a @<atSign>@ denoting complete
       // of the handshake
-      if (result.startsWith('data:') ||
-          (result.startsWith('error:')) ||
+      if (result.contains('data:') ||
+          (result.contains('error:')) ||
```
gkc: I still don't understand this. Why 'contains' instead of 'startsWith'?

at_secondary/at_secondary_server/pubspec.yaml
```diff
@@ -20,12 +20,12 @@ dependencies:
...
-#    git:
-#      url: https://github.com/atsign-foundation/at_tools.git
-#      path: at_commons
-#      ref: trunk
+dependency_overrides:
+  at_commons:
+    git:
+      url: https://github.com/atsign-foundation/at_tools.git
+      path: at_commons
+      ref: exception_handling_changes
```
gkc, line 28: Same comment as earlier

gkc (changes requested):

at_functional_test/test/commons.dart
sitaram-kalluri: This i missed changing back to startsWith. Reverted back to startsWith and pushed the change.

at_end2end_test/pubspec.yaml
sitaram-kalluri: Added a Note on why dependency-overrides is un-commented.

at_functional_test/test/config_verb_test.dart
```diff
@@ -79,15 +82,21 @@ void main() {
     await socket_writer(socketFirstAtsign!, 'config:block:add:@@kevin');
     var response = await read();
     print('config verb response : $response');
-    assert(response.contains('error:AT0003-Invalid syntax'));
+    response = response.replaceFirst('error:', '');
+    var decodedResponse = jsonDecode(response);
+    expect(decodedResponse['errorCode'], 'AT0003');
+    expect(decodedResponse['errorDescription'], contains('Invalid syntax'));
```
gkc, line 88: Won't this (and other similar tests) fail against current prod version of the server? Perhaps you need to assert that either the old check response.contains(...) or the new check (decodedResponse.errorCode and errorDescription) succeeds?

at_secondary/at_secondary_server/lib/src/exception/global_exception_handler.dart
```diff
@@ -151,6 +152,9 @@ class GlobalExceptionHandler {
 
   void _writeToSocket(AtConnection atConnection, String prompt,
       String? errorCode, String errorDescription) {
-    atConnection.write('error:$errorCode-$errorDescription\n$prompt');
+    atConnection.write('error:${jsonEncode({
+          'errorCode': errorCode,
+          'errorDescription': errorDescription
+        })}\n$prompt');
```
gkc, line 158: NB: We can't deploy this to production until all clients have code which can cope with either the current error response or this new error response, or we have code on the server-side which can detect what capability the client has. @VJag should we consider having the client be able to send some info telling the server what version of the at_client it is running?

gkc: NB: We can't deploy this to production until all clients have code which can cope with either the current error response or this new error response, or we have code on the server-side which can detect what capability the client has. @VJag should we consider having the client be able to send some info telling the server what version of the at_client it is running?

at_functional_test/test/config_verb_test.dart
gkc: Won't this (and other similar tests) fail against current prod version of the server? Perhaps you need to assert that either the old check response.contains(...) or the new check (decodedResponse.errorCode and errorDescription) succeeds?

gkc (changes requested):

at_functional_test/test/config_verb_test.dart
sitaram-kalluri: The functional test run against the feature branch docker image. Hence we will always have error responses in JSON format. The test against the prod server are in end2end tests where we have the IF conditions

at_secondary/at_secondary_server/lib/src/exception/global_exception_handler.dart
sitaram-kalluri: Sure Gary, will ensure the code goes to production only after the client packages are published.

gkc: It's not just about packages being published. It's about every existing client in an installed app continuing to work. Let's figure out how to make that possible

sitaram-kalluri:
The change related to json encoding the error responses are deferred to next releases. Hence closing the PR.

PR #677, opened 2022-05-10
feat: Bypass cache test

purnimavenkatasubbu:
**- What I did**
Added e2e tests for bypass_cache

**- How I did it**

**- How to verify it**
bypass_cache test should pass in the Github actions


at_end2end_test/test/cached_key_test.dart
```diff
@@ -126,6 +126,62 @@ void main() {
...
+    assert(
+        (!response.contains('Invalid syntax')) && (!response.contains('null')));
+
+    ///lookup verb alice  atsign_2
+    await sh2.writeCommand('lookup:username$atSign_1');
+    response = await sh2.read();
+    print('lookup verb response : $response');
+    expect(response, contains('data: $value'));
+
+    // TODO
```
cpswan: This can't simply be in the code flow, as it has to be in the GitHub Actions workflow. So the test needs to run up to this point, and then the action SSHs into the CICD VM and stops the secondary and then we start running another test that picks up from there etc.? @kalluriramkumar should be able to help, as he's familiar with the workflow definition.

Also how long do we expect these tests to take? If we're waiting around for minutes then that probably blows the 10m target for our commit cycle.

cpswan (changes requested):

at_secondary/at_secondary_server/config/config.yaml
```diff
@@ -132,4 +132,4 @@ sync:
 #set to true in testing mode, false by default
 #IMPORTANT NOTE : please set testingMode to true only if you know what you're doing. Set to false when not testing
 testing:
-  testingMode: false
+  testingMode: true
```
gkc, line 135: This should always be false in config.yaml, because this file is what gets sent to prod. Should only ever be set to true by injecting into docker container env via the github workflow

at_end2end_test/test/cached_key_test.dart
```diff
@@ -132,6 +132,70 @@ void main() {
...
+
+    /// lookup with bypass_cache set to false
+    /// should return the old value
+    await sh2.writeCommand('lookup:bypassCache:false:fav-city$atSign_1');
+    response = await sh2.read();
+    print('lookup verb response : $response');
+    expect(response, contains('data: $oldValue'));
+
+    // reset the autoNotify to default
+    await sh1.writeCommand('config:reset:autoNotify');
```
gkc: This should be in a finally {} block so that even if the test assertions fail, we still set autoNotify back to what it should be

gkc (changes requested):

at_end2end_test/test/cached_key_test.dart
gkc: The changing of config is now done via the config verb (available only when the server is in testingMode), so this comment doesn't apply any more.

at_secondary/at_secondary_server/config/config.yaml
gkc: This should always be false in config.yaml, because this file is what gets sent to prod. Should only ever be set to true by injecting into docker container env via the github workflow

Given that this came up in another situation also, I'd like there to be a test which asserts that the config.yaml never contains "testingMode: false"

gkc (changes requested):

gkc (changes requested):
couple of changes needed

at_end2end_test/test/cached_key_test.dart
purnimavenkatasubbu: Added the code snippet to the finally block.

at_secondary/at_secondary_server/config/config.yaml
purnimavenkatasubbu: Reverted the "testingMode: false". To make sure this PR is only for bypass cache feature. I'll include the test for testingMode in a different branch.

sitaram-kalluri (approved):

gkc (approved):

at_end2end_test/test/cached_key_test.dart
gkc: @cpswan can you re-review please so Purnima can merge if you're happy?

cpswan (approved):

PR #681, opened 2022-05-12
Fix flaking cached_key_test.dart

gkc:
**- What I did**
Fix flaking cached_key_test.dart

**- How I did it**
Random number range 1..10 is not enough to prevent flaking due to concurrent test runs. Changed from Random().nextInt(10) to DateTime.now().millisecondsSinceEpoch, and updated the tests which make use of it to add their own uniqueness on top of the millisecondsSinceEpoch

**- How to verify it**
Tests pass. This particular test stops flaking

Fix flaking cached_key_test.dart


xavierchanth (approved):
LGTM

PR #698, opened 2022-05-19
Revert from dart 2.17.0 to dart 2.16.2

gkc:
Revert from dart 2.17.0 to dart 2.16.2 as we are seeing some weird TLS errors on the canary secondaries resulting in server crash


cconstab (approved):

cpswan:
It might be worth checking the Dart 2.17.1 release notes, though when I checked this morning it hadn't got to Docker Hub

cpswan (approved):

PR #703, opened 2022-05-20
added unit tests for resource manager

### LINKED ISSUE #633: Implement unit tests to prevent regression of the notifications bug fixed by #621
**Describe the bug**
See #621 for full detail. TL;DR, that pull request fixed this issue:
<blockquote>
In [this method](https://github.com/atsign-foundation/at_server/blob/883623b080b5247f5b6528edad88fe112d8639ef/at_secondary/at_secondary_server/lib/src/notification/resource_manager.dart#L90)

* if an exception happens while sending a notification
* and the iterator has other notifications not yet processed
* then those other notifications are not added to the errorList
* and therefore they are not re-enqueued
</blockquote>

**Expected behaviour**
The bug has been fixed, but we need a set of unit tests which cover this case, to prevent regression in future. The Notification ResourceManager is complex but we should be able to implement the required unit tests by allowing dependencies to be injected into ResourceManager, and injecting mocks of those dependencies at test time.

**Additional context**
As mentioned above, the Notifications ResourceManager has complex behaviour and lots of complex dependencies. This is likely 2-3 days of work to cover just this particular scenario - but that work will also provide the foundation that will allow us to cover all of the Notifications ResourceManager's error handling with unit tests - which is much to be desired, as notifications need to be rock solid.

purnimavenkatasubbu:
**- What I did**
Added unit tests for the resource_manager

**- How I did it**
Created a mock object for an outbound client
Passed a notification iterator,  atsign, and mock outbound client to sendNotifications()
Verified whether the dequeue returns the same 3 notifications

NOTE: Added @visibleForTesting annotation to make the code testable.

**- How to verify it**
Unit tests should pass in the Github Actions

### DIFF
--- a/at_secondary/at_secondary_server/test/resource_manager_test.dart
+++ b/at_secondary/at_secondary_server/test/resource_manager_test.dart
@@ -12,10 +12,12 @@ import 'package:mocktail/mocktail.dart';
 class MockOutboundClient extends Mock implements OutboundClient {}
 
 void main() async {
+  // mock object for outbound client
   OutboundClient mockOutboundClient = MockOutboundClient();
   ResourceManager rm = ResourceManager.getInstance();
   var storageDir = Directory.current.path + '/test/hive';
 
+  //  forcing the notification sending to fail with an exception
   setUp(() {
     reset(mockOutboundClient);
     when(() => mockOutboundClient.notify(any())).thenAnswer((_) async {
@@ -23,6 +25,10 @@ void main() async {
     });
   });
 
+  /// Purpose of the test is to do the following
+  /// 1) Passing an notification iterator having valid & invalid notifications.
+  /// 2) Verifying that all of the queued notifications are still enqueued when one of the notification is invalid.
+  /// 3) dequeue() should return the same 3 notifications as the one's we passed
   group('A group of notify verb test', () {
     setUp(() async => await setUpFunc(storageDir));
     test('Test send notifications', () async {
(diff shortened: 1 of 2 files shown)


at_secondary/at_secondary_server/test/resource_manager_test.dart
```diff
@@ -0,0 +1,158 @@
...
+      await rm.sendNotifications(
+          atsign, mockOutboundClient, notificationIterator);
+      var atNotificationList = [];
+      var itr = QueueManager.getInstance().dequeue(atsign);
+      while (itr.moveNext()) {
+        atNotificationList.add(itr.current);
+      }
+      expect(atNotificationList[0].id, '121');
+      expect(atNotificationList[1].id, '122');
+      expect(atNotificationList[2].id, '123');
```
gkc, line 72: Please add a code comment with some explanation of the meaning of the test - i.e. what behaviour are we verifying, and how are we making it happen

gkc, line 82: Please add a code comment with some explanation of the meaning of the test - i.e. what behaviour are we verifying, and how are we making it happen

gkc: Same comment for all of the 'expect' assertions in the tests - please add brief code comments explaining what behaviour is being checked and why it should be happening

gkc (changes requested):
Nice work! Please just add some code comments near the expect calls explaining what behaviour is being checked and how the test setup is making that behaviour happen. e.g. "We're forcing the notification sending to fail with an exception, and we want to verify that all of the queued notifications are still enqueued when that happens."

at_secondary/at_secondary_server/test/resource_manager_test.dart
purnimavenkatasubbu: Added comments for the 'expect'

gkc (approved):
Thank you @purnimavenkatasubbu

PR #707, opened 2022-05-20
Fix HandshakeException handling

gkc:
**- What I did**
* Fixed issue where server would terminate upon a HandshakeException from ServerSocket.listen's onError handler

**- How I did it**
* GlobalExceptionHandler.handle(...) adjusted to match the new exception hierarchy
* Our onError() callback for ServerSocket.listen no longer throws an exception but just logs a warning, as there's no useful action to take
* Fixed the code in GlobalExceptionHandler so if we ever do need to terminate the secondary, then it shuts down cleanly by calling stop() on the secondary

**- How to verify it**
Tests should pass

**- Additional context**
* When this problem manifested after 3.0.16 was released, why was @cconstab's secondary crashing apparently on demand? Because he was running a version of a client app which had old code in it which would trigger a HandshakeException in the secondary every single time the app started up and connected to the secondary. So Colin's secondaries were crashing by far the most often. However HandshakeExceptions can and do happen for a whole bunch of reasons, hence a much lower but non-zero rate of server terminations in other secondaries.
* Bug was caused by change in at_commons which made AtServerException a super-class of multiple other server-side exceptions, including InternalServerError; ServerSocket.listen's onError handler was throwing InternalServerError - this was perfectly fine when InternalServerError was just a subclass of AtException but it's now a subclass of AtServerException; in turn, the GlobalExceptionHandler was treating AtServerExceptions, which are only directly created and thrown by unrecoverable exceptions during startup, as fatal, and terminating the secondary - even though there is no code path where an actual AtServerException can make its way to the GlobalExceptionHandler (although now that it's a superclass, its subclasses can)

Fix HandshakeException handling

### DIFF
--- a/at_secondary/at_secondary_server/lib/src/exception/global_exception_handler.dart
+++ b/at_secondary/at_secondary_server/lib/src/exception/global_exception_handler.dart
@@ -72,16 +72,21 @@ class GlobalExceptionHandler {
       logger.warning(exception.toString());
       await _sendResponseForException(exception, atConnection);
 
-    } else if (exception is AtServerException ||
-        exception is ArgParserException) {
-      // In case of AtServerException terminate the server
+    } else if (exception is ArgParserException) {
+      // TODO [gkc] I suspect this code block is not reachable. Verify and delete
       logger.shout("Terminating secondary due to ${exception.toString()}");
-      _terminateSecondary();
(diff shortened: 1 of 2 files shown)


cconstab (approved):
Simple but effective

PR #708, opened 2022-05-21
GitHub action workflow for root server deployment

kumarnarendra701:


.github/workflows/at_server_dev_deploy.yaml
```diff
@@ -0,0 +1,62 @@
+name: root_server
+on:
+  push:
+    branches:
+      - trunk
```
cpswan, line 5: Can we filter on just changes to the at_root code?

.github/workflows/at_server_dev_deploy.yaml
```diff
@@ -0,0 +1,62 @@
...
+jobs:
+  Docker_Build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
cpswan: Should be 3.0.2 (and other actions should have full latest version pinned)

.github/workflows/at_server_prod_deploy.yaml
```diff
@@ -0,0 +1,62 @@
+name: root_server
+on:
+  push:
+    tags:
+      - 'v*.*.*'
```
cpswan: We're going to need to use different tags, as v is used for secondary versions - I'd suggest r

.github/workflows/at_server_dev_deploy.yaml
```diff
@@ -0,0 +1,62 @@
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      
+      # Build the Docker image for Dev
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          file: at_root/at_root_server/Dockerfile
+          context: at_root/at_root_server
+          push: true
+          tags: atsigncompany/root:dev_env-${{ env.BRANCH }}-gha${{ github.run_number }}
```
cpswan: can we also have a :dev-env tag that acts as the latest

.github/workflows/at_server_prod_deploy.yaml
```diff
@@ -0,0 +1,62 @@
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      
+      # Build the Docker image for Dev
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          file: at_root/at_root_server/Dockerfile
+          context: at_root/at_root_server
+          push: true
+          tags: atsigncompany/root:prod-${{ env.BRANCH }}-gha${{ github.run_number }}
```
cpswan: there should also be a :prod tag for the latest prod release

cpswan (changes requested):
I was expecting these workflows to go into at_server.yaml, though I guess they can run standalone.

.github/workflows/at_server_dev_deploy.yaml
kumarnarendra701: Fixed

.github/workflows/at_server_prod_deploy.yaml
kumarnarendra701: Fixed

.github/workflows/at_server_prod_deploy.yaml
kumarnarendra701: Fixed

.github/workflows/at_server_dev_deploy.yaml
kumarnarendra701: fixed

.github/workflows/at_server_dev_deploy.yaml
kumarnarendra701: I have changed it to at_root.

cpswan: I don't think changing the `name:` achieves what I'd like to see here. The `on:` parameter needs to be much more selective.

cpswan (changes requested):

.github/workflows/at_server_dev_deploy.yaml
```diff
@@ -2,7 +2,7 @@ name: at_root
 on:
   push:
     branches:
-      - trunk
+      - ci-cd-test
```
cpswan: I think trunk still needs to be the branch we use, as `ci-cd-test` will go once this PR is merged.

cpswan (changes requested):

.github/workflows/at_server_dev_deploy.yaml
kumarnarendra701: This is not a final request for review, just added the ci-cd-test branch to test it.

.github/workflows/at_server_dev_deploy.yaml
cpswan: I'm not seeing that fix

kumarnarendra701: Earlier added to deploy workflow only, Now fixed for all YAML files.

cpswan (approved):
There are a few files here with no changes of substance, but I don't see any harm.

PR #723, opened 2022-05-31
Fix issue in returning lastSyncedEntry

### LINKED ISSUE #721: Server stuck on old value even though syncing is happening
Steps to reproduce:
https://github.com/atsign-foundation/at_server/issues/709#issuecomment-1140223112

sitaram-kalluri:
**- What I did**
1. Sort the commit entries to return commit entry with maximum commit Id from lastSyncedEntry method.
2. Introduce _lastSyncedEntryCacheMap to optimize the lastSyncedEntry method. Without having to sort the commitlog keystore values everytime.

**- How I did it**
1. Added sort method.

**- How to verify it**


gkc (changes requested):
1. Please add comments for future maintainers explaining purpose of lastSyncedEntry and that it is only used by clients which have local secondary key stores, but not by the secondary server
2. Question - Seems like the current code means that this new map will be maintained on the server-side also. But this isn't needed, right?

sitaram-kalluri:
> 1. Please add comments for future maintainers explaining purpose of lastSyncedEntry and that it is only used by clients which have local secondary key stores, but not by the secondary server
> 
>     2. Question - Seems like the current code means that this new map will be maintained on the server-side also. But this isn't needed, right?

Sure Gary, will update the code comments. Also, the update method and lastSyncedEntry method are used by the at_client SDK only. So yes, it is not needed on server side.

gkc:
Thanks @kalluriramkumar 

@VJag do we need to start thinking about breaking at_persistence_secondary_server into three packages (core, client additions, server additions) ? At a minimum I think we need some restructuring to make it clear where there is client-only or server-only code

VJag:
I agree. To start with should we introduce annotations like @client @server and annotate the code?

gkc (approved):
LGTM although I would like to see some comments explaining that the update() is used only client-side

sitaram-kalluri:
> LGTM although I would like to see some comments explaining that the update() is used only client-side

Update the code comments in the following [commit](https://github.com/atsign-foundation/at_server/pull/723/commits/26c0a2fc284855a5d883480ef8ddb6e4387fbe5a)

PR #724, opened 2022-05-31
feat: config verb support to dynamically change server config

### LINKED ISSUE #588: Dynamically adjust server config to facilitate testing
We need ability - intended for use in test environments **_only_** - to dynamically adjust server config from an authenticated client. See https://github.com/atsign-foundation/at_server/issues/542 for a specific example of a test that is currently effectively impossible to implement

Implementation could be via a new verb, 'set:param1=val1,param2=val2,param3=val3' where parameter could be server-wide (possibly indicated by prefix of 'server.' to param name) or for just this connection (possibly indicated by prefix of 'connection.' to param name). In server code, request parameters could be handled by having the config verb handler 'store' those parameters on the connection handler, then individual verb handlers could pick up those params when they are executing.

Initial scope :
* Add a verb (e.g. 'set') which requires user authentication to execute
* For each specific config parameter required by https://github.com/atsign-foundation/at_server/issues/542 (inbound_connection_limit, compactionJobRun)
  * Find its usages and ensure that if it is changed then the changed value will be used
  * Add explicit support to the verb handler for just that config parameter
* Add a server config parameter which by default has the 'set' verb disabled for use; and let's specifically enable it for the testing servers

srieteja:
closes https://github.com/atsign-foundation/at_server/issues/588
**- What I did**

added support to change server config during runtime

**- How I did it**

- added option through config verb to set new values to existing config parameters
- added config variable name "testingMode" in at_sec_server config.yaml which shall be set to true when testing and to enable this feature
- added methods in AtSecondaryConfig that can be used to subscribe and broadcast config changes to wherever required
- explicit support to modify size parameter in inbound_connection_pool through inbound_connection_manager
- added new init parameter "isColdInit" to inbound_connection_pool which shall be ignored while server start, but shall be set to false in other cases(work-around to modify inbound_connection_limit)

**- How to verify it**
tests yet to be written


at_secondary/at_secondary_server/lib/src/connection/inbound/inbound_connection_manager.dart
```diff
@@ -49,9 +52,9 @@ class InboundConnectionManager implements AtConnectionFactory {
 
   /// Initialises inbound client pool with a given size.
   /// @param - size - Maximum clients the pool can hold
-  void init(int size) {
+  void init(int size, {bool isColdInit = true}) {
```
VJag, line 55: Please add dart docs explaining the purpose of the variable "isColdInit"

at_secondary/at_secondary_server/lib/src/server/at_secondary_impl.dart
```diff
@@ -153,6 +153,25 @@ class AtSecondaryServerImpl implements AtSecondaryServer {
     await commitLogCompactionJobInstance
         .scheduleCompactionJob(atCommitLogCompactionConfig);
 
+    if (AtSecondaryConfig.testingMode!) {
```
VJag: Move all of these code blocks to separate methods ex: addConfigChangeListeners(); 

addConfigChangeListeners() {
     addCompactionConfigChangeListener();
     .. others
}

It is not good to have more than 50 to 80 lines in a single method for readability

at_secondary/at_secondary_server/lib/src/verb/handler/config_verb_handler.dart
```diff
@@ -83,6 +83,43 @@ class ConfigVerbHandler extends AbstractVerbHandler {
           result = 'unknown operation';
           break;
       }
+
+      switch (setOperation) {
+        case 'set':
+          if (AtSecondaryConfig.testingMode &&
```
VJag: @srieteja we have to check with Gary if he is fine with making "testingMode" part of the config

at_secondary/at_secondary_server/lib/src/verb/handler/config_verb_handler.dart
```diff
@@ -83,6 +83,43 @@ class ConfigVerbHandler extends AbstractVerbHandler {
...
+                ModifiableConfigs.values.contains(verbParams[CONFIG_NAME])) {
+              result = AtSecondaryConfig.getLatestConfigValue(
+                  ModifiableConfigs.values.byName(verbParams[CONFIG_NAME]!));
+            } else {
+              result = 'null';
+            }
+          }
+          break;
+        case 'print':
+          result = 'print operation';
```
VJag: Is this work in progress?

VJag (dismissed):
Hi Teja,

Overall, very well done. Few minor comments. Please take a look.

+ Please add some functional tests.

at_secondary/at_secondary_server/lib/src/verb/handler/config_verb_handler.dart
gkc: Yes I'm good with this. It should default to 'false' of course, and we will need to inject it as 'true' into the docker env for the secondaries we use in testing

at_secondary/at_secondary_server/lib/src/verb/handler/config_verb_handler.dart
srieteja: i might have missed pushing the latest changes sir. I will push the print implementation along with the review changes suggested.

at_secondary/at_secondary_server/config/config.yaml
```diff
@@ -123,4 +123,9 @@ sync:
...
   bufferSize: 5242880
   # Represents the numbers of keys to send in each batch.
-  pageLimit: 25
\ No newline at end of file
+  pageLimit: 25
+
+#set to true in testing mode, false by default
+#IMPORTANT NOTE : please set testingMode to true only if you know what you're doing. Set to false when not testing
+testing:
+  testingMode: true
```
gkc: It should also be false here - this is the file we push to all secondary servers.

at_secondary/at_secondary_server/lib/src/server/at_secondary_config.dart
```diff
@@ -114,6 +118,14 @@ class AtSecondaryConfig {
     }
   }
 
+  static bool get testingMode {
+    var result = _getBoolEnvVar('testingMode');
```
gkc, line 122: Note that this (environment) is how we will get the cicd secondaries to have testingMode set to true :  do `docker service update --force --env-add testingMode=true` on the cicd secondary containers

gkc (changes requested):

at_functional_test/test/all_verbs_test.dart
```diff
@@ -100,6 +99,37 @@ void main() async {
...
+//     print('config verb response $response');
+//     expect(response, contains('data:ok'));
+//
+//     await socket_writer(
+//         socketFirstAtsign!, 'config:print:maxNotificationRetries');
+//     await Future.delayed(Duration(seconds: 2));
+//     response = await read();
+//     print('config verb response $response');
+//     expect(response, contains('data:25'));
+//   });
```
gkc: Please uncomment these tests. When people are running tests they need to set the env for their container. Perhaps add the documentation as to how to do this. Perhaps you can assert that testingMode is true on the server somehow (?) and if it's not then show a message describing what you need to do to the container you are running tests against.

gkc (changes requested):

at_functional_test/test/all_verbs_test.dart
srieteja: I have enabled the tests and am asserting that testingMode is true.

gkc (approved):

at_secondary/at_secondary_server/pubspec.yaml
```diff
@@ -17,16 +17,26 @@ dependencies:
   crypton: ^2.0.1
   collection: ^1.15.0
   basic_utils: ^3.0.2
-  at_persistence_secondary_server: ^3.0.28
+  at_persistence_secondary_server: ^3.0.29
   at_lookup: ^3.0.25
```
murali-shris: at_lookup should be 3.0.26

srieteja: upgraded

murali-shris (approved):

PR #734, opened 2022-06-02
Replace null commitIds with hive internal keys

### LINKED ISSUE #719: get latest commit id - handle null commit ids
Better defensive code to deal with a null commit id in logic that keeps track of the latest commit id

    
     - Logic that keeps track of the latest server commit id cannot skip the value with a null commit id. This would cause the system to report the wrong value for the latest commit id. Potentially we can make of internal key and try to return the right value.

Snippet from _getCommitIdMap method of  CommitLogKeyStore:

     if (value.commitId == null) {
        _logger.severe(
            'CommitID is null for ${value.atKey}. Skipping to update entry into commitLogCacheMap');
        continue;
      }

sitaram-kalluri:
**- What I did**
1. On Server start, replace null commit id's in commit log keystore with hive internal key's

2. Introduced NullCommitEntry class which can be returned in places where null has to be returned when a matching commitEntry is not found.

3. In pubspec.yaml added meta package under dev_dependencies section to add @visibleForTesting annotation.

**- How I did it**
1. Added repairCommitLog method which set''s the commitId with hive internal key if commit is null.

2. In lastSyncedEntry method returns when a CommitEntry with maximum commit id or returns NullCommitEntry


VJag (approved):

at_secondary/at_persistence_secondary_server/lib/src/compaction/commit_log_compaction_service.dart
```diff
@@ -24,7 +24,7 @@ class CommitLogCompactionService implements AtChangeEventListener {
             keysToCompactCount = keysToCompactCount + 1;
           }
           _commitLogEntriesMap.putIfAbsent(
-              commitEntry.atKey, () => CompactionSortedList());
+              commitEntry.atKey!, () => CompactionSortedList());
```
gkc, line 27: Why is the bang needed here on commitEntry.atKey?

at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
```diff
@@ -39,10 +40,16 @@ class CommitLogKeyStore
     var lastCommittedSequenceNum = lastCommittedSequenceNumber();
     _logger.finer('last committed sequence: $lastCommittedSequenceNum');
 
-    // Cache the latest commitId of each key.
-    // Add entries to commitLogCacheMap when initialized from at_secondary_server
-    // and refrain for at_client_sdk.
+    // Ensures the below code runs only when initialized from secondary server.
+    // enableCommitId is set to true in secondary server and to false in client SDK.
     if (enableCommitId) {
```
gkc, line 45: Maybe now is a good time to rename 'enableCommitid' to 'isServer'? Since they are synonymous at the moment, and 'isServer' is more explanatory

gkc (changes requested):

at_secondary/at_persistence_secondary_server/lib/src/compaction/commit_log_compaction_service.dart
sitaram-kalluri: I have type casted the toMap method in commitLogKeystore to Future<Map<int, CommitEntry>> from Future<Map<dyamic,dynamic>> 

Since the in the CommitEntry class, we have atkey defined as nullable - `final String? _atKey`, added a bang operator here. 
Further we will not get into a scenario of atKey going null since we have checks for them in at_client side and at_secondary_server

at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
sitaram-kalluri: Hi Gary, As discussed in the call, leaving enableCommitId as is since it might effect the backward compatibility.

gkc (approved):

PR #754, opened 2022-06-15
fix: List.lastWhere does not return null and update regex

### LINKED ISSUE #755: List.lastWhere does not return null and lastSyncedEntryCache does not update on reserved keys
**Describe the bug**
1. With the new dart version, the `lastWhere` clause does not allow to return `null`.
2. The lastSyncedEntryCacheMap does not update the reserved keys(encryption public key) commit id into cache.

**Expected behavior**
1. When stats:3 is executed with regex the `lastCommittedSequenceNumberWithRegex` method should return the commitId.
2. The lastSyncedEntryCacheMap should cache the reserved keys commit id also.

sitaram-kalluri:
**- What I did**
1. In the latest Dart version, List.lastWhere clause throws error when returning null. Hence modified code to return `NullCommitEntry`

2. Modified code related lastSyncedEntryCacheMap to accept internal keys.


VJag (approved):
LGTM

PR #817, opened 2022-07-20
feat: added workflow step to deploy the canary image to prod

kumarnarendra701:
**- Added workflow step to deploy the canary image to prod**

### DIFF
--- a/.github/workflows/at_server.yaml
+++ b/.github/workflows/at_server.yaml
@@ -782,8 +782,8 @@ jobs:
             --amend atsigncompany/secondary:canary-${{ env.VERSION }}-amd64
           docker manifest push atsigncompany/secondary:canary-${{ env.VERSION }}  
 
+  # Deploy canary image to secondary image
   deploy_canary_image_to_prod_secondary_image:
-    # Runs only after functional tests are completed.
     if: ${{ github.event.inputs.deploy_canary_image_to_prod_secondary_image == 'true' }}
     env:
       working-directory: at_server
--- a/.github/workflows/at_server.yaml
+++ b/.github/workflows/at_server.yaml
@@ -3,7 +3,22 @@ name: at_server
 # 1. on pull request raised to trunk branch.
 # 2. on push event to trunk branch.
 # 3. on tagging a release
+# 4. on manual workflow(Run job manually)
 on:
+  workflow_dispatch:
+    inputs:
+      rollback_prod_secondary_image:
+        type: boolean
+        default: true
+        description: Rollback Prod secondary image
+      rollback_canary_secondary_image:
+        type: boolean
+        default: true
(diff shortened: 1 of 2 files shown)


at_secondary/Dockerfile.canary_to_prod
```diff
@@ -0,0 +1,5 @@
+FROM atsigncompany/secondary:canary
+ENV HOMEDIR=/atsign
+RUN \
+  cp -rpf pubspec.yaml $HOMEDIR/; \
+  chown -R atsign:atsign $HOMEDIR ;
```
cpswan, line 5: I know that we have ENTRYPOINT defined in the underlying image, but do we need it defined here too?

cpswan (approved):
Generally LGTM. Just one question on ENTRYPOINT before we merge.

Any thoughts on how we test this?

cpswan:
@gkc and @kalluriramkumar I've added you as reviewers as I'd like your eyes on this before we merge.

sitaram-kalluri (approved):

.github/workflows/at_server.yaml
```diff
@@ -711,6 +726,125 @@ jobs:
...
+  # Rollback prod secondary image.
+  rollback_prod_secondary_image:
+    if: ${{ github.event.inputs.rollback_prod_secondary_image == 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.0.2
+      
+      # Extract version for docker tag
+      - name: Get version
+        run: echo "VERSION=${GITHUB_REF##*/}" >> $GITHUB_ENV
```
gkc, line 739: Am I correct that when rolling back, we will run this job from the branch we want to rollback **to**?

.github/workflows/at_server.yaml
```diff
@@ -711,6 +726,125 @@ jobs:
...
+    env:
+      working-directory: at_server
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3.0.2
+
+      # Extract version for docker tag
+      - name: Get version
+        run: echo "VERSION=${GITHUB_REF##*/}" >> $GITHUB_ENV
```
gkc, line 820: Which branch would we run this workflow on?

gkc (commented):
I have asked a couple of questions to clarify my understanding of how this all works, but LGTM

.github/workflows/at_server.yaml
cpswan: It has nothing to do with branches here, as it's retagging Docker images.

.github/workflows/at_server.yaml
cpswan: When running it we pick a tag for a previous release.

PR #830, opened 2022-07-27
fix: invalidate commit log cache

murali-shris:
**- What I did**
- Invalidated commit log cache when an entry is removed from commit log
**- How I did it**
- modified implementation of commit log key store --> remove method. if commit entry of a passed commit id is not null, remove the key from cache
**- How to verify it**
- run the group of unit tests in commit log test -  A group of tests to verify commit log cache map

### DIFF
--- a/at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
+++ b/at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
@@ -135,7 +135,8 @@ class CommitLogKeyStore
       if (commitEntry != null) {
         final key = commitEntry.atKey;
         final value = _commitLogCacheMap.remove(key);
-        _logger.finest('removed value: $value');
+        _logger.finest(
+            'removed key : ${commitEntry.atKey} from commit log. value: $value');
       }
     } on Exception catch (e) {
       throw DataStoreException('Exception deleting entry:${e.toString()}');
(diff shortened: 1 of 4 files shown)


at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
```diff
@@ -129,7 +129,14 @@ class CommitLogKeyStore
...
   Future<void> remove(int commitId) async {
     try {
+      final commitEntry = (await _getBox() as Box).get(commitId);
       await _getBox().delete(commitId);
+      // invalidate cache for the removed entry
+      if (commitEntry != null) {
+        final key = commitEntry.atKey;
+        final value = _commitLogCacheMap.remove(key);
+        _logger.finest('removed value: $value');
+      }
```
VJag, line 139: We can reduce the temp variables here:

 if (commitEntry != null) {
     
         _commitLogCacheMap.remove(commitEntry.atKey);
        _logger.finest('removed key : $commitEntry.atKey from commit log');
      }

VJag: + Log statement can be a bit more descriptive

gkc (commented):
LGTM once Jagan's suggestions incorporated

VJag (approved):

PR #831, opened 2022-07-27
feat: remove malformed keys at server startup

murali-shris:
**- What I did**
- removed malformed keys on server startup
**- How I did it**
- configure list of malformed keys in config.yaml
- added a method _removeMalformedKeys in at_secondary_impl.dart. Read the malformed keys list from config and remove from keystore and commit log
**- How to verify it**
- start a secondary server without the changes. Run the below command in auth mode
update:public:publickey <some_random_value>
- restart the server with the changes
- public:publickey should not appear in scan and sync:from commands

### DIFF
--- a/at_secondary/at_secondary_server/lib/src/server/at_secondary_impl.dart
+++ b/at_secondary/at_secondary_server/lib/src/server/at_secondary_impl.dart
@@ -210,6 +210,9 @@ class AtSecondaryServerImpl implements AtSecondaryServer {
       await initDynamicConfigListeners();
     }
 
+    // clean up malformed keys from keystore
+    await _removeMalformedKeys();
+
     try {
       _isRunning = true;
       if (useSSL!) {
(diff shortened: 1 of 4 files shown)


at_secondary/at_secondary_server/lib/src/server/at_secondary_impl.dart
```diff
@@ -556,6 +559,23 @@ class AtSecondaryServerImpl implements AtSecondaryServer {
...
 
+  Future<void> _removeMalformedKeys() async {
+    try {
+      List<String> malformedKeys = AtSecondaryConfig.malformedKeysList;
+      logger.finest('malformed keys from config: $malformedKeys');
+      for (String key in malformedKeys) {
+        int? commitId = await _secondaryPersistenceStore
+            .getSecondaryKeyStore()!
+            .remove(key);
+        logger.finest('commitId for removed key $key: $commitId');
```
gkc: (1) Only log if the key existed and was therefore deleted (2) log at WARNING if so

gkc (changes requested):

at_secondary/at_secondary_server/lib/src/server/at_secondary_impl.dart
murali-shris: done. please review

gkc (approved):
LGTM - thanks @murali-shris !

PR #832, opened 2022-07-27
feat: add encoding as optional parameter to put and create methods

sitaram-kalluri:
**- What I did**
- Add `encoding` as optional parameter which represents the type of encoding to put and create methods in keystore.dart to support encoding of new line character in AtKey value.

**- How I did it**

**- How to verify it**

- Add encoding as optional parameter to put and create methods


gkc (changes requested):
Rather than just a Boolean isEncoding, instead please can we have encoding:<encoding scheme> (eg encoding:base64) which then enables us later to switch to a different encoding scheme

gkc (approved):

PR #841, opened 2022-08-01
fix: refactor error messages and JSON encode the error responses.

### LINKED ISSUE #509: Error code 'forwarding'
**Describe the bug**
* In the case where a client makes a request to server A, and an exception is thrown in server A because of an error response from server B, we should try to preserve error codes where they make sense in the response to the ultimate client
  * For example: Client calls plookup on Server A; Server A  calls lookup on Server B; server B returns "error:AT0015-Key not found : ..."
  ```
  @sitaram@lookup:phone@murali
  error:null-null : Request to remote secondary @murali at vip.ve.atsign.zone:7001 received error
    response 'error:AT0015-key not found : @sitaram:phone@murali does not exist in keystore'
  ```
  * Server A will currently throw an AtConnectException with the error response it got from B, providing the additional information that it came from B
    * The "null-null" is caused by https://github.com/atsign-foundation/at_tools/issues/152

**Expected behaviour**
  * When returning to the client we would like to preserve that AT0015 (key not found) as it is most correct, but provide the additional context of the full error message from the other end. For example:
  ```
     @sitaram@lookup:phone@murali
     error:AT0015-key not found : Request to remote secondary @murali at vip.ve.atsign.zone:7001 received error
       response 'error:AT0015-key not found : @sitaram:phone@murali does not exist in keystore'
  ```
  * This can be achieved in outbound_message_listener, for example, by having it
    * When there's an error code from remote server, inspect it and match it to its corresponding AtException
      * If there is a matching Exception then throw that, adding the entire message received from remote
      * Else do what's being done now (throw AtConnectException)
  * Consider extending at_commons error_message.dart with a way to create (not throw, just create) the correct exception given an error code. (We currently have mapping from exception to error-code but not the other way around)

sitaram-kalluri:
**- What I did**
- Return error codes and error description on exception.
- Generate the exception type from the error codes on the exception.

**- How I did it**
- When exception is raised, JSON encode the error codes and error descriptions.
- On receiving the exception, decode the JSON, and create the exception type from the error code and set the description.

**- How to verify it**
- Run the scenarios that trigger exception - look for a non existent key and verify the `KeyNotFoundException` type is returned along with error code and error description.

- Forward the error codes and error description.

**-Tests from at_server command line**

Look for a key when root server is not started.
@ sitaram@lookup:phone.wavi@murali
error:{"errorCode":"AT0011","errorDescription":"Internal server exception : Failed connecting to root server url vip.ve.atsign.zone on port 64"}

Look for a key when target secondary is not started.
@ sitaram@lookup:phone.wavi@murali
error:{"errorCode":"AT0007","errorDescription":"Secondary server not found : unable to connect to secondary"}

Look for a non existent key
@ sitaram@lookup:phone.wavi@murali
error:{"errorCode":"AT0015","errorDescription":"key not found : Exception: @ sitaram:phone.wavi@ murali does not exist in keystore"}

### DIFF
--- a/at_end2end_test/test/plookup_verb_test.dart
+++ b/at_end2end_test/test/plookup_verb_test.dart
@@ -1,3 +1,5 @@
+import 'dart:convert';
+
 import 'package:test/test.dart';
 
 import 'e2e_test_utils.dart' as e2e;
@@ -33,7 +35,8 @@ void main() {
     await sh1.writeCommand('update:public:phone$atSign_1 9982212143');
     String response = await sh1.read();
     print('update verb response $response');
-    assert((!response.contains('Invalid syntax')) && (!response.contains('null')));
+    assert(
+        (!response.contains('Invalid syntax')) && (!response.contains('null')));
 
     ///PLOOKUP VERB
     await sh2.writeCommand('plookup:phone$atSign_1');
@@ -47,7 +50,8 @@ void main() {
     await sh1.writeCommand('update:$atSign_2:mobile$atSign_1 9982212143');
     String response = await sh1.read();
     print('update verb response $response');
-    assert((!response.contains('Invalid syntax')) && (!response.contains('null')));
+    assert(
+        (!response.contains('Invalid syntax')) && (!response.contains('null')));
 
     ///PLOOKUP VERB
     await sh2.writeCommand('plookup:mobile$atSign_1$atSign_2');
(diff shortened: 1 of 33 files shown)


at_end2end_test/test/plookup_verb_test.dart
```diff
@@ -64,15 +68,26 @@ void main() {
     await sh1.writeCommand('plookup:no-key$atSign_1');
     String response = await sh1.read();
     print('plookup verb response $response');
-    expect(response, contains('error:AT0015-key not found : public:no-key$atSign_1 does not exist in keystore'));
+    if (atSign_1 != '@cicd5') {
```
gkc, line 69: Please create a github issue to amend these tests once cicd5 is running server version 3.0.21 (i.e. once we have promoted current canary to all of prod)

gkc (changes requested):

at_end2end_test/test/plookup_verb_test.dart
gkc: Alternatively, rewrite this test so the if condition is not about the atSign, but about the version of the server software that the atSign is running

sitaram-kalluri:
Closing this PR and moved the relevant changes to the following PR: https://github.com/atsign-foundation/at_server/pull/850

PR #850, opened 2022-08-09
feat: refactor error messages and JSON encode the error responses.

### LINKED ISSUE #509: Error code 'forwarding'

sitaram-kalluri:
**- What I did**
- Return error codes and error description on exception.
- Generate the exception type from the error codes on the exception.

**- How I did it**
- When exception is raised, JSON encode the error codes and error descriptions.
- On receiving the exception, decode the JSON, and create the exception type from the error code and set the description.

**- How to verify it**
- Run the scenarios that trigger exception - look for a non existent key and verify the `KeyNotFoundException` type is returned along with error code and error description.

- Forward the error codes and error description.

**-Tests from at_server command line**

Look for a key when root server is not started.
@ sitaram@lookup:phone.wavi@murali
error:{"errorCode":"AT0011","errorDescription":"Internal server exception : Failed connecting to root server url vip.ve.atsign.zone on port 64"}

Look for a key when target secondary is not started.
@ sitaram@lookup:phone.wavi@murali
error:{"errorCode":"AT0007","errorDescription":"Secondary server not found : unable to connect to secondary"}

Look for a non existent key
@ sitaram@lookup:phone.wavi@murali
error:{"errorCode":"AT0015","errorDescription":"key not found : Exception: @ sitaram:phone.wavi@ murali does not exist in keystore"}


at_end2end_test/test/e2e_test_utils.dart
```diff
@@ -229,3 +227,15 @@ void _loadTheYaml() {
...
+
+extension Utils on SimpleOutboundSocketHandler {
+  Future<int> getVersion() async {
+    await writeCommand('info\n');
+    var version = await read();
+    version = version.replaceAll('data:', '');
+    // Since secondary version has gha<number> appended, remove the gha number from version
+    // Hence using split.
+    var versionStr = jsonDecode(version)['version'].split('+')[0];
+    return int.parse(versionStr.replaceAll('.', ''));
```
gkc: This will break when we go, let's say, from version 3.0.21 (3021) to version 3.1.1 (311)
Semantically, 3.1.1 is greater than 3.0.21

I'd suggest using https://pub.dev/packages/version

at_end2end_test/test/plookup_verb_test.dart
```diff
@@ -64,7 +66,17 @@ void main() {
     await sh1.writeCommand('plookup:no-key$atSign_1');
     String response = await sh1.read();
     print('plookup verb response $response');
-    expect(response, contains('error:AT0015-key not found : public:no-key$atSign_1 does not exist in keystore'));
+    if (await sh1.getVersion() >= 3021) {
```
gkc: See earlier comment

at_end2end_test/test/update_verb_test.dart
```diff
@@ -67,8 +68,20 @@ void main() {
     ///LLOOKUP VERB - with out @sign does not return value.
     await sh1.writeCommand('llookup:country$atSign_1');
     response = await sh1.read();
-    print('llookup verb response without private key in llookup verb: $response');
-    expect(response, contains('error:AT0015-key not found : country$atSign_1 does not exist in keystore'));
+    print(
+        'llookup verb response without private key in llookup verb: $response');
+    //TODO: remove else block in the next release -  when all servers json encode the error responses
+    if (await sh1.getVersion() >= 3021) {
```
gkc: See earlier comment

at_secondary/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
```diff
@@ -82,17 +82,29 @@ class OutboundMessageListener {
...
-          throw AtConnectException("Request to remote secondary ${outboundClient.toAtSign} at ${outboundClient.toHost}:${outboundClient.toPort} received error response '$result'");
+          try {
+            result = result.toString().replaceFirst('error:', '');
+            var errorMap = jsonDecode(result);
+            throw AtExceptionUtils.get(
+                errorMap['errorCode'], errorMap['errorDescription']);
+          } on FormatException {
+            // Catching the FormatException to preserve backward compatibility - responses without jsonEncoding.
+            // TODO: Can remove the catch block in the next release (once all the existing servers are migrated to new version).
+            throw AtConnectException(
```
gkc, line 93: GIven that the old behaviour was to throw AtConnectException, you need to check for code in the server which was catching AtConnectException, and ensure it works properly with the new exception i.e. whatever `AtExceptionUtils.get` creates

at_secondary/at_secondary_server/lib/src/exception/global_exception_handler.dart
```diff
@@ -124,11 +125,17 @@ class GlobalExceptionHandler {
...
         errorCode ??= 'AT0011';
 
-        var errorDescription;
-        if (exception is AtException) {
-          errorDescription = '${getErrorDescription(errorCode)} : ${exception.message}';
+        String errorDescription;
+        // To avoid duplication of error description in the error message add
+        // errorDescription only on the sender side(not on the receiver side).
+        // Hence checking if the connection is authenticated
+        // The receiver side connections will pol authenticated.
```
gkc: Can you add a little more detail to the comments here as I don't quite understand the scenario and why whether you are authenticated or not makes a difference. Are you sure there are no scenarios where the connection will **not** be authenticated, but we still need to add the description?

at_secondary/at_secondary_server/lib/src/exception/global_exception_handler.dart
```diff
@@ -152,6 +159,10 @@ class GlobalExceptionHandler {
 
   void _writeToSocket(AtConnection atConnection, String prompt,
       String? errorCode, String errorDescription) {
-    atConnection.write('error:$errorCode-$errorDescription\n$prompt');
+    var errorJsonMap = {
```
gkc: Won't an old client fail to parse the JSON?

gkc (changes requested):

at_end2end_test/test/e2e_test_utils.dart
sitaram-kalluri: This is implemented.

at_end2end_test/test/update_verb_test.dart
sitaram-kalluri: This is implemented.

at_secondary/at_secondary_server/lib/src/exception/global_exception_handler.dart
sitaram-kalluri: This is taken care in at_lookup library.

In at_lookup, On receiving the error response, tries to JSON decode the error response. If FormatException is thrown, exception is caught and error response is parsed in the older way.

gkc: Yes it's taken care of ... IF you have the new at_lookup library

My point is that old clients will break. This is a fundamentally breaking change

I believe we need a way for a client to let a server know some things about itself - eg what version of client software am I running, and maybe a set of "features the client supports"

gkc: Let's discuss on architecture call

gkc (approved):

PR #859, opened 2022-08-12
fix: handle invalid atkey exception on server

murali-shris:
** What I did
     - Send InvalidAtKeyException in error response
** How  I did it
    - throw InvalidAtKeyException from UpdateVerbHandler
    - handle the exception in GlobalExceptionHandler
    - added dependency overrides in pubspec for persistence and at_commons.
** How to verify it
    - unit tests, functional and end2end tests should pass


murali-shris:
@gkc Should I comment the end2end test that is failing and put it back after releasing the server ?

gkc:
👍

gkc (approved):

PR #870, opened 2022-08-19
fix: update command validate sharedBy is same as currentAtSign

### LINKED ISSUE #848: update command allows clients to create / update entries whose keys' sharedBy is not the current atSign
**Describe the bug**
update command allows clients to create / update entries whose keys' sharedBy is not the current atSign

**To Reproduce**
issue an update: command directly where sharedBy is different. e.g. if connected to `@alice`'s secondary server as `@alice`, issue an update command where the sharedBy is `@bob`

**Expected behavior**
Should be rejected, as it by Dart at_client package's AtClientImpl and (now) by the Java client

**Additional context**
When I first encountered this, I think I had managed to somehow bypass the client validation using at_daemon but I can't see how that would have been the case since there is validation in the at_client package. I need to follow up to see if I can reproduce that

murali-shris:
**- What I did**
- added a validation check for update command
**- How I did it**
- in update_verb_handler, if sharedBy atsign is set and if it is not same as current atsign then return InvalidAtKeyException
- fixed a code bug in two places where sharedBy and sharedWith were set incorrectly. No impact due to this bug since it was incorrect in couple of places, eventually key was constructed in the correct format.
- added unit tests to "A group of tests to validate sharedBy atsign" in update_verb_test
- fixed few tests which were failing due to new validation
**- How to verify it**
- run the unit tests in secondary server

### DIFF
--- a/at_secondary/at_secondary_server/lib/src/verb/handler/update_verb_handler.dart
+++ b/at_secondary/at_secondary_server/lib/src/verb/handler/update_verb_handler.dart
@@ -64,11 +64,19 @@ class UpdateVerbHandler extends ChangeVerbHandler {
     // Sets Response bean to the response bean in ChangeVerbHandler
     await super.processVerb(response, verbParams, atConnection);
     var updateParams = _getUpdateParams(verbParams);
-
+    if (updateParams.sharedBy != null &&
+        updateParams.sharedBy!.isNotEmpty &&
+        AtUtils.formatAtSign(updateParams.sharedBy) !=
+            AtSecondaryServerImpl.getInstance().currentAtSign) {
+      logger.warning(
+          'Invalid update command sharedBy atsign ${AtUtils.formatAtSign(updateParams.sharedBy)} should be same as current atsign ${AtSecondaryServerImpl.getInstance().currentAtSign} ');
+      throw InvalidAtKeyException(
+          'SharedBy atsign should be the same as current atsign');
+    }
     try {
       // Get the key and update the value
-      var forAtSign = updateParams.sharedBy;
-      var atSign = updateParams.sharedWith;
+      var forAtSign = updateParams.sharedWith;
+      var atSign = updateParams.sharedBy;
       var key = updateParams.atKey;
       var value = updateParams.value;
       var atData = AtData();
@@ -187,14 +195,13 @@ class UpdateVerbHandler extends ChangeVerbHandler {
 
   UpdateParams _getUpdateParams(HashMap<String, String?> verbParams) {
     if (verbParams['json'] != null) {
-      print('update json');
       var jsonString = verbParams['json']!;
       Map jsonMap = jsonDecode(jsonString);
       return UpdateParams.fromJson(jsonMap);
     }
     var updateParams = UpdateParams();
-    updateParams.sharedBy = verbParams[FOR_AT_SIGN];
-    updateParams.sharedWith = verbParams[AT_SIGN];
+    updateParams.sharedBy = verbParams[AT_SIGN];
+    updateParams.sharedWith = verbParams[FOR_AT_SIGN];
     updateParams.atKey = verbParams[AT_KEY];
     updateParams.value = verbParams[AT_VALUE];
     var metadata = Metadata();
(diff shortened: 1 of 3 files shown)


at_secondary/at_secondary_server/lib/src/verb/handler/update_verb_handler.dart
```diff
@@ -187,14 +195,13 @@ class UpdateVerbHandler extends ChangeVerbHandler {
...
-      print('update json');
       var jsonString = verbParams['json']!;
       Map jsonMap = jsonDecode(jsonString);
       return UpdateParams.fromJson(jsonMap);
     }
     var updateParams = UpdateParams();
-    updateParams.sharedBy = verbParams[FOR_AT_SIGN];
-    updateParams.sharedWith = verbParams[AT_SIGN];
+    updateParams.sharedBy = verbParams[AT_SIGN];
+    updateParams.sharedWith = verbParams[FOR_AT_SIGN];
```
gkc, line 204: Thank you for cleaning this up

at_secondary/at_secondary_server/lib/src/verb/handler/update_verb_handler.dart
```diff
@@ -64,11 +64,19 @@ class UpdateVerbHandler extends ChangeVerbHandler {
...
+          'Invalid update command sharedBy atsign ${AtUtils.formatAtSign(updateParams.sharedBy)} should be same as current atsign ${AtSecondaryServerImpl.getInstance().currentAtSign} ');
+      throw InvalidAtKeyException(
+          'SharedBy atsign should be the same as current atsign');
+    }
     try {
       // Get the key and update the value
-      var forAtSign = updateParams.sharedBy;
-      var atSign = updateParams.sharedWith;
+      var forAtSign = updateParams.sharedWith;
+      var atSign = updateParams.sharedBy;
```
gkc, line 79: Thank you for cleaning this up

at_secondary/at_secondary_server/lib/src/verb/handler/update_verb_handler.dart
```diff
@@ -64,11 +64,19 @@ class UpdateVerbHandler extends ChangeVerbHandler {
...
     var updateParams = _getUpdateParams(verbParams);
-
+    if (updateParams.sharedBy != null &&
+        updateParams.sharedBy!.isNotEmpty &&
+        AtUtils.formatAtSign(updateParams.sharedBy) !=
+            AtSecondaryServerImpl.getInstance().currentAtSign) {
+      logger.warning(
+          'Invalid update command sharedBy atsign ${AtUtils.formatAtSign(updateParams.sharedBy)} should be same as current atsign ${AtSecondaryServerImpl.getInstance().currentAtSign} ');
+      throw InvalidAtKeyException(
+          'SharedBy atsign should be the same as current atsign');
```
gkc: Maybe make the exception message the same as what is logged above - i.e. include the actual sharedBy and current atSigns

at_secondary/at_secondary_server/test/update_verb_test.dart
```diff
@@ -678,27 +681,69 @@ void main() {
...
+              .getSecondaryPersistenceStore(
+                  AtSecondaryServerImpl.getInstance().currentAtSign)!;
+      handler.keyStore = secondaryPersistenceStore
+          .getSecondaryKeyStoreManager()!
+          .getKeyStore();
+      var response = Response();
+      var verbParams = handler.parse(command);
+      var atConnection = InboundConnectionImpl(null, null);
+      expect(
+          () => handler.processVerb(response, verbParams, atConnection),
```
gkc: processVerb is async ... does this `expect` definitely execute fully or should it be `await handler.processVerb` ?

gkc (changes requested):
One question, one minor change suggested

at_secondary/at_secondary_server/test/update_verb_test.dart
murali-shris: test was failing when I added async/await. Fixed the test now. expectLater is right method to use

at_secondary/at_secondary_server/lib/src/verb/handler/update_verb_handler.dart
murali-shris: done

gkc (approved):

PR #874, opened 2022-08-23
feat: add clientConfig to from verb

sitaram-kalluri:
**- What I did**
1. When client version is passed, add it to the InboundConnection's AtConnectionMetaData.
2. If client version is not passed, default it to `n/a` (not available)

**- How I did it**
1. Introduce a new instance variable to AtConnectionMetaData - 'clientVersion' in `at_connection.dart`  and default it to 'n/a'
2. If client version is passed, set the client version in AtConnectionMetaData

**- How to verify it**
1. If client version is passed, clientVersion in at_connection metadata should be populated. other latest will set by default.

feat: add clientConfig to from verb


at_server_spec/lib/src/connection/at_connection.dart
```diff
@@ -31,4 +31,7 @@ abstract class AtConnectionMetaData {
   bool isPolAuthenticated = false;
   bool isStream = false;
   String? streamId;
+  /// Represents the version of the client initiated the connection.
+  /// Defaults to latest which indicates the latest release.
+  String clientVersion = 'latest';
```
gkc: Can't default to 'latest'. By definition the older a client is, the less likely it is to be sending its version. Suggest default to 'n/a' (not available)

gkc (changes requested):

at_server_spec/lib/src/connection/at_connection.dart
gkc: I've made this change and pushed to the branch

gkc (approved):

PR #892, opened 2022-09-02
feat: setting socketOption tcpNoDelay -> true

srieteja:
**- What I did**
setting socketOption tcpNoDelay -> true while creating sockets

**- How I did it**
setting the socket option in base_connection class constructor which is underlying connection for inbound and outbound connections


at_secondary/at_secondary_server/lib/src/connection/base_connection.dart
```diff
@@ -47,7 +49,8 @@ abstract class BaseConnection extends AtConnection {
       throw ConnectionInvalidException('Connection is invalid');
     }
     try {
-      logger.info('SENT: [${getMetaData().sessionID}] ${BaseConnection.truncateForLogging(data)}');
+      logger.info(
```
VJag: Pls not include the other style changes

VJag (approved):

PR #902, opened 2022-09-07
fix: randomize the cron jobs and add reset method to AtNotification

sitaram-kalluri:
**- What I did**
- Randomize the compaction job's start time
- Reduce the default notification expiry to 15 minutes from 24 hours
- Add reset method to AtNotificationBuilder which reset's the state of AtNotificationBuilder
- Update the string concatenation with String interpolation as per dart analyze suggestion.
- Update the version's in pubspec.yaml and changelog.md

**- How I did it**
- Added `Future.delayed` by random generated number (max 12) of seconds before the compaction job start's
- Change the default constant value in AtNotificationBuilder.

**- How to verify it**

### DIFF
--- a/at_secondary/at_persistence_secondary_server/lib/src/notification/at_notification.dart
+++ b/at_secondary/at_persistence_secondary_server/lib/src/notification/at_notification.dart
@@ -416,7 +416,8 @@ class MessageTypeAdapter extends TypeAdapter<MessageType?> {
 
 /// AtNotificationBuilder class to build [AtNotification] object
 class AtNotificationBuilder {
-  static const int _defaultTTLInHrs = 24;
+  static const int _defaultTTLInMins = 15;
+
   String? id = Uuid().v4();
 
   String? fromAtSign;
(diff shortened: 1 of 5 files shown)


at_secondary/at_persistence_secondary_server/lib/src/notification/at_notification.dart
```diff
@@ -459,4 +460,26 @@ class AtNotificationBuilder {
     }
     return AtNotification._builder(this);
   }
+
+  reset() {
```
VJag, line 464: Should we introduce reset along with the call? I am fine if there is a reason for this,

VJag (approved):

at_secondary/at_persistence_secondary_server/lib/src/notification/at_notification.dart
sitaram-kalluri: The caller of the reset method in the at_secondary_server. Wanted to publish at_persistence_secondary_server and then use the reset method in the at_secondary_server

PR #913, opened 2022-09-14
fix: Change duration of ttln in notify_verb_handler.dart

sitaram-kalluri:
**- What I did**
- Change the notification expiry time from milliseconds to minutes.

**- How I did it**

**- How to verify it**


sitaram-kalluri:
The changes are not required. Instead of changing the milliseconds to minutes, on client side expose Duration and convert to milliseconds and send to server. Hence closing the PR.

PR #958, opened 2022-10-07
fix: skip statsNotificationId to sync from client to secondary

sitaram-kalluri:
**- What I did**
- Skip 'statsNotificationId' to sync from client to cloud secondary

**- How I did it**
- Added 'skipNotificationId' to regex in 'commit' method to skip the from being added to commit log.
- The keys are being converted to lower case, hence in-order to satisfy the IF the condition set the `caseSentivity: false` in the Regex

**- How to verify it**
- The skipNotificationId should not be synced to the cloud secondary

- Skip commit id for the 'statsNotificationId'

### DIFF
--- a/at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/at_commit_log.dart
+++ b/at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/at_commit_log.dart
@@ -1,3 +1,4 @@
+import 'package:at_commons/at_commons.dart';
 import 'package:at_persistence_secondary_server/at_persistence_secondary_server.dart';
 import 'package:at_persistence_secondary_server/src/event_listener/at_change_event.dart';
 import 'package:at_persistence_secondary_server/src/event_listener/at_change_event_listener.dart';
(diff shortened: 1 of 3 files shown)


VJag (dismissed):

at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/at_commit_log.dart
```diff
@@ -31,7 +32,7 @@ class AtCommitLog implements AtLogType {
     // The private: and privatekey: are not synced. so return -1.
     if (!key.startsWith('public:__') &&
         (key.startsWith(RegExp(
-                'private:|privatekey:|public:_|public:signing_publickey')) ||
+                'private:|privatekey:|public:_|public:signing_publickey|${statsNotificationId.toLowerCase()}')) ||
```
gkc: Maybe make the whole regex case-insensitive rather than depending on the case of the things we're checking against. https://api.dart.dev/stable/2.18.2/dart-core/RegExp/isCaseSensitive.html

gkc (changes requested):

at_secondary/at_persistence_secondary_server/lib/src/log/commitlog/at_commit_log.dart
sitaram-kalluri: set the caseSensitive to false in Regex.

gkc (approved):
LGTM thanks @kalluriramkumar

PR #959, opened 2022-10-07
fix: dess inc armv7 releases when canary promoted

### LINKED ISSUE #945: Dess secondary images haven't been updated since 3.0.19 due to using promotion from canary.
**Describe the bug**

The present atsigncompany/secondary:dess image carries a 3.0.19 secondary, as that's the last version that we built a production release for. Since then we've been promoting canary images, which doesn't update the dess tags or build an Armv7 image.

**To Reproduce**

1. First I login to a VM running dess
2. Then I `sudo docker service update --image atsigncompany/secondary:dess $DESS_SECONDARY`
3. And then `openssl s_client 0:$DESS_PORT` and run `info:brief`

**Expected behavior**

Dess builds should keep up with prod

**Additional context**

The problem piece here is Armv7, which we support for dess due to earlier Raspberry Pis etc. For a fully functional dess image we need to build Armv7, and there's a risk that the Armv7 image could get out of sync with other aspects of a release if it's built at another time (from potentially different dependencies).

**Workaround**

Armv8  and x86_64 dess hosts can simply use the prod image  `sudo docker service update --image atsigncompany/secondary:prod $DESS_SECONDARY`

cpswan:
Fixes #945 

**- What I did**

Arm v7 and multi arch steps cloned from existing dess builds into canary so that any canary build will support all the architectures when they're promoted.

Workflow will also run for a release tagged with `p*.*.*` for **promote** and this should be our normal way of doing promotions going forward so that they're visible as releases.

**- How I did it**

Liberal reuse of existing steps with s/dess/canary/

**- How to verify it**

We will need to wait for a canary that's ready for promotion.

fix: dess inc armv7 releases when canary promoted

### DIFF
--- a/.github/workflows/at_server.yaml
+++ b/.github/workflows/at_server.yaml
@@ -557,6 +557,47 @@ jobs:
           platforms: |
             linux/amd64
 
+      # Retag prod x64 and Arm64 images as canary
+      - name: Create canary-x64 label
+        run: |
+          docker pull atsigncompany/secondary:canary
+          docker tag atsigncompany/secondary:canary atsigncompany/secondary:canary-x64
+          docker push atsigncompany/secondary:canary-x64
+
+      - name: Create dess-arm64 label
+        uses: appleboy/ssh-action@v0.1.5
+        with:
+          host: "arm64cicd.atsign.wtf"
+          username: ubuntu
+          key: ${{ secrets.CICD_SSH_KEY }}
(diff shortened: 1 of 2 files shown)


gkc (approved):
LGTM

cpswan:
I should add that we need to do another canary release before we do a promotion (or things will fail due to the armv7 image not being there).

PR #966, opened 2022-10-11
fix: sync bug on expired keys

murali-shris:
**- What I did**
- fixed a bug in sync where expired keys are not removed from cache resulting in multiple delete operations on commit log
**- How I did it**
- put method in hive_keystore.dart was adding key without encoding to metadata cache. Fixed the code to add encoded keys to cache
- modified logic to delete from cache - handled both encoded and actual keys
- added unit tests to group "Verify metadata cache" in hive_keystore_impl_test
**- How to verify it**
- run the group of tests "Verify metadata cache" in hive_keystore_impl_test
Fixes a bug encountered during testing of https://github.com/atsign-foundation/at_client_sdk/issues/742


at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
```diff
@@ -235,9 +236,17 @@ class HiveKeystore implements SecondaryKeyStore<String, AtData?, AtMetaData?> {
...
-      result = await _commitLog.commit(key, CommitOp.DELETE);
+      final atMetaData = _removeKeyFromMetadataCache(key);
+      if (atMetaData != null || isKeyPresent) {
+        // add entry to commit log only if key is removed from cache or key is present in keystore
+        result = await _commitLog.commit(key, CommitOp.DELETE);
+      } else {
+        logger.finer(
+            'entry for $key is not present in cache. Skipping commit log delete op');
+      }
       return result;
```
gkc: what's the impact of returning null (when we skip the commit log delete op) ?

at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
```diff
@@ -249,6 +258,21 @@ class HiveKeystore implements SecondaryKeyStore<String, AtData?, AtMetaData?> {
     }
   }
 
+  AtMetaData? _removeKeyFromMetadataCache(String key) {
+    // try removing the passed key as it is. key may be encoded or not encoded.
```
gkc: Cache is created upon process start, so shouldn't we know definitively how to remove the key from the cache?

gkc (changes requested):

at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
murali-shris: remove is called in multiple places. out of which result is consumed in two places
1) local_secondary.delete - data:null will be returned if key is not present in cache/keystore
2) in at_secondary_impl --> _removeMalformedKeys()..we are passing the commitId to  await _commitLog.commitLogKeyStore.remove(commitId); This needs a null check before remove is called.

at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
murali-shris: you are right Gary.. After the fix when we restart the server, cache should have keys in correct format.I will re-test this scenario and fix the code.

VJag (dismissed):
LGTM. Let's document the test scenario's in the ticket that proves that the changes do not cause any regressions. Once we document if there are any gaps we can add unit tests for those scenarios.

murali-shris:
@gkc I have made the below changes
* store actual keys in cache(previous impl stored encoded keys in cache) and encoded keys in hive store(existing behavior)
* refactored code to use keystoreHelper.prepareKey wherever we store/retrieve/delete key from hive keystore 
* remove from metadata cache will remove the passed key. Doesn't require encoding.

at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
gkc: Follow-up questions:
1. What is the impact of local_secondary.delete returning "data:null" ?
2. Has that null check been implemented?

gkc (changes requested):

at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
murali-shris: An existing functional test fail due to this change
https://github.com/atsign-foundation/at_server/actions/runs/3265641359/jobs/5368172492
When we delete non existent key should we return data:null or KeyNotFoundException?  will bring this up in architecture call.

gkc: Usual follow-up question: What is the impact of returning data:null vs throwing KeyNotFoundException?

gkc: I think it would be good to describe the behaviours through the various call chains before this PR, and what they are now. We're trying to implement the scenario where where a key *does* exist, but HiveKeyStore is going to return null when deleting the object, because we are not generating an entry in the commit log.

We've got multiple intertwined things happening here (1) what we do for a keyStore delete if the key does not exist (2) whether or not we generate a commitID for a given key deletion,  (3) what gets returned at the atProtcocol level when this is a server operation while handling a `delete:` verb request (4) what gets returned by LocalSecondary when this is a client-side delete

gkc:
I have a question about sync and delete requests. Consider this scenario:
* Client deletes a key locally
* Sync runs, sends the delete to the server
* Server executes the delete
* But the network connection drops and the response doesn't get to the client
* **I believe** current sync session will stop and sync will start anew
* Client-side **I believe** will then receive the commitID for the delete from the server

1. If my understanding above is correct, will this be handled OK?
2. If my understanding above is incorrect, what would the actual sequence of events be?

at_secondary/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
murali-shris: Screenshot from 2022-10-26 11-28-19

murali-shris:
> I have a question about sync and delete requests. Consider this scenario:
> 
> * Client deletes a key locally
> * Sync runs, sends the delete to the server
> * Server executes the delete
> * But the network connection drops and the response doesn't get to the client
> * **I believe** current sync session will stop and sync will start anew
> * Client-side **I believe** will then receive the commitID for the delete from the server
> 
> 1. If my understanding above is correct, will this be handled OK?
> 2. If my understanding above is incorrect, what would the actual sequence of events be?

@gkc I  have explained the  existing code behaviour for the scenario you have requested.
https://docs.google.com/spreadsheets/d/14rgJy4WF1wms0zQL4p95pk72W87VCti5pIKoJgfeA30/edit?pli=1#gid=865734980
After the changes in this PR, when non existent key is deleted, error response will be returned by the server for deleted key. In sync service we update commitId=-1 in local commit log, when there is no data:<commit_id> response from server.

gkc:
Thanks @murali-shris

gkc:
I think this is good to go. As this is a critical part of the codebase, it might be no harm to do one last walk through the code changes with a colleague before merging.

gkc (dismissed):
@murali-shris 
I think this is good to go. As this is a critical part of the codebase, it would probably be a good idea to do one last walk through the code changes with a colleague before you merge

murali-shris:
> @murali-shris I think this is good to go. As this is a critical part of the codebase, it would probably be a good idea to do one last walk through the code changes with a colleague before you merge

sure Gary

murali-shris:
> > @murali-shris I think this is good to go. As this is a critical part of the codebase, it would probably be a good idea to do one last walk through the code changes with a colleague before you merge
> 
> sure Gary

* Went through the code with @VJag . Made minor changes to documentation and added two more unit tests

VJag (approved):

PR #979, opened 2022-10-16
feat: skip local keys in commit log

sitaram-kalluri:
**- What I did**
- Skip local keys from adding to commit log.
- Updated the unit tests accordingly
- TODO : Added dependency overrides for at_commons for local key changes. Currently changes are available in trunk. Need to publish at_commons and update the at_commons version in pubspec.yaml

**- How I did it**
- Updated the IF condition to skip keys starting 'local:' to skip commit log.

**- How to verify it**
- Create local keys and keys will not be sync'ed between local and cloud secondary

- Refrain adding local keys to commit log

### DIFF
--- a/at_secondary/at_persistence_secondary_server/CHANGELOG.md
+++ b/at_secondary/at_persistence_secondary_server/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 3.0.40
+- feat: Refrain adding local keys to commit log.
 ## 3.0.39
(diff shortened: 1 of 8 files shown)


gkc (dismissed):
LGTM. One request which is to add an explicit warning to code re Local keys that they should only ever exist on the client side. (If for example the server in future got a code change which created Local keys and persisted them in the server, it would wreak havoc on sync.)

murali-shris (approved):

PR #1007, opened 2022-10-27
fix: notificationDateTime has correct date

### LINKED ISSUE #1003: Memory optimisations for notifications have resulted in epochMillis not being set correctly for notifications
### Describe the bug

1. We are now re-using the same AtNotificationBuilder object in the NotifyVerbHandler every time. However, we are not explicitly setting the notificationDateTime in the verb handler. As a result, the notificationDateTime that is used is not the current time, but is instead either the time the AtNotificationBuilder was first created (when the verb handler was created at server startup) or when it was 'reset' during the last previous execution of the verb handler.
2. The code as it stands is vulnerable to race conditions. Right now If we are going to reuse a single AtNotificationBuilder then we need to wrap the entire body of NotifyVerbHandler.processVerb in a `try {get mutex; ...; ...; } finally {release mutex}`
3. As it's possible that other bugs were introduced in these code changes, the changes should be re-reviewed in detail.

### Steps to reproduce

Send some notifications.

### Expected behavior

epochMillis should be correct

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

Bug observed by @ralfeus

sitaram-kalluri:
Fixes #1003 

**- What I did**
1. Add mutex to the the NotifyVerbHandler.processVerb to prevent race conditions as we now have long-lived state (the NotificationBuilder which we are now reusing rather than creating a new one for every notification)
2. Move the reset method to the start of the NotifyVerbHandler.processVerb. So that it gets called before the start of every notification.
3. Added unit test, functional test and end-to-end test to assert that the second notification has a correct date and time.

**- How I did it**
1. Moved the reset method to the start of the NotifyVerbHandler.processVerb in NotifyVerbHandler to fix the issue of subsequent notifications having incorrect date-time.

**- How to verify it**
1. The subsequent notification should have the correct date and time.
2. The unit tests, functional tests and end-to-end tests assert the functionality,

1. fix: Subsequent notifications have the incorrect date-time.

### DIFF
--- a/packages/at_secondary_server/lib/src/verb/handler/notify_verb_handler.dart
+++ b/packages/at_secondary_server/lib/src/verb/handler/notify_verb_handler.dart
@@ -14,6 +14,7 @@ import 'package:at_secondary/src/verb/verb_enum.dart';
 import 'package:at_server_spec/at_server_spec.dart';
 import 'package:at_server_spec/at_verb_spec.dart';
 import 'package:at_utils/at_utils.dart';
+import 'package:mutex/mutex.dart';
 
 enum Type { sent, received }
 
@@ -23,6 +24,8 @@ class NotifyVerbHandler extends AbstractVerbHandler {
   NotifyVerbHandler(SecondaryKeyStore? keyStore) : super(keyStore);
 
   AtNotificationBuilder atNotificationBuilder = AtNotificationBuilder();
+
+  Mutex processNotificationMutex = Mutex();
 
   /// A hashmap which holds the AtMetadata objects.
   /// The key represents if the notification text is encrypted or not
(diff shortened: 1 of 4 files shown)


gkc:
> TODO: Explicitly add date time to notification in the NotifyVerbBuilder. @gkc: please confirm if we need this change or if changes made would suffice

Now that you're calling reset() at the start, it should be fine


> Note: The end2end tests are failing the changes are not available in the test atsign's. Will look for an approach; perhaps add a version to the tests.

Yes. I think @murali-shris is adding a server version check to tests in another branch at the moment also.

sitaram-kalluri:
> > TODO: Explicitly add date time to notification in the NotifyVerbBuilder. @gkc: please confirm if we need this change or if changes made would suffice
> > Now that you're calling reset() at the start, it should be fine
> 
> > Note: The end2end tests are failing the changes are not available in the test atsign's. Will look for an approach; perhaps add a version to the tests.
> > Yes. I think @murali-shris is adding a server version check to tests in another branch at the moment also.

@gkc: Skipped the e2e test for now. Will remove the skip tag and add a version check once we have the changes merged to the trunk branch.

The reason we skipped is, at the moment the trunk branch is already on v3.0.25 and the prod server is running on v3.0.24.

If we have a check to run tests on a version greater than or equal to 3.0.25, then the test fails on atsign running on the trunk - since the changes are not merged yet. Hence, we thought to merge to the trunk branch, and then have a version check to run tests between the trunk and prod servers.

gkc:
@kalluriramkumar ACK

gkc (approved):

PR #1017, opened 2022-10-29
feat: Modified cert renewal script for github workflow 

kumarnarendra701:
**- What I did**
- Modified cert renewal script for GitHub workflow 
**- How I did it**
- Created a separate script to generate a cert for workflow.
**- How to verify it**
- By using the last run in GitHub action and certs renewal date.


.github/workflows/refreshcerts.yaml
```diff
@@ -2,74 +2,61 @@ name: Refreshcerts
...
+          cd .. && rm -rf  vip.ve.atsign.zone* secondaries-scripts
           git config --global user.name 'Getcert Action'
           git config --global user.email '41898282+github-actions[bot]@users.noreply.github.com'
+          set -e
         env:
-          GITHUB_API_TOKEN: ${{ secrets.MY_GITHUB_TOKEN }}
           DO_KEY: ${{ secrets.DO_KEY }}
+          gChat_url: ${{ secrets.GOOGLE_CHAT_WEBHOOK }}
 
-      - name: store logs and responses
```
cpswan, line 66: Why has this step been removed?

cpswan (changes requested):
This is an improvement on what was here before, though I'm not clear why the step to archive logs was removed (it was there so we could see what went wrong if there was a problem).

.github/workflows/refreshcerts.yaml
kumarnarendra701: @cpswan - The reason to remove these steps is to sync all cert renewal workflow with the latest script(create_cert_workflow.sh) and we can see error logs during cert generation in the action console.

cpswan (approved):

PR #1018, opened 2022-10-30
fix: add `await` when calling acquire() on the processNotificationMutex

gkc:
**- What I did**
fix: add `await` when calling acquire() on the processNotificationMutex

**- How to verify it**
tests pass

fix: add `await` when calling acquire() on the processNotificationMutex


gkc:
@kalluriramkumar Realized that you weren't using `await` when calling `acquire()` on the mutex. Have to use await when calling acquire()

nickelskevin (approved):

PR #1065, opened 2022-11-17
fix: add SHAs to Python dependency files

cpswan:
**- What I did**

* Configured poetry to use the dependencies identified by Dependabot
* Generated a requirements.txt from poetry

**- How I did it**

 ```
curl -sSL https://install.python-poetry.org | python3 -
poetry init
poetry export --format requirements.txt --output requirements.txt
```

**- How to verify it**

Remaining Pinned-Dependencies issues should be removed

fix: add SHAs to Python dependency files

### DIFF
--- a/tools/poetry.lock
+++ b/tools/poetry.lock
@@ -1,0 +1,166 @@
+[[package]]
+name = "certifi"
+version = "2022.9.24"
+description = "Python package for providing Mozilla's CA Bundle."
+category = "main"
+optional = false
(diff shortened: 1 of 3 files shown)


gkc (approved):

gkc:
@cpswan I've approved but was wondering if the poetry.lock is required to be in the repo?

cpswan:
@gkc yes, this is a case where the lock file should be committed

PR #1087, opened 2022-12-05
fix: remove invalid keys on server start-up

sitaram-kalluri:
**- What I did**
1. Remove malformed keys from the Keystore
   - Keys that start with "public:cached:" or IF equals "public:publickey".

**- How I did it**
1. Modify the _removeMalformedKeys method to remove the keys that starts with "public:cached:"
2. Add a config parameter "shouldRemoveMalformedKeys" which is set to false by default.
3. Setting it to true, will look for invalid keys and delete them on server start-up.

1. Remove malformed keys on server start-up


gkc (changes requested):
Two things

(1) I made some changes and pushed a commit to the branch: `refactor: Changed some newly-added variable names, env variable names and yaml config names for consistency. (If we're using 'malformed' in existing code then it should be called 'malformed' in new code as well, not 'invalid')`

Please review and make sure you're happy with my change

(2) Please add unit tests (at least one) to verify that _removeMalformedKeys works as expected

sitaram-kalluri:
> Two things
> 
> (1) I made some changes and pushed a commit to the branch: `refactor: Changed some newly-added variable names, env variable names and yaml config names for consistency. (If we're using 'malformed' in existing code then it should be called 'malformed' in new code as well, not 'invalid')`
> 
> Please review and make sure you're happy with my change
> 
> (2) Please add unit tests (at least one) to verify that _removeMalformedKeys works as expected

Sure Gary, will add unit test and push the changes

murali-shris (dismissed):

packages/at_secondary_server/config/config.yaml
```diff
@@ -38,6 +38,10 @@ hive:
   expiringRunFrequencyMins: 12
   # List of malformed keys that will be deleted on server startup
   malformedKeys: 'public:publickey'
+  # Delete keys that starts with "public:cached" on server startup
+  # This is set to true by default. To retain the invalid keys on server start-up,
+  # set the flag to false.
+  shouldRemoveInvalidKeys: true
```
gkc: Incorrect name, does not match code

gkc (changes requested):

packages/at_secondary_server/config/config.yaml
sitaram-kalluri: Will update the name and push the changes.

gkc (approved):

PR #1107, opened 2022-12-16
chore: Code gardening / maintenance. Sped up some unit tests, cleaned up some lint, upgraded dependencies

gkc:
**- What I did**
* Made unit tests run faster
* Cleaned up some lint warnings
* Upgraded some dependencies

**- How I did it**
* test: Tweaked multiple tests which depend on time, to shorten the durations where they were unnecessarily long, so that the tests can run faster
* refactor: Allow optional passing of the timer duration to StatsNotificationService.schedule() for test purposes
* fix: StatsNotificationService.schedule() will throw a StateError if it is currently being scheduled or has already been scheduled. (This was not enforced previously but was stated as an expectation on the class)
* feat: Added StatsNotificationService.cancel() which will cancel the existing timer (if any) and allow schedule() to be called again
* style: Clarified comments in StatsNotificationService
* style: Cleaned up a bunch of lint warnings in at_secondary_config.dart
* build: Removed mockito dependency. Made existing dependency on mocktail explicit by adding to pubspec.yaml
* build: Upgraded dependencies to their latest versions

**- How to verify it**
Tests should pass

- Code gardening / maintenance. Sped up some unit tests, cleaned up some lint, upgraded dependencies


packages/at_secondary_server/lib/src/notification/stats_notification_service.dart
```diff
@@ -43,30 +45,53 @@ class StatsNotificationService {
...
+  static final Duration zeroDuration = Duration(microseconds: 0);
+
   // Counter for number of active monitor connections. Used for logging purpose.
   int numOfMonitorConn = 0;
 
   Notification notification = Notification.empty();
 
+  @visibleForTesting
+  /// Set to true while the job is being scheduled, false once it has been scheduled.
+  bool scheduling = false;
```
cpswan: Could this be made less confusing?

As a native English speaker I can tell the difference between scheduling and scheduled, but once the code below starts using them for conditionals it can get murky what's going on.

I think what we have here is a state machine, and so it might be preferable to have a single variable that holds the state of the machine (and then maybe a switch statement to act on the state).

packages/at_secondary_server/lib/src/server/at_secondary_config.dart
```diff
@@ -5,11 +5,10 @@ import 'package:at_commons/at_commons.dart';
 import 'package:at_secondary/src/conf/config_util.dart';
 
 class AtSecondaryConfig {
-  static Map<ModifiableConfigs, ModifiableConfigurationEntry> _streamListeners =
-      {};
+  static final Map<ModifiableConfigs, ModifiableConfigurationEntry> _streamListeners = {};
   //Certs
-  static final bool? _useSSL = true;
-  static final bool? _clientCertificateRequired = true;
+  static final bool _useSSL = true;
```
cpswan: Nit - SSL is long deprecated vs TLS (though I realise we often still talk about 'SSL')

cpswan (dismissed):
Nothing here that should stop this being merged, but a couple of comments for your consideration.

packages/at_secondary_server/lib/src/notification/stats_notification_service.dart
gkc: Good suggestion - thanks - done

cpswan: I'm not seeing a new commit pushed yet

packages/at_secondary_server/lib/src/server/at_secondary_config.dart
gkc: Done

packages/at_secondary_server/lib/src/notification/stats_notification_service.dart
gkc: You're too keen :-) I didn't say I'd pushed it. I will re-request your review once the branch has been pushed and the checks have passed

cpswan (approved):

PR #1158, opened 2023-01-18
fix: update sync and functional tests

### LINKED ISSUE #1126: Version in the metadata not being updated
### Describe the bug

The "version" field in the metadata does not get updated after updating the key

### Steps to reproduce

1. First create a new key
2. Then  update the same key again
3. And then get() the value

### Expected behavior

The version field should be updated from 0 to 1 after the key is updated.

Also, the updatedAt, createdAt fields shows incorrect value

### Screenshots

_No response_

_No response_

### Additional context

_No response_

sitaram-kalluri:
**- What I did**
* Add new tests to assert the changes related to fixing the AtMetadata.version does not update upon updating of the key.
* Implemented the TODO's in sync_unit_test that asserts the AtMetadata.
* Renamed "multiple_keys_update_test.dart" as "create_update_key_test.dart" 
**- How I did it**
* In at_functional_tests, rename "multiple_keys_update_test.dart" file to "create_update_key_test.dart" and the following new tests
  * A test to assert default fields in metadata is populated on the creation of a key
  * A test to assert default fields in metadata is populated on the update of a key

**- How to verify it**
* The unit and functional tests should pass

* Update functional tests


gkc:
@sitaram-kalluri Is this ready for review? The tests still have some comments listing assertions that aren't yet being made (about createdAt and updatedAt and etc)

sitaram-kalluri:
> @sitaram-kalluri Is this ready for review? The tests still have some comments listing assertions that aren't yet being made (about createdAt and updatedAt and etc)

@gkc: Thank you for pointing out what I missed. I have updated the test which has ToDo for "createdAt" and "updatedAt" and pushed the changes.

packages/at_secondary_server/test/sync_unit_test.dart
```diff
@@ -112,6 +112,7 @@ void main() {
               commitEntryListBeforeUpdate.first!.atKey, '@alice:phone@alice');
           expect(commitEntryListBeforeUpdate.first!.commitId, 0);
           // Update the same key again
+          var keyUpdationDateTime = DateTime.now().toUtc();
```
gkc: Please can you replace all usages of 'updation' with 'update' in code and comments; 'updation' is not standard english and is not used outside of india / pakistan / sri lanka / bangladesh

gkc (changes requested):

packages/at_secondary_server/test/sync_unit_test.dart
sitaram-kalluri: Changed  "updation" to "update"

packages/at_secondary_server/pubspec.yaml
```diff
@@ -26,17 +26,17 @@ dependencies:
...
-#  at_persistence_secondary_server:
-#    git:
-#      url: https://github.com/atsign-foundation/at_server.git
-#      path: packages/at_persistence_secondary_server
-#      ref: rollback_delete_keynotfound
+  at_persistence_secondary_server:
+    git:
+      url: https://github.com/atsign-foundation/at_server.git
+      path: packages/at_persistence_secondary_server
+      ref: fix_version_not_updated
```
gkc: Will need to remove the dependency override before we merge to trunk

gkc (changes requested):

packages/at_secondary_server/pubspec.yaml
sitaram-kalluri: Remove dependency_overrides and update with hosted version.

gkc (approved):

PR #1159, opened 2023-01-18
fix: version field in atMetadata not updating

### LINKED ISSUE #1126: Version in the metadata not being updated

sitaram-kalluri:
**- What I did**
* Fix the issue of updating the key does not update the following fields in the metadata
   * version
   * createdBy
   * updatedBy

**- How I did it**
* Move the logic of incrementing the version to AtMetadataBuilder class.
* Pass atSign as an argument to AtMetadataBuilder class constructor to populate createdBy and updatedBy fields.
* Add a condition to prevent updating of "createdAt" field on update of a key. "createdAt" should be populated on the creation of a key and should remain unchanged. On updating the same key, "updatedAt" will updated.

**- How to verify it**
* Added the following unit tests to assert the changes:
  * A test to default field in metadata is set on a new key creation
  * A test to verify version field in metadata is set to 1 on updating the existing key
  * A test to verify version field in metadata is set to 1 when updating metadata using putMeta method
  * A test to verify version field in metadata is set to 1 when using putAll method
  
* Update the sync unit tests and functional tests to assert the changes. The changes are in the following PR: https://github.com/atsign-foundation/at_server/pull/1158

* AtMetadata.version does not update on the update of a key


packages/at_persistence_secondary_server/lib/src/model/at_metadata_builder.dart
```diff
@@ -26,11 +26,25 @@ class AtMetadataBuilder {
...
     atMetaData!.updatedBy = atSign;
+    // updatedAt indicates the date and time of the key updated.
+    // For a new key, the updatedAt is same as createdAt and on key
+    // update, set the updatedAt to the currentDateTime.
     atMetaData!.updatedAt = currentUtcTime;
     atMetaData!.status = 'active';
+    // The version indicates the number of updates a key has received.
+    // Version is set to 0 for a new key and for each update the key receives,
+    // the version increases by 1
+    (existingMetaData?.version == null)
```
murali-shris, line 45: Can you check in the callers whether existingMetaData is populated from hive. In case if some callers fails to pass existingMetaData for an existing key, then version will be set to 0.

murali-shris (changes requested):

packages/at_persistence_secondary_server/lib/src/model/at_metadata_builder.dart
sitaram-kalluri: Thank you Murali, Except in "putAll" method, existingMetadata is fetched and passed to AtMetadataBuilder. I have added for "putAll" method as well. 
Units tests are also updated for put, putMeta and putAll method to update the key and verify version is set to 1

gkc:
@sitaram-kalluri @VJag @nickelskevin Are createdBy and updatedBy actually required in the metadata? To me it seems that neither of them are, as they are derivable from the key name, since (1) the key name itself provides the information about who created it, and (2) only the atSign which created a key can subsequently update it.

gkc (approved):

murali-shris (approved):

Issue #1212, opened 2023-02-19
Sharing issues after resets


gkc:
### Describe the bug

Let's say the `@alice` atSign is reset. Other atSigns, which had previously securely shared information with `@alice`, are unable to successfully share information with `@alice` after it has been reset. Specifically, the information will reach `@alice`, but `@alice` client will be unable to decrypt it. See [here](https://docs.google.com/document/d/1KQJxJ3gNWmFEqo4lDyo5Fg_O5_0eEBhq9oUU2748NOk/edit#) for more details
### 

### Steps to reproduce

* Have `@bob` securely share information with `@alice`
* Reset `@alice` atServer
* Have `@bob` share some new information with `@alice`
* `@alice` will be unable to decrypt

### Expected behavior

After `@alice` has been reset, `@alice` should be able to decrypt information that `@bob` shares

gkc:
Expect PR ready for review on Feb 21. Implements option 1.2a as described [here](https://docs.google.com/document/d/1KQJxJ3gNWmFEqo4lDyo5Fg_O5_0eEBhq9oUU2748NOk/edit#bookmark=id.h1eaff2raij0)

gkc:
Complete. Requires review. Dropping SP to zero for next sprint

gkc:
Closed by https://github.com/atsign-foundation/at_server/releases/tag/c3.0.28c

Issue #1211, opened 2023-02-19
Cache refresh job does not work


gkc:
### Describe the bug

The cache refresh job silently fails with no effect every time it is run if the atServer contains a cached copy of another atSign's public key. Why does this happen?
* [This other bug](https://github.com/atsign-foundation/at_server/issues/1210) leads to the cached copy of the other atSign's public key having a ttr of 0
  * The original sin here is that the publickey entry is originally created with a value of null, and the proxy lookup verb handler doesn't check the ttr value - it just caches anyway. See this other bug for more detail.
* The cache refresh job, which runs once per day at a random hour, starts by [retrieving all cached keys from the keystore](https://github.com/atsign-foundation/at_server/blob/trunk/packages/at_secondary_server/lib/src/refresh/at_refresh_job.dart#L92-L94)
* As it iterates through the keys, it first does [this check](https://github.com/atsign-foundation/at_server/blob/trunk/packages/at_secondary_server/lib/src/refresh/at_refresh_job.dart#L36-L38) whose intent is to not refresh the key if the ttr is -1.
  ```dart
      if (metadata == null || metadata.ttr == -1) {
        continue;
      }
  ```
* When it finds a key with metadata.ttr != -1 it continues on by [checking if we've reached the refreshAt time](https://github.com/atsign-foundation/at_server/blob/trunk/packages/at_secondary_server/lib/src/refresh/at_refresh_job.dart#L39-L42)
  ```dart
      if (metadata.ttr != -1 &&
          metadata.refreshAt!.millisecondsSinceEpoch > nowInEpoch) {
        continue;
      }
  ```
* But if ttr was originally set as null or zero, then refreshAt is null, and so the code snippet above immediately fails with a `Null check operator used on a null value` error
* This error is uncaught by the refresh job, no messages are logged
* And the error is silently swallowed by the Cron package which we use for scheduled jobs
  ```dart
    Future.microtask(() => _task()).then((_) => null, onError: (_) => null);
  ```

### Steps to reproduce

(Reproducing this takes a long time. Would not recommend.)
* Do the steps described in #1210.
* Set logLevel to FINEST and then wait for the cache refresh job to run.
* You will see in the log file something like `FINEST|2023-02-18 22:00:00.001213|AtRefreshJob|Scheduled Refresh Job started`
* ... and you will not see a log message that says "scheduled Refresh Job completed", because the error described above has been thrown.

### Expected behavior
[...]

PR #1213, opened 2023-02-19
refactor: Refactoring for testability of cache-impacting operations

gkc:
**- What I did**
A **_lot_** of refactoring. But it's just refactoring in this PR - no behaviour changes.
Required for #1211 and #1212

**- How I did it**
* Refactoring
  * Renamed 'lib/src/refresh' to 'lib/src/caching'
  * Renamed at_refresh_job.dart to cache_refresh_job.dart, renamed AtRefreshJob to AtCacheRefreshJob
  * Introduced AtCacheManager
  * Migrated code doing cache operations from AtCacheRefreshJob into AtCacheManager
  * Migrated code doing cache operations functionality from LookupVerbHandler and ProxyLookupVerbHandler into AtCacheManager
  * Removed use of OutboundClientManager.getInstance() from all verb handlers which use it. Test and metrics code still use the singleton but that's OK for now, as AtSecondaryServerImpl injects the singleton into the VerbManagerHandler
    * Purpose of injecting the outboundClientManager into the verb handlers is so that we can test the verb handlers in unit tests with mock outbound clients.
  * Eliminated use of DefaultVerbHandlerManager().getInstance()
  * Now injecting the AtCacheManager into the verb handlers that need it
  * Removed obsolete SyncFromVerbHandler and SyncVerbHandler and all references to them    
* **And also ...**
  * much renaming of variables, fixing imports, adding types to variables (or removing types from variables) etc to eliminate lint warnings
  * much updating of test code to inject mockKeyStores where previously they were setting `null` etc
* **And also ...**
  * build: Updated the config.yaml in tools/build_virtual_environment to bring it up to date with the one in packages/at_secondary_server
  * fix: Updated the usage() info for the update verb in packages/at_server_spec
  * test: Updated tests in at_secondary_server so they use the usage message from the verb class itself rather than having that message hard-coded in tests. (i.e. test is no longer coupled to the precise value in the update verb class in at_server_spec

**- How to verify it**
Tests pass

refactor: Refactoring for testability of cache-impacting operations


cconstab (approved):
Approved after chatting with @gkc

PR #1215, opened 2023-02-21
fix: corrected assertions of the scan verb test

### LINKED ISSUE #1194: some tests in scan_verb_test are void due to faulty logic and assertions
### Describe the bug

Lines 133, 134: Assertion will never be true because (1) the response will never contain 'update:public' (2) the key will be in lowercase
```dart
    expect(
        false, response.contains('"update:public:verifyingTTL$firstAtsign"'));
```

Additionally, it's not clear how the test is intended to work. These lines 126 to 129 suggest that by the time the scan is issued, the ttl has not yet passed, and so the key SHOULD be returned from a scan
```dart
    //update ttl so that key is expired
    await socket_writer(socketFirstAtsign!,
        'update:ttl:6000:public:verifyingTTL$firstAtsign 1');
    sleep(Duration(seconds: 2));
    await socket_writer(socketFirstAtsign!, 'scan');
```

Lines 156, 157: Assertion will never be true because (1) the response will never contain 'update:public' (2) the key will be in lowercase
```dart
    expect(
        false, response.contains('"update:public:verifyingTTB$firstAtsign"'));
```

### Steps to reproduce

Look at the code

### Expected behavior

Tests should have good logic and assertions. For example in the ttl test we could first test the positive case and then the negative, and start by setting a TTL of 6000, doing a scan, asserting the key IS returned; then set a TTL of 200, sleep for 250, do a scan, and assert that the key is NOT returned. Similar changes for ttb verifying the negative and then the positive case.

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

_No response_

### LINKED ISSUE #1216: keys with capital letters doesn't show up in scan response
### Describe the bug

A key that contains a capital letter doesn't show up in the scan verb response.

### Steps to reproduce

1. First create a key that has a capital letter in it. 
example: @alice🛠@update:public:capS@alice🛠 check

2. Then do a scan

@alice🛠@update:public:capS@alice🛠 check
data:4
@alice🛠@scan
data:["@alice🛠:signing_privatekey@alice🛠","public:publickey@alice🛠","public:signing_publickey@alice🛠"]

### Expected behavior

Scan verb result should display the key 

### Screenshots

Screenshot from 2023-02-21 18-21-58

### Smartphones

_No response_

_No response_

### Additional context

_No response_

purnimavenkatasubbu:
**- What we did**
* Changed the incorrect assertions
* Added test case for keys with mixed case - discovered #1216 
* Fixed #1216

**- How we did it**
* Removed the update: public command in the scan result assertion
* Ensure HiveKeystore public methods start by converting their 'key' parameter to lower case. This ensures that the _metaDataCache's keys are in lower-case, same as the keyStore itself.

**- How to verify it**
* all tests should pass

### DIFF
--- a/tests/at_functional_test/test/scan_verb_test.dart
+++ b/tests/at_functional_test/test/scan_verb_test.dart
@@ -120,7 +120,7 @@ void main() {
 
     ///UPDATE VERB
     await socket_writer(
-        socketFirstAtsign!, 'update:ttl:3000:ttlkey.me$firstAtsign 1245');
+        socketFirstAtsign!, 'update:ttl:3000:ttlKEY.me$firstAtsign 1245');
     var response = await read();
     print('update verb response : $response');
     assert(
@@ -130,11 +130,12 @@ void main() {
     await socket_writer(socketFirstAtsign!, 'scan');
     response = await read();
     print('scan verb response : $response');
-    expect(response, contains('"ttlkey.me$firstAtsign"'));
+    expect(false, response.contains('"ttlKEY.me$firstAtsign"')); // server ensures lower-case
+    expect(true, response.contains('"ttlkey.me$firstAtsign"'));
 
     // update ttl to a lesser value so that key expires for scan
     await socket_writer(
-        socketFirstAtsign!, 'update:ttl:200:ttlkey.me$firstAtsign 1245');
+        socketFirstAtsign!, 'update:ttl:200:ttlKEY.me$firstAtsign 1245');
     response = await read();
     assert(
         (!response.contains('Invalid syntax')) && (!response.contains('null')));
@@ -145,6 +146,7 @@ void main() {
     response = await read();
     print('scan verb response : $response');
     expect(false, response.contains('"ttlkey.me$firstAtsign"'));
+    expect(false, response.contains('"ttlKEY.me$firstAtsign"'));
   }, timeout: Timeout(Duration(seconds: 120)));
 
   test('Scan verb does not return unborn keys', () async {
--- a/tests/at_functional_test/test/scan_verb_test.dart
+++ b/tests/at_functional_test/test/scan_verb_test.dart
@@ -120,18 +120,31 @@ void main() {
 
     ///UPDATE VERB
     await socket_writer(
-        socketFirstAtsign!, 'update:public:verifyingTTL$firstAtsign Working?');
+        socketFirstAtsign!, 'update:ttl:3000:ttlkey.me$firstAtsign 1245');
     var response = await read();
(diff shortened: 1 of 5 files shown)


tests/at_functional_test/test/scan_verb_test.dart
```diff
@@ -120,18 +120,31 @@ void main() {
 
     ///UPDATE VERB
     await socket_writer(
-        socketFirstAtsign!, 'update:public:verifyingTTL$firstAtsign Working?');
+        socketFirstAtsign!, 'update:ttl:3000:ttlkey.me$firstAtsign 1245');
```
gkc: Can you modify this test so that the entity is 'ttlKey.me' (i.e. capitalize the k) and modify the expects accordingly? Would be good to be asserting that in the scan response, the entity is in lowercase

gkc (changes requested):

tests/at_functional_test/test/scan_verb_test.dart
purnimavenkatasubbu: sure

purnimavenkatasubbu: Updated the test with a capital K. But the scan response doesn't show up the key. Raised a bug for the same.
[bug](https://github.com/atsign-foundation/at_server/issues/1216)

gkc: You haven't changed the expects to match :-)

gkc: Oh I see, you didn't commit the changes

gkc:
@purnimavenkatasubbu I'm pushing the fix for #1216 to this PR. I've updated the scan_verb_test also. Once the checks have passed, I will publish new at_persistence_secondary_server, remove the dependency override from at_secondary_server's pubspec, push the branch again, wait for checks to pass and then merge to trunk.

tests/at_functional_test/test/scan_verb_test.dart
purnimavenkatasubbu: Yes @gkc . I haven't pushed my changes because the test will fail in actions

purnimavenkatasubbu:
> @purnimavenkatasubbu I'm pushing the fix for #1216 to this PR. I've updated the scan_verb_test also. Once the checks have passed, I will publish new at_persistence_secondary_server, remove the dependency override from at_secondary_server's pubspec, push the branch again, wait for checks to pass and then merge to trunk.

Thanks

gkc (approved):

PR #1219, opened 2023-02-22
fix: Make plookup:all command work correctly if the data has not previously been cached in the server.

gkc:
**Note**
1. the new unit tests which will prevent this from regressing are in another branch which will be in a follow-up PR
2. The version change in pubspec.yaml is because I messed up the version yesterday in [this PR](https://github.com/atsign-foundation/at_server/pull/1215) when I set it to 3.0.48 instead of 3.0.28. (3.0.27 is in canary and is currently being rolled out to prod)

**- What I did**
fix: Make plookup:all command work correctly if the data has not previously been cached in the server.

**- How I did it**
Previously, a plookup:all would fetch the remote value and cache it, but would return 'null' to the client because prepareResponseData wasn't being called when 'operation:all' was supplied. (Note that plookup:meta and just plain plookup both worked fine)
- Removed if statement in the ProxyLookupVerbHandler which was not preparing any response data if (1) the value was being looked up remotely rather than fetched from cache, and (2) the operation was 'all' (as in, plookup:all in this case).
  - Just a guess, but it seems highly likely that that if statement was added because prepareResponseData would previously crash when calling toJson on the AtData object constructed from the remote lookup. I've fixed that bug in [the PR just before this one.](https://github.com/atsign-foundation/at_server/pull/1218)
- Added optional parameter to SecondaryUtil.prepareResponseData so it can add a 'key' value to the JSON response in order to be consistent with the response to a plookup:all when the data is retrieved from cache.

**- How to verify it**
Tests pass

fix: Make plookup:all command work correctly if the data has not previously been cached in the server.

### DIFF
--- a/packages/at_secondary_server/lib/src/utils/secondary_util.dart
+++ b/packages/at_secondary_server/lib/src/utils/secondary_util.dart
@@ -99,7 +99,7 @@ class SecondaryUtil {
     return base64Encode(signature);
   }
 
-  static String? prepareResponseData(String? operation, AtData? atData) {
+  static String? prepareResponseData(String? operation, AtData? atData, {String? keyToUseIfNotAlreadySetInAtData}) {
     String? result;
     if (atData == null) {
       return result;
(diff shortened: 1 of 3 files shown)


nickelskevin (approved):
It is so easy to review these when they are done incrementally

PR #1220, opened 2023-02-23
fix: Refactor notify_verb_handler.dart

sitaram-kalluri:
**- What I did**
* Refactored notify_verb_handler.dart which addressess the below mentioned issues:
     * At the moment, most of the code in notify_verb_handler.dart is in processVerb method which is a bit difficult in terms of readability.
     * For simple notifications which does not include caching of keys, on the receiver side, we still check the logic related to caching of keys.
     *  The metadata in the atNotification bean is not set when messageType is key.

**- How I did it**

* Refactoring changes are as follows:
    * Add below IF condition to avoid running through caching keys "IF" conditions for simple notifications
         * ```dart
            if (operationType != OperationType.delete && (_getTimeToRefresh(verbParams[AT_TTR]) == null)) {
                   response.data = 'data:success';
                   return;
            }
         
    * Fix the issue where metadata is not set to atNotification.metadata when message type is "Key"
    * Moved the logic to populate the AtNotificationBuilder to a new private method "_populateNotificationBuilder"
    * Moved the logic to populate the AtMetadata to a new private method "_getAtMetadataForNotification"
    * Move the validation logic and logic to convert the String to respective enums to separate private methods
         
**- How to verify it**

* Added unit tests to assert the following:
   * Convert String representation of "OperationType", and "MessageType" to enum's 
   * Validations of setting the default value of notification expiry.
* All the existing tests have to pass.
* Uncomment and remove the skip tag on the existing test cases since the changes are merged to the trunk branch.

* chore: Refactor notify_verb_handler.dart to increase readability and fix minor issues


gkc:
I wonder if we should just fix this in the client, and leave server code unchanged?

gkc (changes requested):
As discussed - please keep the refactoring changes (good work!) but revert changes to how the "key" in the notification is populated

packages/at_secondary_server/lib/src/verb/handler/notify_verb_handler.dart
```diff
@@ -57,260 +65,163 @@ class NotifyVerbHandler extends AbstractVerbHandler {
...
+    }
+    // Store the notification to the notification keystore.
+    await NotificationUtil.storeNotification(atNotificationBuilder.build());
+    OperationType operationType = getOperationType(verbParams[OPERATION]);
+    // When Operation is update, cache key only when TTR is set.
+    // So if TTR is null,  do nothing.
+    // Also, If operation is delete removed the cached key - irrespective of TTR value.
+    // So, If operation is not delete and TTR is null, return.
+    if (operationType != OperationType.delete &&
+        (_getTimeToRefresh(verbParams[AT_TTR]) == null)) {
```
gkc, line 120: a value of 0 is equivalent to a value of null for ttr, please check for null or zero

gkc (changes requested):
LGTM just one minor change please

packages/at_secondary_server/lib/src/verb/handler/notify_verb_handler.dart
sitaram-kalluri: The method internally calls "validateTTR" in atmetadata_utils.dart in at_utils package which returns null when ttr is either "null" or "0"

```dart
if (ttr_ms == null || ttr_ms == 0) {
      return null;
    }
```

gkc: ah I see.

gkc (approved):

PR #1241, opened 2023-03-04
fix: auto_notify may be setting ttl incorrectly

### LINKED ISSUE #1175: UpdateVerbHandler may be setting expiresAt incorrectly on the auto-notification it generates
### Describe the bug

See this code snippet from the `_notify` method of `UpdateVerbHandler`
```dart
    DateTime? expiresAt;
    if (atMetaData.ttl != null) {
      expiresAt = DateTime.now().add(Duration(seconds: atMetaData.ttl!));
    }
```

(1) notification expiry time is not currently expected to be coupled with the underlying data entry's ttl
(2) even if we were coupling the expiry time, `ttl` is in _milliseconds_ and we are treating it here as _seconds_

### Expected behavior

* `expiresAt` on the auto-notification should be as per at_protocol specification
* If the specification is unclear then we should clarify the specification and make the implementation match it.

srieteja:
closes https://github.com/atsign-foundation/at_server/issues/1175

**- What I did**
- remove the implementation that sets expiresAt for auto_notifications in update_verb_handler
- set standard ttl of 1440 minutes for all auto_notifications in update_verb_handler
- change the config value of notificationExpiryInMins in _config.yaml_ from 15mins -> 1440 minutes(24 hrs)
- fix: auto_notifications contain only key part of atkey instead of _sharedWith:key@sharedBy_
- make _notify() in update_verb_handler visibleForTesting [now notify()]
- change the return type of notify() from void to dynamic, to make it testable
- notify() now returns the notification that has been created. Although this does not change the flow of any process and the notification returned will not be part of any response to the user. It can only be used when notify() is called independently, this way testing this method should be possible.

**- How to verify it**
- added tests in at_notification_test(at_persistence_seconary) and update_verb_test(at_secondary_server)

fix: auto_notify may be setting ttl incorrectly
fix: auto_notifications contain only key part of atkey instead of _sharedWith:key@sharedBy_


packages/at_persistence_secondary_server/test/at_notification_test.dart
```diff
@@ -18,5 +18,20 @@ void main() async {
...
+      var notifExpiresAt = DateTime.now().toUtc().add(Duration(minutes: 15));
+      //when ttl is not passed, 15-mins is used as the default ttl
+      //notification.expiresAt and notifExpiresAt have the difference of a
+      // couple of milli seconds and they cannot asserted to be equal
+      expect(notification.expiresAt?.year, notifExpiresAt.year);
+      expect(notification.expiresAt?.month, notifExpiresAt.month);
+      expect(notification.expiresAt?.day, notifExpiresAt.day);
+      expect(notification.expiresAt?.hour, notifExpiresAt.hour);
+      expect(notification.expiresAt?.minute, notifExpiresAt.minute);
+      expect(notification.expiresAt?.second, notifExpiresAt.second);
```
gkc, line 34: As an alternative, why not assert that the milisecondSinceEpoch of both are within some number of milliseconds of each other? Seems like this test could flake if the second ticks over during the test

gkc: As an alternative, why not replace all of the above assertions with just one assertion that the milisecondSinceEpoch of both are within some number of milliseconds of each other? Seems like this test could flake if the second ticks over during the test

packages/at_secondary_server/test/update_verb_test.dart
```diff
@@ -703,16 +703,45 @@ void main() {
...
+      expect(autoNotification.id, isNotNull);
+      expect(autoNotification.ttl, ttlInMillis);
+      //autoNotification.expiresAt and notifExpiresAt have the difference of a
+      // couple of milli seconds and they cannot asserted to be equal
+      expect(autoNotification.expiresAt?.year, notifExpiresAt.year);
+      expect(autoNotification.expiresAt?.month, notifExpiresAt.month);
+      expect(autoNotification.expiresAt?.day, notifExpiresAt.day);
+      expect(autoNotification.expiresAt?.hour, notifExpiresAt.hour);
+      expect(autoNotification.expiresAt?.minute, notifExpiresAt.minute);
+      expect(autoNotification.expiresAt?.second, notifExpiresAt.second);
```
gkc: Same request as above

gkc (changes requested):
LGTM - couple of small changes to the tests please

packages/at_persistence_secondary_server/test/at_notification_test.dart
srieteja: this change has been made as part of [7dc827f](https://github.com/atsign-foundation/at_server/pull/1241/commits/7dc827f5a9ef7b2c738c0ccf78c45a29c114a368)

packages/at_secondary_server/test/update_verb_test.dart
srieteja: this change has been made as part of [7dc827f](https://github.com/atsign-foundation/at_server/pull/1241/commits/7dc827f5a9ef7b2c738c0ccf78c45a29c114a368)

gkc (approved):
LGTM

PR #1246, opened 2023-03-08
fix: ensure that a sync request always returns at least one entry if there is at least one entry to return

### LINKED ISSUE #1227: SyncProgressive's buffer size and InboundMessageListener's buffer size out of step
### Describe the bug

- SyncProgressive's buffer size and InboundMessageListener's buffer size out of step
- InboundMessageListener has a hard-coded value
- SyncProgressive's buffer size is set by config
- When InboundMessageListener's buffer size is larger (currently it is nearly twice as large) then it is possible for a client to send an update with a data size greater than sync's buffer limit, which subsequently cannot be sync'd back to a client

### Steps to reproduce

* Start a server with a buffer size of 10 (10 bytes)
* Do an update
* sync will never return anything

### Expected behavior

InboundMessageListener buffer capacity should be set to 10KB less than the sync buffer size

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

_No response_

srieteja:
closes https://github.com/atsign-foundation/at_server/issues/1227

**What I did**
- In case an entry is larger than the sync's buffer size (i.e. the maximum desired sync response size), it would never be synced. To resolve this, a fix has been made to ensure that at least one entry will always be synced
- refactored code in SyncProgressiveVerbHandler, created a new method prepareResponse() which allows us to test the response preparation in isolation from the verb handler as a whole

**How I did it**
- prepareResponse now adds items from the commitEntryIterator to the syncResponse until either
  1. there is at least one item in the syncResponse, and the response length is greater than the desired max syncResponse length, or
  3. there are AtSecondaryConfig.syncPageLimit items in the syncResponse

**How I verified it**
- added test cases which exercise all of the conditions in prepareResponse()
- tested with a snapshot of the real storage from the `@0living` atSign taken as of a few days ago
  - with normal syncBufferSize of 5,242,880 bytes ([0living.normalSyncBufferSize.fullSync.log](https://github.com/atsign-foundation/at_server/files/10961710/0living.normalSyncBufferSize.fullSync.log))
  - with syncBufferSize of 100 bytes ([0living.tinySyncBufferSize.fullSync.log](https://github.com/atsign-foundation/at_server/files/10961718/0living.tinySyncBufferSize.fullSync.log))

fix: ensure that a sync request always returns at least one entry if there is at least one entry to return


packages/at_secondary_server/lib/src/verb/handler/sync_progressive_verb_handler.dart
```diff
@@ -41,44 +41,56 @@ class SyncProgressiveVerbHandler extends AbstractVerbHandler {
...
         }
         keyStoreEntry.value = atData.data;
         keyStoreEntry.atMetaData = _populateMetadata(atData);
       }
       // If syncBuffer reaches the limit, break the loop.
-      if (syncBuffer.isOverFlow(utf8.encode(jsonEncode(keyStoreEntry)))) {
+      //[syncBuffer.length() > 0]  s to ensure that atleast one entry
+      // will be synced even though data is overflowing the buffer
+      if (syncResponse.isNotEmpty &&
+          syncBuffer.isOverFlow(utf8.encode(jsonEncode(keyStoreEntry)))) {
```
srieteja: @gkc do you think we need an additional check to ensure that the current entry is not overflowing the inbound_message listener's buffer? (since we are trying to sync one entry in any case)

gkc: An additional check? No. The check as it is is fine. If you're asking if we should remove that check, the answer is also no - servers must be able to control how much data they are prepared to receive.

packages/at_secondary_server/lib/src/verb/handler/sync_progressive_verb_handler.dart
```diff
@@ -41,44 +41,56 @@ class SyncProgressiveVerbHandler extends AbstractVerbHandler {
...
     // and when syncResponse length equals the [AtSecondaryConfig.syncPageLimit]
-    while (itr.moveNext() &&
+    await populateSyncBuffer(syncBuffer, syncResponse, commitEntryIterator);
+
+    response.data = jsonEncode(syncResponse);
+  }
+
+  @visibleForTesting
+  Future<void> populateSyncBuffer(ByteBuffer syncBuffer,
+      List<dynamic> syncResponse, Iterator<dynamic> commitEntryInterator) async {
```
gkc: Typo in `commitEntryInterator`

gkc (changes requested):

packages/at_secondary_server/lib/src/verb/handler/sync_progressive_verb_handler.dart
srieteja: I agree we should have the check. I'm asking if we need to check if this entry is overflowing the inbound_listener's buffer since we are overriding the syncBuffer in some cases.

Say there is a very large entry which has a size greater than syncBuffer and inboundListener's buffer, but when it's the only entry in the queue we would try to sync it with the current change in this PR which would fail in the later stages. 

Should we put a check, something like (inboundListenerBuffer.isOverflow(syncBuffer)) so that when we are overriding the syncBuffer size, we ensure that it does not exceed the inboundListener's limit?

gkc: No. Client handling of sync responses is a separate issue

gkc: And if you're talking about the server's inbound listener - again, no. 

The point of this PR is to ensure that sync will never get stuck because of this issue. Let's just get this done now and then you can think about how the broader issues if you like

packages/at_secondary_server/lib/src/verb/handler/sync_progressive_verb_handler.dart
srieteja: Rectified as part of [8fb60c0](https://github.com/atsign-foundation/at_server/pull/1246/commits/8fb60c06b05397811f6d46af460042a0115f4e4f)

gkc (approved):

PR #1255, opened 2023-03-13
fix: Cleaned up exception handling in a few places

gkc:
**- What I did**
fix: Cleaned up exception handling in a few places
fix: Made DummyInboundConnection.getSocket throw an UnimplementedError rather than an un-typed throw
fix: Added a check to the InboundConnectionImpl.equals method to prevent DummyInboundConnection.getSocket being thrown

**- How to verify it**
Tests pass


VJag (approved):
LGTM

PR #1257, opened 2023-03-15
fix: Check if sharedBy atSign is authorized to send notifications

### LINKED ISSUE #1180: NotifyVerbHandler needs extra check
### Describe the bug
* It is currently possible for a malicious atServer `@bob` to craft notifications for `@alice` which appear to come from `@colin` rather than from `@bob`
* Faking the originator of forText notifications is not currently possible
* Faking the originator of forUpdate notifications is partially possible. For example this notification was delivered to the `@bob` atClient having originated from the `@alice` atServer
```
notification: {"id":"9ce26891-143e-45b6-9390-ef63de7b1737","from":"@alice","to":"@bob","key":"@bob:1675611466053347.testing@colin","value":"@alice was here, pretending to be @colin","operation":"null","epochMillis":1675611466062,"messageType":"MessageType.key","isEncrypted":false}
```
* As you can see, `from` is correctly set to `@alice`, so it's not possible to fake that part. However, you can see that the key appears to be shared by `@colin`

### Expected behavior
* NotifyVerbHandler should have explicit checks to ensure that forUpdate notifications are only for keys which are sharedBy the other atServer's atSign

sitaram-kalluri:
**- What I did**
* Fix the git issue: https://github.com/atsign-foundation/at_server/issues/1180

**- How I did it**
* Add "IF" condition to check if sharedBy atSign is currentAtSign. If yes, allow to send notifications, else throw UnAuthorized exception

**- How to verify it**
* Added unit test to verify UnAuthrizedException is thrown when sharedBy is not equal to currentAtSign
* The existing notification tests should pass

* Check if sharedBy atSign is authorized to send notifications


packages/at_secondary_server/lib/src/verb/handler/notify_verb_handler.dart
```diff
@@ -71,7 +71,8 @@ class NotifyVerbHandler extends AbstractVerbHandler {
       var currentAtSign = AtSecondaryServerImpl.getInstance().currentAtSign;
       // If '@' is missing before an atSign, the formatAtSign method prefixes '@' before atSign.
       verbParams[FOR_AT_SIGN] = AtUtils.formatAtSign(verbParams[FOR_AT_SIGN]);
-      verbParams[AT_SIGN] = AtUtils.formatAtSign(verbParams[AT_SIGN]);
+      verbParams[AT_SIGN] =
+          AtUtils.formatAtSign(verbParams[AT_SIGN]) ?? currentAtSign;
```
murali-shris: Srieteja recently merged a PR. since there were two methods formatAtSign and fixAtSign, I suggested to deprecated formatAtSign and use fixAtSign. Please uptake latest version of at_utils and change this method to fixAtSign
https://github.com/atsign-foundation/at_tools/pull/300

murali-shris (dismissed):

packages/at_secondary_server/lib/src/verb/handler/notify_verb_handler.dart
sitaram-kalluri: Sure Murali.

sitaram-kalluri: @murali-shris: I did look for the usages and there are around 20 usages in at_secondary_server. Is it fine to do these changes in a separate PR?

gkc:
I just tried to take up at_utils and do the global search and replace and discovered that fixAtSign is not a straight replacement of formatAtSign. fixAtSign takes a `String` as argument and returns a `String`. formatAtSign takes a `String?` as argument and returns a `String?`

@srieteja @murali-shris @sitaram-kalluri can you please get together to discuss this and figure out what is the right way forward?

Meanwhile @sitaram-kalluri I'll review this PR as is

packages/at_secondary_server/lib/src/verb/handler/notify_verb_handler.dart
gkc: This doesn't look right. Why are you setting `verbParams[AT_SIGN]` to currentAtSign if `verbParams[AT_SIGN]` wasn't supplied?

packages/at_secondary_server/lib/src/verb/handler/notify_verb_handler.dart
```diff
@@ -439,4 +446,8 @@ class NotifyVerbHandler extends AbstractVerbHandler {
     }
     return '${verbParam[FOR_AT_SIGN]}:${verbParam[AT_KEY]}${verbParam[AT_SIGN]}';
   }
+
+  bool _isAuthorizedToSendNotification(String? sharedBy, String currentAtSign) {
+    return sharedBy == currentAtSign;
```
gkc, line 455: As a result of the code above, `verbParams[AT_SIGN] =
          AtUtils.formatAtSign(verbParams[AT_SIGN]) ?? currentAtSign;` this check will always succeed if the sharedBy atSign is not supplied in the verb params

gkc (changes requested):

packages/at_secondary_server/lib/src/verb/handler/notify_verb_handler.dart
sitaram-kalluri: When messageType is "text", by syntax, sharedBy atSign is not populated. So, false will be returned when validating if atSign is authorized to send notifications (null not equal to currentAtSign). Hence verbParam[AT_SIGN] is set to currentAtSign. 
When messageType is "text", verbParams[AT_SIGN] is not added to the notification key.

gkc: But at this point you haven't checked if you are in an authenticated, polAuthenticated or un-authenticated session

sitaram-kalluri: Is it ok to move the logic to set currentAtSign to verbParams[AT_SIGN] into "_handleAuthenticatedConnection" method?

gkc: Yes I think so.

Just make sure your overall logic works for auth or polAuth connections, and maybe throw an exception for unauthenticated connections

sitaram-kalluri: Sure Gary.

gkc (approved):
LGTM

PR #1274, opened 2023-03-21
fix: update inbound/outbound message listeners exception text

### LINKED ISSUE #1205: Insufficiently informative message from server when request payload from client is too large
### Describe the bug

Currently we create this exception in InboundMessageListener
```
BufferOverFlowException('buffer overflow')
```
This gets back to the client as  an AT0005 error (Buffer limit exceeded) with the message "Buffer limit exceeded: buffer overflow" ... we can and should provide a more informative error message.

### Expected behavior

Need a better message. It would be helpful to see something on the client side like
"Buffer limit exceeded: server received request which was longer than the maximum of <N> bytes. Terminating the connection."

srieteja:
closes https://github.com/atsign-foundation/at_server/issues/1205

**- What I did**
- chore: update inbound/outbound message listeners exception text
- previously inbound buffer overflow exception text was _buffer overflow_ changed to -> _InboundBuffer overflow: server received request which exceeded the buffer size limit. Terminating the connection._
- previously outbound buffer overflow exception text was _buffer overflow_ changed to -> OutboundBuffer overflow: server received request which exceeded the buffer size limit. Terminating the connection._

chore: update inbound/outbound message listener's exception text

### DIFF
--- a/packages/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
+++ b/packages/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
@@ -50,7 +50,9 @@ class OutboundMessageListener {
       }
     } else {
       _buffer.clear();
-      throw BufferOverFlowException('Buffer overflow on outbound connection');
+      throw BufferOverFlowException('OutboundBuffer overflow: server received'
+          ' request which was longer than the maximum of bytes.'
+          ' Terminating the connection.');
     }
     if (_buffer.isEnd()) {
       result = utf8.decode(_buffer.getData());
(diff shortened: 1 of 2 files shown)


packages/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
```diff
@@ -50,7 +50,9 @@ class OutboundMessageListener {
       }
     } else {
       _buffer.clear();
-      throw BufferOverFlowException('Buffer overflow on outbound connection');
+      throw BufferOverFlowException('OutboundBuffer overflow: server received'
```
murali-shris: server sent.. not received in outbound message listener

murali-shris (dismissed):

packages/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
srieteja: message has been updated in the latest commit

murali-shris (approved):

PR #1290, opened 2023-03-30
fix: add optional flag to generate at_demo_data.dart script and skip …

sitaram-kalluri:
**- What I did**
* Fix the failing e2e tests. 
* Please refer to the run: https://github.com/atsign-foundation/at_server/actions/runs/4559428003 which has all the tests pass successfully.

**- How I did it**
* Added an optional flag `-p pkam` to the `dart bin/generate_at_demo_data.dart` command.

**- How to verify it**
* All the other tests are running successfully.

### DIFF
--- a/.github/workflows/e2e_test.yaml
+++ b/.github/workflows/e2e_test.yaml
@@ -190,7 +190,7 @@ jobs:
           --header 'Authorization: ${{secrets.NODE_API_DELETE}}' \
           --header 'Content-Type: application/json' \
           --data-raw '{
-              "atsign" : "ATSIGN_2_NAME"
+              "atsign" : "$ATSIGN_2_NAME"
           }'
         env:
           ATSIGN_1_NAME: ${{ steps.atsign_names.outputs.ATSIGN_1_NAME }}   
--- a/.github/workflows/e2e_test.yaml
+++ b/.github/workflows/e2e_test.yaml
@@ -201,3 +201,19 @@ jobs:
         env:
           AT_SIGN_1: ${{ steps.atsign_names.outputs.AT_SIGN_1 }}   
           AT_SIGN_2: ${{ steps.atsign_names.outputs.AT_SIGN_2 }}  
+
+      # Verify if Atsign exists
+      - name: Verify if Atsign exists
+        run: |  
+          sleep 20
+          AT_SIGN_1_HOST_RESP=$((echo $AT_SIGN_1; sleep 1) | openssl s_client -connect root.atsign.wtf:64 2>/dev/null | grep --color=none "^@.*:" | cut -d'@' -f2)
+          AT_SIGN_2_HOST_RESP=$((echo $AT_SIGN_2; sleep 1) | openssl s_client -connect root.atsign.wtf:64 2>/dev/null | grep --color=none "^@.*:" | cut -d'@' -f2)
+          if [ -z "$AT_SIGN_1_HOST_RESP" ] && [ -z "$AT_SIGN_2_HOST_RESP" ]; then
+            echo "Atsign deleted successfully"
+          else
+            echo "Atsign still exists"
+            exit 1
+          fi          
+        env:
+          AT_SIGN_1: ${{ steps.atsign_names.outputs.AT_SIGN_1 }}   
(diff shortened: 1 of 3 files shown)


gkc (changes requested):
We shouldn't mark the test as skipped; however, by design right now, we don't want to be able to switch on testingMode on these new servers. Can you instead find a way to skip it just in this situation? Could base it on an initial check to see if testingMode  is enabled or not.

kumarnarendra701 (approved):

sitaram-kalluri:
> We shouldn't mark the test as skipped; however, by design right now, we don't want to be able to switch on testingMode on these new servers. Can you instead find a way to skip it just in this situation? Could base it on an initial check to see if testingMode is enabled or not.

@gkc: Modified the test which checks if the testing mode is enabled on the secondary server. Test executes only if enabled, else skips.

Below is the test result of the bypass cache test when the testing mode is disabled:
```
test/cached_key_test.dart: update-lookup verb passing bypassCache 
  Clearing socket response queues
  command sent: config:print:autoNotify
  Response: data:testing mode disabled by default
  Testing mode is not enabled on the secondary server: @test7bbaa28e_5bd5_4332_97fd_717b868d77901. So skipping the test
```

tests/at_end2end_test/test/cached_key_test.dart
```diff
@@ -143,12 +142,22 @@ void main() {
   try {
     test('update-lookup verb passing bypassCache ', () async {
       ///Update verb on atsign_1
+      // Check if TestingMode is enabled on the secondary server
+      // If yes allow the test to run, else skip
+      await sh1.writeCommand('config:print:autoNotify\n');
+      String response = await sh1.read();
+      // If the response is not true, testingMode is NOT enabled on the secondary server
+      // So, do not run the test.
+      if (!response.contains('true')) {
```
gkc: A more exact comparison would be better. Ideally an equality something like `if (!response == 'data:true')` and just for completeness compare with `data:false` also

.github/workflows/e2e_test.yaml
```diff
@@ -201,3 +201,19 @@ jobs:
...
+      # Verify if Atsign exists
+      - name: Verify if Atsign exists
+        run: |  
+          sleep 20
+          AT_SIGN_1_HOST_RESP=$((echo $AT_SIGN_1; sleep 1) | openssl s_client -connect root.atsign.wtf:64 2>/dev/null | grep --color=none "^@.*:" | cut -d'@' -f2)
+          AT_SIGN_2_HOST_RESP=$((echo $AT_SIGN_2; sleep 1) | openssl s_client -connect root.atsign.wtf:64 2>/dev/null | grep --color=none "^@.*:" | cut -d'@' -f2)
+          if [ -z "$AT_SIGN_1_HOST_RESP" ] && [ -z "$AT_SIGN_2_HOST_RESP" ]; then
+            echo "Atsign deleted successfully"
+          else
+            echo "Atsign still exists"
```
gkc, line 214: Would be nice to have it echo the atSigns in the message also. Makes it easier to debug if / when it happens

gkc (changes requested):

tests/at_end2end_test/test/cached_key_test.dart
sitaram-kalluri: Replaced the IF condition as you suggested. But when testingMode is disable, the is response is not `data:false`. We get the following response. So not adding the check with `data:false`

```
@sitaram@config:print:autoNotify
data:testing mode disabled by default
```

.github/workflows/e2e_test.yaml
sitaram-kalluri: @kumarnarendra701: Can you please take care of this?

tests/at_end2end_test/test/cached_key_test.dart
gkc: My point was that autoNotify might be set to false ... i.e. if testingMode is enabled, you might get a response of true or false; but if testingMode is disabled, you get a response like "testing mode disabled by default"

sitaram-kalluri: Got it Gary, Updated the IF condition and pushed the changes.

gkc (approved):

PR #1322, opened 2023-04-27
fix: Enhance the AtConnection log messages

### LINKED ISSUE #64: Additional information in client config

One current complication we face when diagnosing issues is that we don't have any way of distinguishing between the many different client programs which connect to the atServer.

In addition to the changes proposed in #63 (add a client ID) we should also allow the sending of additional information in the client config such as the name of the app, the version of the app, the platform upon which the app is running, etc

_No response_

### Additional context

See also #63 

This has been created as a separate issue for clarity but we could choose to implement them at the same time

sitaram-kalluri:
**- What I did**
* Introduce clientId, appName, appVersion and platform in the InboundConnectionMetadata
* Add the above fields to the server logs to distinguish the request that comes from several clients.

**- How I did it**
* Added new fields into InboundConnectionMetadata
* Add "AtConnectionMetadataLogging" as extension to the AtSignLogger.
* Overwrite the "write" method in InboundBoundConnectionImpl and OutBoundConnectionImpl to log the request and responses.
* Add new fields in log messages of InboundMessageListener, OutboundMessageListener and in AtSecondaryServer.

### DIFF
--- a/packages/at_secondary_server/CHANGELOG.md
+++ b/packages/at_secondary_server/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 3.0.31
+- feat: Introduce clientId, appName, appVersion and platform to distinguish requests from several clients in server logs.
 ## 3.0.30
 - fix: When metadata attributes are not set, merge the existing metadata attributes
 - fix: When metadata attributes are explicitly set to null, reset the metadata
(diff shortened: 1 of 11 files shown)


gkc (changes requested):
Have merged your at_commons PR

Please can you publish new version of at_commons please, then update pubspec in this PR and re-request review

gkc (approved):

PR #1344, opened 2023-05-09
fix: fetch latest commitId of each key

sitaram-kalluri:
**- What I did**
* Extend the "stats" verb to return the "latestCommitId" and "commitOp" of each key.

**- How I did it**
* Create a new class "LatestCommitEntryOfEachKey" that extends "MetricProvider" class which provides the latest commit-Id and commitOp of each key

**- How to verify it**
* Added a unit test that asserts the following
  * The latest commit-Id of each key is returned.
  * The latest commit-Op. of each key is returned
  
* Enhance the stats verb to return the latest CommitEntry of each key.


packages/at_secondary_server/lib/src/verb/metrics/metrics_impl.dart
```diff
@@ -479,3 +481,27 @@ class NotificationCompactionStats implements MetricProvider {
...
+    var responseMap = <String, Map<String, dynamic>>{};
+    var atCommitLog = await (AtCommitLogManagerImpl.getInstance()
+        .getCommitLog(AtSecondaryServerImpl.getInstance().currentAtSign));
+
+    var commitEntryIterator = atCommitLog!.getEntries(-1);
+
+    while (commitEntryIterator.moveNext()) {
+      responseMap[commitEntryIterator.current.key] = {
+        'commitId': commitEntryIterator.current.value.commitId,
+        'commitOp': commitEntryIterator.current.value.operation.toString()
```
gkc: Just a suggestion, but for brevity in the response you could consider having an array rather than an object, and using the symbol for the op - for example instead of this
```json
{
  "@bob:some_thing.some_app@alice":{"commitId":101,"commitOp":"UPDATE"},
  "public:firstname.wavi@alice":{"commitId":105,"commitOp":"UPDATE_ALL"},
  "@bob:some_other_thing.some_app@alice":{"commitId":112,"commitOp":"DELETE"}
}
```
the response could instead look like
```json
{
  "@bob:some_thing.some_app@alice":[101,"+"],
  "public:firstname.wavi@alice":[105,"*"],
  "@bob:some_other_thing.some_app@alice":[112,"-"],
}
```

sitaram-kalluri: Sure Gary, Will make the changes.

gkc (approved):

PR #1368, opened 2023-05-23
fix: Filter LatestCommitEntryOfEachKey with regex in stats verb

sitaram-kalluri:
**- What I did**
* Uptake the at_commons version 3.0.47 which filters the LatestCommitEntryOfEachKey with regex
* In stats_verb_handler, fetch the regex and pass it to "getEntries" method to filter the keys

### DIFF
--- a/packages/at_secondary_server/lib/src/verb/handler/stats_verb_handler.dart
+++ b/packages/at_secondary_server/lib/src/verb/handler/stats_verb_handler.dart
@@ -113,7 +113,7 @@ class StatsVerbHandler extends AbstractVerbHandler {
     var metric = _getMetrics(id);
     var name = metric.name!.getName();
     dynamic value;
-    if (id == '3' && _regex != null) {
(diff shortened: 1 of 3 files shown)


gkc (approved):

packages/at_secondary_server/lib/src/verb/handler/stats_verb_handler.dart
```diff
@@ -113,7 +113,7 @@ class StatsVerbHandler extends AbstractVerbHandler {
     var metric = _getMetrics(id);
     var name = metric.name!.getName();
     dynamic value;
-    if (id == '3' && _regex != null) {
+    if ((id == '3' || id == '15') && _regex != null) {
```
murali-shris, line 116: can we replace this hard coding with the value from the map or add a map which has stats type to id mapping.

murali-shris (changes requested):

PR #1386, opened 2023-06-01
fix: root pubspec to use git ref for at_persistence_root_server

cpswan:
Recent build of the dev root image haven't had the root binary in them due to `../at_persistence_root_server` not resolving within Docker.

**- What I did**

* Reverted to using a git reference for the package
* `set -eux` so that any future failures will cause the build to fail in a noticeable way

**- How to verify it**

I have verified locally with `sudo docker build -t atsigncompany/root .`

fix: root pubspec to use git ref for at_persistence_root_server


gkc (approved):
Oops, sorry about that

PR #1392, opened 2023-06-05
feat: Retain current pool management logic, but be a **LOT** less aggressive when closing idle **authenticated** connections

gkc:
**- What I did**
feat: Retain current inbound pool management logic, but be a **LOT** less aggressive when closing idle **authenticated** inbound connections

**- How I did it**
- Added new atServer config for 'authenticatedInboundIdleTime', defaulting to 30 days. Why 30 days? Well, there's a bug in at_lookup (there's [a PR in at_libraries](https://github.com/atsign-foundation/at_libraries/pull/354) to fix that) which leads to unhandled exceptions when server destroys idle connections. This change will ensure that that problem no longer affects existing clients
- Left existing `inboundIdleTime` at its current default value of 10 minutes, but we are now using it as the configured value for **unauthenticated** connections only
- Set default limit on total number of inbound and outbound connections to 200
- Added better instance variable comments to AtSecondaryContext
- Modified existing inbound connection pool management test accordingly. Note that the test remains essentially unchanged. Added plenty of commentary to explain the test behaviour in more detail
- Sprinkled some more comments throughout

**- How to verify it**
Tests pass

feat: Retain current inbound pool management logic, but be a **LOT** less aggressive when closing idle **authenticated** inbound connections

### DIFF
--- a/packages/at_secondary_server/config/config.yaml
+++ b/packages/at_secondary_server/config/config.yaml
@@ -45,20 +45,31 @@ hive:
 
 # The @Protocol connection configurations.
 connection:
-  # The maximum time in milliseconds for an inbound connection to expire.
-  # This should NEVER be set to less than 60 seconds. Internally, the effective
-  # value used is reduced progressively as the number of connections approaches the max limit
+  # At any point, at most [inbound_max_limit] inbound connections are allowed. The inbound connection
+  # pool takes care of ensuring that idle connections are closed as necessary
+  inbound_max_limit: 200
+
+  # The maximum time in milliseconds for an **unauthenticated** inbound connection to expire.
+  # Default value is 10 minutes (10 * 60 * 1000 == 600000)
+  # Internally, the effective value used is reduced progressively towards 5 seconds as the current
+  # number of connections approaches the inbound_max_limit.
   inbound_idle_time_millis: 600000
+
+  # The maximum time in milliseconds for an **AUTHENTICATED** inbound connection to expire.
+  # Default value is 30 days (30 * 24 * 60 * 60 * 1000 == 2592000000)
+  # Internally, the effective value used is reduced progressively towards 60 seconds as the current
+  # number of connections approaches the max limit. So for example if the max number of connections is
+  # 200, and there are currently 200 connections, and a new client tries to connect,
+  # then any authenticated connection which has been idle for more than 60 seconds will be closed
+  # However: if there are only, say, 100 connections currently, then only authenticated connections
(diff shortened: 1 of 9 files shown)


cconstab (dismissed):
All looks good...

My only thought was to document how we roll back to normal while its fresh in your mind now. So when the time comes its easy, for some one else to do. 
Plus set up a remind ourselves to roll back in the calendar or something so things do get migrated back when the clients are all updated..

packages/at_secondary_server/config/config.yaml
```diff
@@ -45,20 +45,31 @@ hive:
 
 # The @Protocol connection configurations.
```
cpswan: Let's use the opportunity to s/@Protocol/atProtocol/

packages/at_secondary_server/lib/src/server/at_secondary_config.dart
```diff
@@ -61,9 +61,10 @@ class AtSecondaryConfig {
   static final int _runRefreshJobHour = 3;
 
   //Connection
-  static final int _inboundMaxLimit = 10;
-  static final int _outboundMaxLimit = 10;
-  static final int _inboundIdleTimeMillis = 600000;
+  static final int _inboundMaxLimit = 200;
```
cpswan: I know there's a bit of following suit here with the existing code, but could (should?) these be `const` rather than `static final`?

cpswan (dismissed):
LGTM, but maybe we can make it better whilst we're here in this code?

packages/at_secondary_server/lib/src/server/at_secondary_config.dart
gkc: > All looks good...
> 
> My only thought was to document how we roll back to normal while its fresh in your mind now. So when the time comes its easy, for some one else to do. Plus set up a remind ourselves to roll back in the calendar or something so things do get migrated back when the clients are all updated..

Sure thing. I don't see any real need to revert later but if we do then it's a change to config.yaml where I've documented these config parameters well.

I will create a ticket to review in a couple of months.

gkc: > I know there's a bit of following suit here with the existing code, but could (should?) these be `const` rather than `static final`?

👍 Will change

packages/at_secondary_server/config/config.yaml
gkc: 👍

cpswan (approved):

PR #1427, opened 2023-07-03
<DO NOT MERGE> fix: Point at_persistence_secondary_server to memory_optimization branch

sitaram-kalluri:
**- What I did** This is a draft PR to run tests with at_persistence_secondary_server pointing to "memory_optimization" branch.

PR #1428, opened 2023-07-03
fix: Replace Iterator with getAt to optimize memory and update unit t…

### LINKED ISSUE #1303: Reduce atServer memory requirements

50MB seems insufficent for many atSigns ... recently we've had to bump memory for `@snowthe18raw` and `@intenseaquarius` as they were constantly being killed by docker swarm manager very shortly after startup

atServer should have a predictable cap on amount of memory used during startup and in steady state

Let's start by profiling atServer memory usage during startup and in steady state to see what we can reduce.

_No response_

### Additional context

_No response_

sitaram-kalluri:
**- What I did** 
* Refactor the code 
    * To avoid using of Iterator to fetch all the keys from the key-store in `_initMetadataCache` and `getKeys` because all the keys are loaded into memory at once causing unnecessary consumption of memory. Instead, fetch each key from the key store individually and add its value to the _metaDataCache.
    * Store only keys which have TTL or  TTB set in the _metaDataCache map. This modification allows us to optimize memory usage by storing only the necessary keys.
    * Since the intent of maintaining the _metaDataCache map is to verify if the key is active or not, store only "expiryAt" and "availableAt" from "AtMetadata" in _metaDataCache Map rather storing the entire AtMetadata object.
    * Rename the `_metaDataCache` map has `_expiryKeysCache` map since we not caching the entire AtMetadata object and `_initMetadataCache` as `_initExpiryKeysCache`
    *  In `getMeta`, fetch the "AtMetadata" from key-store, because the AtMetadata is no longer stored in the _metaDataCache map.
    * Remove `getKeysFromKeyStore` because the method is no longer used.

**- How I did it**
* In the _initMetaDataCache method, iterate over each key in the key-store and fetch the value associated with that key using the getAt method. For each key, we add the key-value pair to the _expiryKeyCache map. This way, we fetch each key individually and add its value to the _expiryKeyCache map, avoiding the need to load all the keys into memory at once.
* In `getKeys` method also, instead of using Iterator use "getAt" to fetch all the keys from the keystore.
* In `getKeys` method, when optional argument `regex` is populated, the list is iterated twice.
   * In the first pass, the list is iterated to fetch the keys that match the regex.
   * The resultant of first pass is again iterated to check for if the key is active. The active keys are added to the final resultant list.
   * Modify the code such that, when `regex` is not populated, default it ".*" to match all the keys. In a single iteration, verify if the key is active and if key matches the regex, add it to the final resultant list.
* Introduce `_updateMetadataCache` to add the keys to `_expiryKeysCache` only when key has TTL or TTB set, else do not add the key to `_expiryKeysCache` map.
* Update and add new unit tests to assert the changes.

**- How to verify it**
* All the unit tests in at_persistence_secondary_server should pass.
* Raised a draft in at_secondary_server pointing the at_persistence_secondary_server to this branch. All the tests in E2E and functional tests passed. https://github.com/atsign-foundation/at_server/pull/1427

* Refactor the code in `hive_keystore.dart` to optimize the memory usage.

Following are the screenshots of memory chat before and after the changes:

1. Screenshot of the Memory chat before the changes (Code from trunk branch):

memory_with_trunk

2. Screenshot of the Memory chat after the changes:

with_memory_optimization_1


sitaram-kalluri:
Following link which contains the time and space complexity comparisons between the "trunk" branch and "memory_optimization" branch - [memory_optimization](https://docs.google.com/spreadsheets/d/14kGBE0ldXlukjFh0vsELc7bxHgZx7wHXc8jzaWMMhk0/edit#gid=1082010380)

purnimavenkatasubbu:
Verified with buzz and wavi apps and also with the client and server tests. Attaching the logs below.

[buzzlogsMemoryOptimization.txt](https://github.com/atsign-foundation/at_server/files/11978397/buzzlogsMemoryOptimization.txt)
[clientTestLogsForTTLandTTBKey.txt](https://github.com/atsign-foundation/at_server/files/11978415/clientTestLogsForTTLandTTBKey.txt)
[waviLogsMemoryOptimization.txt](https://github.com/atsign-foundation/at_server/files/11978422/waviLogsMemoryOptimization.txt)
[server_e2etests_logs_memory_optimization.txt](https://github.com/atsign-foundation/at_server/files/11978484/server_e2etests_logs_memory_optimization.txt)
[serverTTLTTBKeysLogs.txt](https://github.com/atsign-foundation/at_server/files/11978491/serverTTLTTBKeysLogs.txt)

gkc:
LGTM but please @sitaram-kalluri do one more read through the changes, with someone else, and double check that there is at least one test which exercises each function/method which has changed

sitaram-kalluri:
> LGTM but please @sitaram-kalluri do one more read through the changes, with someone else, and double check that there is at least one test which exercises each function/method which has changed

Sure, Gary.

murali-shris (approved):

PR #1443, opened 2023-07-08
feat: Apkam namespace access changes

murali-shris:
**- What I did**
- added namespace access controls to lookup, update, delete and scan verbs
**- How I did it**
- added isAuthorized method to abstract_verb_handler.dart which checks the access flags from enrollment data from keystore and return true/false
- added getEnrollmentNamespaces method to abstract_verb_handler.dart which return list of  namespaces for a given enrollmentId
- called isAuthorized method from verb handler and if the result is false throw UnAuthorizedException
- functional tests enroll_namespace_access_test.dart added by @purnimavenkatasubbu 
**- How to verify it**
- run the functional tests enroll_namespace_access_test.dart


gkc (approved):
Excellent work @murali-shris - nice simple permission check pattern and a good set of tests

PR #1465, opened 2023-07-19
feat: skip commit log param in keystore

murali-shris:
**- What I did**
- added implementation changes for skip commit log param in keystore
**- How I did it**
- in hive_keystore.dart , if skipCommit=true for create/put/remove return -1 and don't add entry to commit log
- added unit tests in hive_keystore_impl_test.dart
**- How to verify it**
- run the test group - A group of tests to verify skip commit in hive_keystore_impl_test.dart

### DIFF
--- a/packages/at_persistence_secondary_server/CHANGELOG.md
+++ b/packages/at_persistence_secondary_server/CHANGELOG.md
@@ -1,7 +1,7 @@
-## 3.0.56
-- fix: Apply Utf7.decode function to decode the keys and atSigns containing emojis.
 ## 3.0.55
 - fix: Refactor Hive keystore to optimize memory usage
+- fix: Apply Utf7.decode function to decode the keys and atSigns containing emojis.
+- feat: add skipCommit flag to keystore implementation which enables skipping commit log for put/create/remove.
 ## 3.0.54
 - fix: Add NotificationType.Self in read and write methods of at_notification.dart
 ## 3.0.53
--- a/packages/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
+++ b/packages/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
@@ -105,7 +105,8 @@ class HiveKeystore implements SecondaryKeyStore<String, AtData?, AtMetaData?> {
       String? encAlgo,
       String? ivNonce,
       String? skeEncKeyName,
-      String? skeEncAlgo}) async {
+      String? skeEncAlgo,
+      bool skipCommit = false}) async {
     key = key.toLowerCase();
     final atKey = AtKey.getKeyType(key, enforceNameSpace: false);
     if (atKey == KeyType.invalidKey) {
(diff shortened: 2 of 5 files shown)


gkc (commented):
LGTM

Will re-review when persistence spec has been published and dependency override has been removed

murali-shris:
> LGTM
> 
> Will re-review when persistence spec has been published and dependency override has been removed

@gkc 
In the end2endtest staging which is failing, current version of persistence secondary in pub.dev(doesn't have optional param in impl class) and persistence spec(has the optional param in interface) are retrieved which is causing the failure. How can we resolve this failure?

gkc:
> > LGTM
> > Will re-review when persistence spec has been published and dependency override has been removed
> 
> @gkc In the end2endtest staging which is failing, current version of persistence secondary in pub.dev(doesn't have optional param in impl class) and persistence spec(has the optional param in interface) are retrieved which is causing the failure. How can we resolve this failure?

- The problem is in the `Activating atSign` step, when doing dart pub get for the at_onboarding_cli package
- This problem will happen everywhere when anyone tries to do pub get of pretty much anything
- Quickest way forward is to approve this PR, merge it, and publish
- In order to prevent this in the future, maybe we can update the pubspec of at_persistence_secondary_server so that it depends on a specific version of at_persistence_spec
  - i.e. instead of `^2.0.12`
  - we should have `'>=2.0.11 <=2.0.12`

gkc:
> > > LGTM
> > > Will re-review when persistence spec has been published and dependency override has been removed
> > 
> > 
> > @gkc In the end2endtest staging which is failing, current version of persistence secondary in pub.dev(doesn't have optional param in impl class) and persistence spec(has the optional param in interface) are retrieved which is causing the failure. How can we resolve this failure?
> 
> * The problem is in the `Activating atSign` step, when doing dart pub get for the at_onboarding_cli package
> * This problem will happen everywhere when anyone tries to do pub get of pretty much anything
> * Quickest way forward is to approve this PR, merge it, and publish
> * In order to prevent this in the future, maybe we can update the pubspec of at_persistence_secondary_server so that it depends on a specific version of at_persistence_spec
>   
>   * i.e. instead of `^2.0.12`
>   * we should have `'>=2.0.11 <=2.0.12`

I think we can trick the pub check with this: `  at_persistence_spec: '>2.0.12 <=2.0.13'`

gkc:
The `end2end_test_staging` check will fail again but once we've published the new at_persistence_secondary_server and keep this pattern going forward, this problem should not recur.

gkc:
Once we've merged to trunk and published, please do another dummy PR to
1. verify that the end2end_test_staging job will succeed
2. ensure that the jobs to build the new staging docker images run (as they are dependent on end2end_test_staging job)

gkc (approved):

gkc:
Approved and merged. @murali-shris please pull from trunk and publish new version

gkc:
@murali-shris in the interests of time, I will publish new version now, and then do the dummy PR to follow up

gkc:
OK this is messy. I'm going to retract the new at_persistence_spec for now, because version 3.0.55 of at_persistence_secondary_server already exists (Sita published last week, but we then had to retract it because of a bug)

gkc:
Have now published new versions of at_persistence_spec and at_persistence_secondary_server

PR #1471, opened 2023-07-24
fix: prevent deletion of protected keys

### LINKED ISSUE #1325: Enhance DeleteVerbHandler so it prevents deletion of important data
### Describe the bug

It is currently possible for an authenticate client to delete data such as the server's signing keypair

### Steps to reproduce
See https://github.com/atsign-foundation/operations/issues/14

### Expected behavior

DeleteVerbHandler should prevent deletion of data which is important to correct functioning. In addition to the signing keypair mentioned above, we should consider whether other data should be deletable (e.g. `public:publickey@<atSign>`)

srieteja:
closes https://github.com/atsign-foundation/at_server/issues/1325

**- What I did**
- Improve the Delete verb to prevent the deletion of protected keys

**- How I did it**
- feat: introduce a new field in config.yaml for protectedKeys, for users to add protected keys
- feat: introduce new class variable 'protectedKeys' in AtSecondaryConfig that consists of ['signing_publickey', 'signing_privatekey', 'publickey', 'at_pkam_publickey']
- feat: introduce methods in DeleteVerbHandler that check if the key to be deleted is present in the list of protected keys, and returns an error message when true.
Note: The list of protected keys in config.yaml overrides the list of protectedKeys in AtSecondaryConfig

**- How to verify it**
- unit and functional tests added

fix: prevent deletion of protected keys


packages/at_secondary_server/lib/src/verb/handler/delete_verb_handler.dart
```diff
@@ -140,4 +149,23 @@ class DeleteVerbHandler extends ChangeVerbHandler {
...
+    for (var key in AtSecondaryConfig.protectedKeys) {
+      // convert generic key name to actual public key
+      // of format: key:@atsign
+      protectedKeys.add('$key$atsign');
+    }
+    return protectedKeys;
+  }
+
+  bool _isProtectedKey(String key) {
+    if (protectedKeys!.contains(key)) {
```
murali-shris, line 164: instead of contains..match the keys with the exact pattern
```
* public:signing_publickey<atsign>
* <atsign>:signing_privatekey<atsign>
* publickey<atsign>
* privatekey:at_pkam_publickey
```

murali-shris (changes requested):

packages/at_secondary_server/lib/src/verb/handler/delete_verb_handler.dart
srieteja: since this is a list, contains would match the entire key with protectedKeys

packages/at_secondary_server/lib/src/server/at_secondary_config.dart
```diff
@@ -675,6 +688,19 @@ class AtSecondaryConfig {
...
   }
 
+  static List<String> get protectedKeys {
+    try {
+      YamlList keys = getConfigFromYaml(['hive', 'protectedKeys']);
+      List<String> protectedKeys = [];
+      for (var key in keys) {
+        protectedKeys.add(key);
+      }
+      return protectedKeys.isEmpty ? _protectedKeys : protectedKeys;
```
murali-shris: change this to
 protectedKeysFromConfig.isNotEmpty ? protectedKeysFromConfig : _defaultProtectedKeys;

murali-shris (dismissed):
gave a review comment

packages/at_secondary_server/lib/src/server/at_secondary_config.dart
srieteja: fixed as part of [567351b](https://github.com/atsign-foundation/at_server/pull/1471/commits/567351b6fe6c7c9af33b2644d579f4a616e8d71e)

packages/at_secondary_server/lib/src/server/at_secondary_config.dart
```diff
@@ -675,6 +688,21 @@ class AtSecondaryConfig {
...
+        protectedKeysFromConfig.add(key);
+      }
+      return protectedKeysFromConfig.isNotEmpty
+          ? protectedKeysFromConfig
+          : _protectedKeys;
+    } on Exception {
+      return _protectedKeys;
+    }
+  }
+
```
gkc, line 709: 1. The YAML config, if any, should _add_ to the default list of protected keys, not override it. The four keys named in `_protectedKeys` should _always_ be protected from deletion, regardless of what is in the YAML

2. Please use Sets instead of Lists. i.e. change the `_protectedKeys` to being a Set<String>, and change the return value of `get protectedKeys` to be a Set<String> and make the appropriate changes that flow from that

packages/at_secondary_server/config/config.yaml
```diff
@@ -42,6 +40,10 @@ hive:
   # This is set to true by default. To retain the malformed keys on server start-up,
   # set the flag to false.
   shouldRemoveMalformedKeys: true
+  # List of keys are crucial for the server to function as expected
+  # The following keys cannot be deleted
+  # The list provided here will override the list of server's protected keys
```
gkc: Should not override the default list; should augment it

packages/at_secondary_server/config/config.yaml
```diff
@@ -42,6 +40,10 @@ hive:
...
+  protectedKeys: []
```
gkc, line 45: If this were committed then, as currently implemented in AtSecondaryConfig, this empty list would override the default list of protected keys and thus they would be deletable

packages/at_secondary_server/lib/src/server/at_secondary_config.dart
```diff
@@ -675,6 +688,21 @@ class AtSecondaryConfig {
     }
   }
 
+  static List<String> get protectedKeys {
+    try {
+      YamlList keys = getConfigFromYaml(['hive', 'protectedKeys']);
```
gkc, line 698: I'd like to see unit tests with various flavours of yaml config to verify the logic

One way of enabling those tests to be written would be to augment AtSecondaryConfig to have a way to inject some yaml and change its `getConfigFromYaml` to use that yaml if it is set

gkc (changes requested):

packages/at_secondary_server/config/config.yaml
gkc: Please also add to the comment in the yaml file to explain the `<@atSign>` formatting

packages/at_secondary_server/lib/src/server/at_secondary_config.dart
srieteja: Change made as part of [d2b6e9d](https://github.com/atsign-foundation/at_server/pull/1471/commits/d2b6e9dd55f062903c77f60dde72db11c9b63d95)

packages/at_secondary_server/config/config.yaml
srieteja: change has been made as part of [99483ce](https://github.com/atsign-foundation/at_server/pull/1471/commits/99483cee49c36e172a76bcbd9f86841d976c5fbc)

packages/at_secondary_server/lib/src/server/at_secondary_config.dart
srieteja: Changes to allow config injection into AtSecondaryConfig and the test itself are in [8ee7c62](https://github.com/atsign-foundation/at_server/pull/1471/commits/8ee7c6296c3707f9eeb1049140f66fa17e16d5dd). And the test config files and config util are part of [77c7be0](https://github.com/atsign-foundation/at_server/pull/1471/commits/77c7be013c407951bfe78a46650476131561b20d). What do you think about this approach @gkc

gkc: Very good, thank you @srieteja

packages/at_secondary_server/test/delete_verb_test.dart
```diff
@@ -96,4 +104,50 @@ void main() {
...
+    test('verify deletion of encryption public key', () {
+      var command = 'delete:$AT_ENCRYPTION_PUBLIC_KEY@alice';
+      var paramsMap = getVerbParam(Delete().syntax(), command);
+      expect(
+          () =>
+              handler.processVerb(Response(), paramsMap, mockInboundConnection),
+          throwsA(
+              predicate((exception) => exception is UnAuthorizedException)));
+    });
+
```
gkc, line 143: You need to add one more test asserting that the pkam public key may not be deleted

packages/at_secondary_server/test/delete_verb_test.dart
```diff
@@ -96,4 +104,50 @@ void main() {
...
+    test(
+        'Verify protectedKeys from configYaml being appended to the list of protectedKeys in AtSecondaryConfig',
+        () {
+          TestConfigUtil.setTestConfig(1);
+          expect(AtSecondaryConfig.protectedKeys.length, 7);
+          TestConfigUtil.setTestConfig(2);
+          expect(AtSecondaryConfig.protectedKeys.length, 6);
+          TestConfigUtil.resetTestConfig();
+        });
+  });
```
gkc, line 191: 1. Please add a test_config_3.yaml which has `protectedKeys: []` and an expect for the resultant `AtSecondaryConfig.protectedKeys.length`
2. For each testConfig 1, 2 and 3, add an expect that `AtSecondaryConfig.protectedKeys` contains the core set of four that should always be present

tests/at_functional_test/test/delete_verb_test.dart
```diff
@@ -152,6 +153,35 @@ void main() {
...
+  test('delete verb for an protected key - pkam_publickey', () async {
+    ///DELETE VERB
+    await socket_writer(
+        socketFirstAtsign!, 'delete:privatekey:at_pkam_publickey');
+    var response = await read();
+    print('delete verb response : $response');
+    // the error is an expected behaviour
+    assert((response.contains('Invalid syntax')));
+  });
+
```
gkc, line 220: Please add explicit tests for the other two core protected keys (public encryption key, signing private key)

gkc (changes requested):

packages/at_secondary_server/test/delete_verb_test.dart
srieteja: This has been added as part of 73086dfc94ab73767bd16b3968ade74f3da04b0c

packages/at_secondary_server/test/delete_verb_test.dart
srieteja: This has been added as part of 73086dfc94ab73767bd16b3968ade74f3da04b0c

tests/at_functional_test/test/delete_verb_test.dart
srieteja: These tests have been added through 6fcdede737deb85ee00f32961f10fffa2d0c9b1a

gkc:
@srieteja is this ready for review again?

srieteja:
> @srieteja is this ready for review again?

Yes. All the requested changes have been made.

gkc (approved):
LGTM! Thanks @srieteja

PR #1480, opened 2023-07-27
build: create dir config inside build_virtual_env/ve

srieteja:
**- What I did**
- Copy the config.yaml belonging to the secondary server (in this branch) into the secondary_server directory that will be used by the functionall tests in github actions

**- How I did it**
- add commands in the github workflow file to do the following:
1.  Create a new directory 'tools/build_virtual_environment/ve/contents/atsign/secondary/config/'
2.  Copy the config.yaml belonging to the secondary server from this branch into the newly created directory above


cpswan:
@srieteja I know this is still draft, but can you please try to explain (in the root comment) what you're trying to accomplish with this?

srieteja:
@cpswan  I have added a new entry into the seconday_server's config.yaml (in https://github.com/atsign-foundation/at_server/pull/1471), which is not being picked up in the functional tests. I'm trying to see if there's a way I can copy the new config.yaml into the build_virtual_environment's secondary. Do you think it's a good idea ?

srieteja:
Closing this PR as we have decided, this approach may not be the best option.

Issue #1474, opened 2023-07-26
end2endtest_staging job is failing every time; atServers are not being created


gkc:
### Describe the bug

- new atSigns are being registered successfully, but something is subsequently failing - for example [here](https://github.com/atsign-foundation/at_server/actions/runs/5667036592/job/15355436609) two new atSigns were created:
```
AT_SIGN_1: test7cd18781_5943_4a8d_a6f1_8c66a9f2031e1
AT_SIGN_2: test3f01bc5d_14b3_4c76_b1ab_e576553497bb1
```
- a manual lookup of those atSigns on `root.atsign.wtf:64` shows that the atServers were indeed assigned a hostname
```
@test7cd18781_5943_4a8d_a6f1_8c66a9f2031e1
e91ab171-82e6-5125-9186-ff1cbff78ee9.staging0001.atsign.zone:1472
@test3f01bc5d_14b3_4c76_b1ab_e576553497bb1
33608b39-d296-5163-b359-347552f94df7.staging0001.atsign.zone:1473
```
- **however** neither of those addresses is reachable
```
gary@gkc2019-2 ~ % openssl s_client -connect e91ab171-82e6-5125-9186-ff1cbff78ee9.staging0001.atsign.zone:1472
getaddrinfo: nodename nor servname provided, or not known
connect:errno=0
gary@gkc2019-2 ~ % openssl s_client -connect 33608b39-d296-5163-b359-347552f94df7.staging0001.atsign.zone:1473
getaddrinfo: nodename nor servname provided, or not known
connect:errno=0
gary@gkc2019-2 ~ % 
```

### Steps to reproduce

Multiple actions runs have failed, a link is provided above

### Expected behavior

Should not fail.

cpswan:
@gkc Looking at a [recent run](https://github.com/atsign-foundation/at_server/actions/runs/5678587078/job/15389378282) I see `Error fetching Atsign Hostname`

If I try to get the name myself then it's in the root:

```
(echo test815d3141_4719_4333_9cd5_efb321df9e571; sleep 1) | openssl s_client -brief -connect root.atsign.wtf:64 2> /dev/null | grep --color=none "^@.*:" | cut -d'@' -f2
2a34e09e-5f04-54d2-b1da-ba35ac0def96.staging0001.atsign.zone:1489
```

NB I'm using `-brief` there, which isn't in the prod script (and should be)
NB2 The prod script should have retry logic rather than just failing on first try.

[...]

PR #1481, opened 2023-07-27
fix: Retry in staging e2e test steps that frequently fail

cpswan:
Trying to stop repeats of #1474

**- What I did**

* Removed concurrency check (as that's only needed for CICD VMs)
* Added names to steps (and removed some comments that were the names or dupes of names)
* Move `env:` block to start of steps so it's obvious where things are coming from
* Added `-s` to curl operations so that logs aren't filled up with HTTP transfers
* Added `-brief` to openssl operations so that we're not getting so much connection delay
* `for` loop around many operations to attempt 5 times
* Moved `sleep` to the end of blocks so that we're at least trying to go quicker
* Added TODOs for some further refinements
* Made `Delete atSigns` step run always so that even if other steps fail, we still clear up.
* Refactored delete step into a function so it's a little more DRY
* s/Atsign/atSign/

**- How to verify it**

Logs from GitHub Actions runs.

fix: Retry in staging e2e test steps that frequently fail

### DIFF
--- a/.github/workflows/at_server.yaml
+++ b/.github/workflows/at_server.yaml
@@ -451,8 +451,8 @@ jobs:
           AT_SIGN_2: ${{ steps.atsign_names.outputs.AT_SIGN_2 }}
         run: |  
           for try in {1..5}; do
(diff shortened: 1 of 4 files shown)


kumarnarendra701 (dismissed):

cpswan:
This is now working pretty much as intended, but I'd like to make Check Connection more robust, which I'll take on in the morning.

cpswan:
@gkc @sitaram-kalluri @kumarnarendra701 this is ready for review now

gkc (approved):
Verrry nice

PR #1489, opened 2023-07-31
feat: Remove previous commit entry on update/delete of key

sitaram-kalluri:
**- What I did**
- When an existing key is updated/deleted, remove the previous commit entry from the commit log keystore.
- This helps to keep the size of the commit log thin and also the compaction job can subsequently be deprecated.

**- How I did it**
- Add a condition in `add` in `commit_log_keystore.dart` when updating the commit entry, remove the previous entry from the commit log keystore.

**- How to verify it**
- Added unit tests to assert the changes

- Remove previous commit entry on update/delete of a key


packages/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
```diff
@@ -68,6 +68,15 @@ class CommitLogKeyStore
...
         // update entry with commitId
         await _getBox().put(internalKey, commitEntry);
+        // Delete old commit entry for the same key from the commit log
+        // For now, just delete the previous entry from _commitLogCacheMap
+        // Eventually, after the compaction kicks in all of the old entries would be deleted.
+        // Subsequently, we can then disable the compaction job.
+        // When commitLogCacheMap.keys.length == getBox().length then we have fully compacted it
+        if (_commitLogCacheMap[commitEntry.atKey!] != null) {
+          await _getBox()
+              .delete(_commitLogCacheMap[commitEntry.atKey!]?.commitId);
```
gkc, line 79: For completeness, please add `&& _commitLogCacheMap[commitEntry.atKey!]?.commitId != null` to the if condition

gkc (changes requested):
one minor change please, otherwise LGTM

packages/at_persistence_secondary_server/lib/src/log/commitlog/commit_log_keystore.dart
sitaram-kalluri: Updated the IF condition as above.

gkc (approved):

Issue #1444, opened 2023-07-09
APKAM - access changes for monitor verb


murali-shris:
### Is your feature request related to a problem? Please describe.

When user is authenticated via APKAM, allow only the keys with authorized namespaces to flow through the verbs.

- Monitor verb: Allow only the notification where keys with authorized namespaces to stream through the monitor connection.
- Notify verb: Allow only the notification where keys with authorized namespaces to be processed through the notify verb.

### Describe the solution you'd like

- Refer to the implementation https://github.com/atsign-foundation/at_server/pull/1443
- In monitor verb handler, add a check to verify if the namespaces of a key are authorized with enrollmentId used to authenticate.
- Similar to monitor verb, add a check in notify verb check for authorization of namespace.

murali-shris:
Not started in PR66.carrying over

sitaram-kalluri:
In PR-67, spent 5 SP, implemented the changes.
The changes are available in [enroll_access_changes](https://github.com/atsign-foundation/at_server/tree/enroll_access_changes).

sitaram-kalluri:
No progress made in PR-68. Waiting on the PR: https://github.com/atsign-foundation/at_server/pull/1496 to merge.
Pending work is pull the latest changes of APKAM and resolve merge conflicts

sitaram-kalluri:
@gkc : Regarding the APKAM changes in "notify" verb, when recipient lacks access to a specific namespace, should a notification be sent or return error message? Additionally, if the recipient has subscribed to a namespace but lacks the authorization to write, should the cached key be stored or ignored?

gkc:
I'd like you to think this through and come to a conclusion, and then we can discuss your conclusion

sitaram-kalluri:
Following is the PR related the APKAM changes in monitor verb: https://github.com/atsign-foundation/at_server/pull/1525

Attaching the sequence diagram of APKAM functionality with respect to monitor verb:

```mermaid
sequenceDiagram
    participant w as Wavi-Approving App
    participant s as Server
    participant b as Buzz-Enrolling App

    note right of w: First App Enrollment
    w->>s: CRAM Authenticated
    s->>w: Auth Successful
    note right of w: __manage, * and wavi are added to enroll namespace list
    w->>s: Start monitor:selfNotifications
    s->>w: Stream all notifications via the monitor
    
    note left of b: Second app requesting enrollment
    b->>s: Request for enrollment - {buzz:rw}
[...]

PR #1496, opened 2023-08-05
feat: apkam spec changes and keys verb impl

sitaram-kalluri:
**- What I did**
- Implemented keys verb , changes to apkam spec, refactoring
**- How I did it**
- Renamed enrollApprovalId to enrollmentId for consistency
- Removed EnrollNamespace class and replaced with Map<String,String> (key- namespace, value-access) for ease of retrieval while checking authorisation for enrollments
- Added AtEnrollmentException to GlobalExceptionHandler
- In AbstractVerbHandler, added restriction for __global and __manage namespace. only keys verb will have access to these namespaces
- Changes to EnrollVerbHandler  
    - store encrypted default encryption private and encrypted default self encryption
    - For first client approval, store apkam public key in existing key name -  AT_PKAM_PUBLIC_KEY. For other approvals store apkam public key in __pkams__.public_keys format.
- For approval notification, stored encrypted apkam symmetric key in notification value.
- Implemented KeysVerbHandler with put, get and delete params
- Made changes in scan verb handler to not display __manage keys
- added unit and functional tests for keys verb, namespace access and enroll verb

**- How to verify it**
- run the unit tests enroll_data_store_value_test.dart, keys_verb_test.dart
- run the functional tests enroll_namespace_access_test.dart, enroll_verb_test.dart, keys_verb_test.dart


packages/at_secondary_server/lib/src/verb/handler/abstract_verb_handler.dart
```diff
@@ -80,7 +80,7 @@ abstract class AbstractVerbHandler implements VerbHandler {
   Future<void> processVerb(Response response,
       HashMap<String, String?> verbParams, InboundConnection atConnection);
 
-  Future<List<EnrollNamespace>> getEnrollmentNamespaces(
+  Future<Map<String, String>> getEnrollmentNamespaces(
```
gkc: Why have this method? It's equivalent to `(await getEnrollDataStoreValue).namespaces`
Unless there is a specific situation where (1) we want to just get the namespaces and (2) the record might not exist in the datastore?

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -26,6 +26,8 @@ class EnrollVerbHandler extends AbstractVerbHandler {
   @override
   Verb getVerb() => enrollVerb;
 
+  final _uuid = Uuid();
```
gkc: I don't think this should be an instance variable.

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -72,68 +78,75 @@ class EnrollVerbHandler extends AbstractVerbHandler {
...
-        .split(';')
-        .map((namespace) =>
-            EnrollNamespace(namespace.split(',')[0], namespace.split(',')[1]))
-        .toList();
-    logger.finer('enrollNamespaces: $enrollNamespaces');
-
-    var enrollmentId = Uuid().v4();
-    var key = '$enrollmentId.$newEnrollmentKeyPattern.$enrollManageNamespace';
+    var enrollNamespaces = enrollParams.namespaces!;
+    var newEnrollmentId = _uuid.v4();
```
gkc: Because you've made _uuid an instance variable, each enrollment would get the same enrollment id. Please revert to using Uuid().v4()

gkc: And please add a unit test to verify that consecutive enrollments get unique ids

packages/at_secondary_server/lib/src/verb/handler/abstract_verb_handler.dart
```diff
@@ -114,6 +114,12 @@ abstract class AbstractVerbHandler implements VerbHandler {
   Future<bool> isAuthorized(
```
gkc: Please add method documentation

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -72,68 +78,75 @@ class EnrollVerbHandler extends AbstractVerbHandler {
   }
 
   Future<void> _handleEnrollmentRequest(
```
gkc, line 97: Please add method comment

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -72,68 +78,75 @@ class EnrollVerbHandler extends AbstractVerbHandler {
...
     enrollmentValue.namespaces = enrollNamespaces;
     enrollmentValue.requestType = EnrollRequestType.newEnrollment;
     AtData enrollData = AtData()..data = jsonEncode(enrollmentValue.toJson());
     logger.finer('enrollData: $enrollData');
-
-    await keyStore.put('$key$currentAtSign', enrollData);
+    await keyStore.put('$key$currentAtSign', enrollData, skipCommit: true);
   }
 
   Future<void> _handleEnrollmentPermissions(
```
gkc, line 157: Please add method comment

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -143,13 +156,39 @@ class EnrollVerbHandler extends AbstractVerbHandler {
...
+        await keyStore.put(apkamPublicKeyInKeyStore, atData);
+        await _storeEncryptionKeys(
+            enrollmentIdFromParams!, enrollParams, currentAtSign);
+      }
     }
-    responseJson['enrollmentId'] = enrollmentId;
+    responseJson['enrollmentId'] = enrollmentIdFromParams;
+  }
+
+  Future<void> _storeEncryptionKeys(
```
gkc, line 204: Please add method comment

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -143,13 +156,39 @@ class EnrollVerbHandler extends AbstractVerbHandler {
...
+        skipCommit: true);
+    var selfKeyJson = {};
+    selfKeyJson['value'] = enrollParams.encryptedDefaultSelfEncryptionKey;
+    await keyStore.put(
+        '$newEnrollmentId.$defaultSelfEncryptionKey.$enrollManageNamespace$atSign',
+        AtData()..data = jsonEncode(selfKeyJson),
+        skipCommit: true);
   }
 
   EnrollStatus _getEnrollStatusEnum(String? enrollmentOperation) {
```
gkc, line 220: Given that the comparisons below are to lowercase, please coerce the enrollmentOperation to lower case if it is non-null, before entering the switch

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -217,33 +256,34 @@ class EnrollVerbHandler extends AbstractVerbHandler {
...
-      if (enrollmentNamespace.name == enrollManageNamespace) {
-        return true;
-      }
-    }
-    return false;
+    return enrollDataStoreValue.namespaces.containsKey(enrollManageNamespace);
   }
 
-  Future<void> _storeNotification(String notificationKey, String atSign) async {
+  Future<void> _storeNotification(
```
gkc, line 287: Please add method comment explaining what this is doing

packages/at_secondary_server/lib/src/verb/handler/keys_verb_handler.dart
```diff
@@ -0,0 +1,195 @@
...
+  // Method to return Instance of verb belongs to this VerbHandler
+  @override
+  Verb getVerb() {
+    return keys;
+  }
+
+  // Method which will process update Verb
+  // This will process given verb and write response to response object
+  // Input : Response, verbParams, AtConnection
+  @override
```
gkc, line 27: The method comment above is useless, as it gives no more information than `AbstractVerbHandler.processVerb`'s comment. Please either remove it or replace it with something more useful

packages/at_secondary_server/lib/src/verb/handler/keys_verb_handler.dart
```diff
@@ -0,0 +1,195 @@
...
+  Future<void> processVerb(
```
gkc, line 28: This method is very very long and difficult to follow. Please refactor

packages/at_secondary_server/lib/src/verb/handler/keys_verb_handler.dart
```diff
@@ -0,0 +1,195 @@
...
+        logger.finer('keyNameFromParams: $keyNameFromParams');
+        response.data =
+            (await keyStore.remove(keyNameFromParams, skipCommit: true))
+                .toString();
+        break;
+    }
+  }
+
+  // Function to add a key to filteredKeys if enrollmentId matches
+  Future<void> _addKeyIfEnrollmentIdMatches(List<dynamic> filteredKeys,
```
gkc, line 184: The method comment is almost literally just repeating the name of the method. Can you make it more helpful?

Also - general point - please use dart docs syntax - class / method / function / variable comments use `///` not `//`.

packages/at_secondary_server/lib/src/verb/handler/keys_verb_handler.dart
```diff
@@ -0,0 +1,195 @@
...
+    final value = await keyStore.get(key);
+    if (value != null && value.data != null) {
+      final valueJson = jsonDecode(value.data);
+      if (valueJson[enrollmentId] == enrollIdFromMetadata) {
+        filteredKeys.add(key);
+      }
+    }
+  }
+
+  String? _getKeyName(HashMap<String, String?> verbParams, String atSign,
```
gkc, line 200: Please add method comment explaining why this method is needed and what it does (NOT how it does it, the code tells us that)

packages/at_secondary_server/test/keys_verb_test.dart
```diff
@@ -0,0 +1,800 @@
...
+        'namespaces': {'name': 'wavi', 'access': 'rw'},
+        'apkamPublicKey': 'testPublicKeyValue',
+        'requestType': 'newEnrollment',
+        'approval': {'state': 'approved'}
+      };
+      var keyName = '$enrollId.new.enrollments.__manage@alice';
+      await secondaryKeyStore.put(
+          keyName, AtData()..data = jsonEncode(enrollJson));
+
+      var encryptedDefaultEncryptionPrivateKey =
```
gkc: Why the hard-coded string?

gkc (changes requested):

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
gkc: Also would be really good to not use a switch statement but instead to use the enum's `values.byName()` method.

The equivalent of this method in one line of code would be something like
`EnrollStatus.values.byName((enrollmentOperation ?? 'pending').toLowerCase());`

gkc (changes requested):

packages/at_secondary_server/test/keys_verb_test.dart
murali-shris: Instead of a dummy value, wanted to test  with actual encrypted value. Server doesn't have dependency for at_demos currently. Is it fine if I add at_demos dependency, get the private key and encrypt with apkam symmetric key?

gkc: But this is a unit test, nothing to do with at_demos

You can cut keypairs and keys as you need to in your unit tests

packages/at_secondary_server/lib/src/verb/handler/abstract_verb_handler.dart
murali-shris: Removed this method and fetched namespaces from getEnrollDataStoreValue(...) result
https://github.com/atsign-foundation/at_server/pull/1496/commits/b22ea2028164aa54299758080d3b9033bf3a98e2

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
murali-shris: done.
https://github.com/atsign-foundation/at_server/pull/1496/commits/b22ea2028164aa54299758080d3b9033bf3a98e2

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
murali-shris: done
https://github.com/atsign-foundation/at_server/pull/1496/commits/b22ea2028164aa54299758080d3b9033bf3a98e2

packages/at_secondary_server/lib/src/verb/handler/abstract_verb_handler.dart
murali-shris: done
https://github.com/atsign-foundation/at_server/pull/1496/commits/ab0f642e5d3288d109bfbbbb912c6bebb3cad4ff

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
murali-shris: done
https://github.com/atsign-foundation/at_server/pull/1496/commits/ab0f642e5d3288d109bfbbbb912c6bebb3cad4ff

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
murali-shris: Done
https://github.com/atsign-foundation/at_server/pull/1496/commits/b22ea2028164aa54299758080d3b9033bf3a98e2

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
murali-shris: Done
https://github.com/atsign-foundation/at_server/pull/1496/commits/b22ea2028164aa54299758080d3b9033bf3a98e2

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
murali-shris: Enum values and the operation name are not identical.  Hence byName will not work in this case.
e.g approved - approve, denied- deny, revoked-revoke.
I have removed switch case and simplified the code
https://github.com/atsign-foundation/at_server/pull/1496/commits/b22ea2028164aa54299758080d3b9033bf3a98e2

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
murali-shris: done.
https://github.com/atsign-foundation/at_server/pull/1496/commits/b22ea2028164aa54299758080d3b9033bf3a98e2

packages/at_secondary_server/lib/src/verb/handler/keys_verb_handler.dart
murali-shris: Removed.
https://github.com/atsign-foundation/at_server/pull/1496/commits/b22ea2028164aa54299758080d3b9033bf3a98e2

packages/at_secondary_server/lib/src/verb/handler/keys_verb_handler.dart
murali-shris: Done.
https://github.com/atsign-foundation/at_server/pull/1496/commits/b22ea2028164aa54299758080d3b9033bf3a98e2

packages/at_secondary_server/lib/src/verb/handler/keys_verb_handler.dart
murali-shris: Done.
https://github.com/atsign-foundation/at_server/pull/1496/commits/b22ea2028164aa54299758080d3b9033bf3a98e2

packages/at_secondary_server/lib/src/verb/handler/keys_verb_handler.dart
murali-shris: done
https://github.com/atsign-foundation/at_server/pull/1496/commits/b22ea2028164aa54299758080d3b9033bf3a98e2

packages/at_secondary_server/test/keys_verb_test.dart
murali-shris: Done
https://github.com/atsign-foundation/at_server/pull/1496/commits/116fa391e6668f9b4534864e8691d523ce47fb6e

packages/at_secondary_server/lib/src/verb/handler/keys_verb_handler.dart
```diff
@@ -0,0 +1,224 @@
...
+    logger.finer('enrollIdFromMetadata:$enrollIdFromMetadata');
+    final key =
+        '$enrollIdFromMetadata.$newEnrollmentKeyPattern.$enrollManageNamespace';
+
+    var enrollData = await _getEnrollData(key, atSign);
+    if (enrollData != null) {
+      final enrollDataStoreValue =
+          EnrollDataStoreValue.fromJson(jsonDecode(enrollData.data!));
+      if (enrollDataStoreValue.approval?.state != 'approved') {
+        throw AtEnrollmentException(
```
gkc, line 52: I don't see a unit test for this case ... if there isn't one, please can you add one?

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -71,69 +75,99 @@ class EnrollVerbHandler extends AbstractVerbHandler {
...
+        enrollParams.deviceName!,
+        enrollParams.apkamPublicKey!);
 
-    if (atConnection.getMetaData().isAuthenticated) {
-      // approve request from connection that are authenticated. This connection may be cram or legacyPkam authenticated
-      enrollNamespaces.add(EnrollNamespace(enrollManageNamespace, 'rw'));
+    if (atConnection.getMetaData().authType != null &&
+        atConnection.getMetaData().authType == AuthType.cram) {
+      // auto approve request from connection that is CRAM authenticated.
+      enrollNamespaces[enrollManageNamespace] = 'rw';
```
gkc, line 126: Should we add another entry into `enrollNamespaces` like `"*":"rw"` ? How are we currently determining access permissions where the only entry in `enrollNamespaces` is `"$enrollManageNamespace":"rw"` ?

gkc (commented):
Nearly there! Last couple of questions

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
murali-shris: we didn't handle this case. if the only access is __manage and keynamespace is .wavi,.buzz etc., then we are returning isAuthorized = false..We can add an entry *:rw for the privileged client to access all namespaces

gkc: Has this been done? And is the `*:rw` namespace permission being checked and understood in the relevant places?

packages/at_secondary_server/lib/src/verb/handler/keys_verb_handler.dart
gkc: Has this been done?

gkc:
Nearly there! Please resolve the conflicts and confirm the answers to my questions in the unresolved comments above

packages/at_secondary_server/lib/src/verb/handler/keys_verb_handler.dart
murali-shris: yes Gary.in keys_verb_test.dart unit test - "keys verb on enrollment which is not approved"

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
murali-shris: yes. Made the relevant change in isAuthorized method to check for *:rw

gkc: I can see there are tests for delete and update verbs for `*:rw` permissions but I don't see those tests for notify / scan / lookup / llookup  ...

gkc (changes requested):
I can see there are tests for delete and update verbs for *:rw permissions but I don't see those tests for notify / scan / lookup / llookup ...

packages/at_secondary_server/test/local_lookup_verb_test.dart
```diff
@@ -247,6 +250,154 @@ void main() {
...
+    String enrollmentId = Uuid().v4();
+    setUp(() async {
+      await verbTestsSetUp();
+      inboundConnection.metadata.isAuthenticated =
+          true; // owner connection, authenticated
+    });
+
+    test(
+        'A test to verify llookup verb is allowed in all namespace when access is *:r',
+        () async {
```
sitaram-kalluri, line 265: @gkc: Here is the test with *:r in the llookup verb. Any other scenarios should we add apart from this?

gkc: Please do the same test again with `*:rw`

And add tests for 
- scan
- lookup verb, for client authenticated connection only. Note that atServer (pol) connections should be presumed always to have `*:r` / `*:rw` access for all lookups

murali-shris: scan:
we have a test - A test to verify enrollment has *:rw access
@sitaram-kalluri we should add a test to check the scan results of an enrollment that has *:rw access
lookup:
authenticated lookup hits another secondary to retrieve a key. Test will be covered as part of llookup in another secondary
unauthenticated lookup can retrieve any public key. We don't restrict by access for this.
notify:
we didn't make changes for APKAM..we have to investigate and check whether any changes are required e.g. cached keys are updated in notify verb handler.
@sitaram-kalluri can you please create a ticket for notify

gkc: For now, other atServers should behave exactly as they do now - ie no namespace checking on requests from other atServers

When clients do lookup, the namespace access checks should be applied to the cache checks

gkc: As for notify - I do believe that, for client connections, namespace access should be checked for notify:update and notify:delete commands. Otherwise, apps could spam other apps' namespaces with notifications

sitaram-kalluri:
The changes related to the Lookup verb are addressed in the PR: https://github.com/atsign-foundation/at_server/pull/1505

packages/at_secondary_server/test/local_lookup_verb_test.dart
gkc: @sitaram-kalluri
- Is there a ticket for the notify changes discussed above?
- Was a test added for checking the scan results?

gkc (changes requested):
See my most recent comment above https://github.com/atsign-foundation/at_server/pull/1496#discussion_r1302000853

sitaram-kalluri:
> See my most recent comment above [#1496 (comment)](https://github.com/atsign-foundation/at_server/pull/1496#discussion_r1302000853)

@gkc: The ticket for notify verb is created: https://github.com/atsign-foundation/at_server/issues/1444
A test in scan_verb_test.dart is updated, and also the tests in scan_verb_test.dart updated and the changes are available in the following PR: https://github.com/atsign-foundation/at_server/pull/1505

gkc (approved):

PR #1505, opened 2023-08-14
fix: Add authorization check for enrollment in lookup verb handler

sitaram-kalluri:
**- What I did**
- When user is authenticated vai the APKAM mechanism, when executing lookup verb, verify if the user has authorization to the key's namespace.
- Refactor lookup_verb_handler.dart to increase the readability.
- Update the tests in scan_verb_test.dart. Populated enrollmentId to the inbound connection to mimic the connection is authenticated via the APKAM and authorization check for namespaces enrolled is performed.

**- How I did it**
- When user authenticates via the APKAM mechanism, fetch enrollmentId from the connection metadata. Retrieve the enrolled namespaces and verify if user has permission for the key's namespace. If yes, return the data; else throw UnAuthorizedException.
- Refactor 'Lookup_verb_handler.dart' 

**- How to verify it**
- Add unit tests in lookup_verb_test.dart to assert the APKAM enrollment changes. Ran all unit tests and functional tests in local and tests pass.
- Add following scenario in scan verb unit test - A test to verify scan returns all keys when enrollment has *:rw access


packages/at_secondary_server/lib/src/verb/handler/lookup_verb_handler.dart
```diff
@@ -56,105 +56,193 @@ class LookupVerbHandler extends AbstractVerbHandler {
...
+  ///     yields the corresponding value of the key "public:phone.wavi@alice".
+  ///
+  ///   - In cases where the 'atSign' does not possess ownership of the data,
+  ///     a pol authentication is initiated towards the 'atSign' that does own the data.
+  ///
+  ///     For instance, when the current "atSign" is "alice" and the request is
+  ///     "lookup:phone.wavi@bob"(from an authenticated connection), an interaction
+  ///     occurs where Alice proves identity to Bob, who subsequently returns the
+  ///     value of the key "@alice:phone.wavi@bob".
+  Future<void> _handleUnAuthAndPolAuthConnection(
```
gkc: I'd like to see code which is specifically checking the isPolAuthenticated flag. Ideally always there would be three clear cases handled in a verb handler: (1) unauthenticated (2) authenticated (3) polAuthenticated

gkc (changes requested):
In addition to my comment on the verb handler implementation, I have one more question - are there tests which cover 'special' records? (1) 'reserved' keys (2) shared symmetric keys (`shared_key.bob@alice` and `@bob:shared_key@alice`) etc?

sitaram-kalluri:
> In addition to my comment on the verb handler implementation, I have one more question - are there tests which cover 'special' records? (1) 'reserved' keys (2) shared symmetric keys (`shared_key.bob@alice` and `@bob:shared_key@alice`) etc?

Regarding the reserved keys, particularly the key format "@bob:shared_key@alice", when a user authenticates via the APKAM (who has full access), the absence of a namespace associated with the key leads to the failure of the authorization check. As a result, the data retrieval process does not succeed. One approach to deal with this is to have a system level namespace for all the reserved keys. 
@gkc: Can you please share your thoughts on this?

gkc:
> > In addition to my comment on the verb handler implementation, I have one more question - are there tests which cover 'special' records? (1) 'reserved' keys (2) shared symmetric keys (`shared_key.bob@alice` and `@bob:shared_key@alice`) etc?
> 
> Regarding the reserved keys, particularly the key format "@bob:shared_key@alice", when a user authenticates via the APKAM (who has full access), the absence of a namespace associated with the key leads to the failure of the authorization check. As a result, the data retrieval process does not succeed. One approach to deal with this is to have a system level namespace for all the reserved keys. @gkc: Can you please share your thoughts on this?

I suggest that if a record id has no namespace then we should say that the client is authorized. This is not an absolute rule - this rule should be applied any broader permission checks (e.g. records which are protected from deletion, immutable checks when we have immutable records, etc)

sitaram-kalluri:
> > > In addition to my comment on the verb handler implementation, I have one more question - are there tests which cover 'special' records? (1) 'reserved' keys (2) shared symmetric keys (`shared_key.bob@alice` and `@bob:shared_key@alice`) etc?
> > 
> > 
> > Regarding the reserved keys, particularly the key format "@bob:shared_key@alice", when a user authenticates via the APKAM (who has full access), the absence of a namespace associated with the key leads to the failure of the authorization check. As a result, the data retrieval process does not succeed. One approach to deal with this is to have a system level namespace for all the reserved keys. @gkc: Can you please share your thoughts on this?
> 
> I suggest that if a record id has no namespace then we should say that the client is authorized. This is not an absolute rule - this rule should be applied any broader permission checks (e.g. records which are protected from deletion, immutable checks when we have immutable records, etc)

Sure Gary.

gkc (approved):
LGTM

packages/at_secondary_server/lib/src/verb/handler/lookup_verb_handler.dart
```diff
@@ -55,106 +55,228 @@ class LookupVerbHandler extends AbstractVerbHandler {
...
+  }
+
+  /// When an 'atSign' is authenticated via APKAM, retrieves enrollment data based
+  /// on the enrollmentId.
+  ///
+  /// If the user possesses authorization for the [lookupKey] namespace, returns true;
+  /// otherwise, returns false.
+  Future<bool> _isAuthorizedToViewData(
+      InboundConnection atConnection, String lookupKey) async {
+    // If a key does not have namespace then the client is authorized
```
gkc: Please add some further explanation of why this is the case.

gkc (changes requested):

packages/at_secondary_server/lib/src/verb/handler/lookup_verb_handler.dart
sitaram-kalluri: Updated the code comments. Please review.

gkc:
@sitaram-kalluri Before I approve and merge this, please run this server locally and run the at_talk program with 'new' atSigns alice and bob, and verify that everything is working correctly. Please also do any other client program testing that is appropriate.

sitaram-kalluri:
> @sitaram-kalluri Before I approve and merge this, please run this server locally and run the at_talk program with 'new' atSigns alice and bob, and verify that everything is working correctly. Please also do any other client program testing that is appropriate.

Sure Gary.

sitaram-kalluri:
> > @sitaram-kalluri Before I approve and merge this, please run this server locally and run the at_talk program with 'new' atSigns alice and bob, and verify that everything is working correctly. Please also do any other client program testing that is appropriate.
> 
> Sure Gary.

- Tested with at_talk and below send and receive messages. Attaching the recording.
- Functional tests and E2E tests ran succesfully.

https://github.com/atsign-foundation/at_server/assets/16645698/8b9e0e69-b2ca-41ab-aaae-990949913e9a

gkc (approved):

PR #1525, opened 2023-08-25
feat: Filter notifications on monitor for APKAM

### LINKED ISSUE #1444: APKAM - access changes for monitor verb

sitaram-kalluri:
**- What I did**
- Add APKAM functionality to monitor verb. 
- If the connection is authenticate via APKAM, send notifications only if enrollment namespaces contains .* or __manage or enrollment is authorized for the key's namespace.
- Refactor the monitor_verb_hander.dart 
- Add unit tests for monitor verb.
- Reduce the log-level in `_sendNotificationByEnrollmentNamespaceAccess` method because it gets invoked for every notification causing server logs more noisy.

**- How I did it**
- Introduce method `_sendNotificationByEnrollmentNamespaceAccess` which verify if the enrollment is authorized to send notifications when connection is authenticated via APKAM
- The existing logic is moved to `_sendLegacyNotification` which gets invoked when the connection is authenticated via PKAM verb.
- Change the name of the method `processReceivedNotification` to `_sendNotificationToClient` to better reflect the purpose of the method.

**- How to verify it**
- All the tests should pass.

- Add APKAM functionality to monitor verb.


packages/at_secondary_server/lib/src/verb/handler/monitor_verb_handler.dart
```diff
@@ -36,74 +40,144 @@ class MonitorVerbHandler extends AbstractVerbHandler {
...
+      return;
+    }
+    if (enrollDataStoreValue.namespaces.isEmpty) {
+      logger.info('No namespaces are enrolled for the enrollmentId:'
+          ' ${(atConnection.getMetaData() as InboundConnectionMetadata).enrollmentId}');
+      return;
+    }
+    // separate namespace from the notification key
+    String namespaceFromKey = notification.notification!
+        .substring(notification.notification!.indexOf('.') + 1);
```
gkc, line 157: Need to handle situation where there is no namespace

gkc (changes requested):
Left a comment

packages/at_secondary_server/lib/src/verb/handler/monitor_verb_handler.dart
sitaram-kalluri: Added a check before this line to check if the `notification.notification` contains ".". If not available, it indicates the namespace is not present and do nothing.

packages/at_secondary_server/lib/src/verb/handler/monitor_verb_handler.dart
```diff
@@ -13,7 +17,7 @@ class MonitorVerbHandler extends AbstractVerbHandler {
 
   late InboundConnection atConnection;
 
-  String? regex;
+  late String regex;
```
gkc, line 20: neither the inbound connection nor the regex should be instance variables

gkc (changes requested):
neither the inbound connection nor the regex should be instance variables in MonitorVerbHandler

sitaram-kalluri:
> neither the inbound connection nor the regex should be instance variables in MonitorVerbHandler

Removed atConnection and regex as instance variables in the monitor_verb_handler.dart. The corresponding changes in at_persistence_secondary_server are in the following PR: https://github.com/atsign-foundation/at_server/pull/1531

packages/at_secondary_server/lib/src/connection/inbound/inbound_connection_metadata.dart
```diff
@@ -26,4 +26,8 @@ class InboundConnectionMetadata extends AtConnectionMetaData {
 
   /// A unique identifier generated for a client's APKAM enroll request
   String? enrollmentId;
+
+  /// This is used by the connections where monitor is set to true to filter notifications
+  /// based on the regular expression.
+  String regex = '.*';
```
gkc: I don't believe we should have this here. Let's discuss.

sitaram-kalluri:
The changes are deferred. Hence closing the PR.

packages/at_secondary_server/lib/src/verb/handler/monitor_verb_handler.dart
sitaram-kalluri: As discussed, the changes are deferred

packages/at_secondary_server/lib/src/connection/inbound/inbound_connection_metadata.dart
sitaram-kalluri: Removed this change.

gkc (approved):

PR #1529, opened 2023-08-29
fix: Return keys based on enrollmentId and refactor scan_verb_handler…

sitaram-kalluri:
….dart

**- What I did**
- Fix the issue, when connection is authenticated via APKAM, and if enrollment have ".*" namespace, all the keys should be returned.
- Refactored the code the optimize the memory usage by removing the temporary list to filter keys. Instead iterate though the index and update the keys in the list.

**- How I did it**
- 1. To fix the issue related to keys not being returned when enrollment has ".*", add a check  in `_filterKeysBasedOnEnrollmentId` where if enrollment namespace have ".*" return all the keys 
- 2. In `_getLocalKeys` method, remove the temporary list to modify the key. Instead use index to fetch the data and iterate through the list and modify the key.

**- How to verify it**
- All test should pass.

### DIFF
--- a/packages/at_secondary_server/lib/src/verb/handler/scan_verb_handler.dart
+++ b/packages/at_secondary_server/lib/src/verb/handler/scan_verb_handler.dart
@@ -59,62 +59,24 @@ class ScanVerbHandler extends AbstractVerbHandler {
     var scanRegex = verbParams[AT_REGEX];
     var showHiddenKeys = verbParams[showHidden] == 'true' ? true : false;
 
-    // Throw UnAuthenticatedException.
-    // When looking up keys of another atsign and connection is not authenticated,
-    if (forAtSign != null && !atConnectionMetadata.isAuthenticated) {
-      throw UnAuthenticatedException(
-          'Scan to another atsign cannot be performed without auth');
-    }
     try {
-      // If forAtSign is not null and connection is authenticated, scan keys of another user's atsign,
-      // else scan local keys.
       var currentAtSign = AtSecondaryServerImpl.getInstance().currentAtSign;
-      var enrollNamespaces = {};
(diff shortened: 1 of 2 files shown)


packages/at_secondary_server/lib/src/verb/handler/scan_verb_handler.dart
```diff
@@ -163,51 +126,47 @@ class ScanVerbHandler extends AbstractVerbHandler {
...
   /// **Returns**
   ///
-  /// Returns the list of keys of current atsign.
-  List<String> _getLocalKeys(InboundConnectionMetadata atConnectionMetadata,
-      List<String> keys, bool showHiddenKeys) {
-    List<String> keysList = [];
-    // Verify if the current user is authenticated or not
-    // If authenticated get all the keys except for private keys
-    // If not, get only public keys
+  /// Returns the list of keys of current atSign.
```
murali-shris: Change the method documentation
/// Filter keys based on authentication type of inbound connection
/// <describe the type of keys for each auth type>

Remove the inline code comments before each auth type

packages/at_secondary_server/lib/src/verb/handler/scan_verb_handler.dart
```diff
@@ -229,4 +188,68 @@ class ScanVerbHandler extends AbstractVerbHandler {
         key.startsWith('public:_') ||
         key.startsWith('_');
   }
+
+  /// Filter and returns keys whose namespaces are authorized for the given
```
murali-shris: /// Filter keys based on namespace access for the given enrollmentId

murali-shris (changes requested):

murali-shris (approved):

PR #1531, opened 2023-08-31
fix: Refactor AtNotificationCallback.dart

sitaram-kalluri:
**- What I did**
- Refactor `AtNotificationCallback` as a part of removing the changes in the PR: https://github.com/atsign-foundation/at_server/pull/1525
- Modify `registerNotificationCallback` to register and invoke the callback method inside the monitor_verb_handler.dart upon storing the notification to keystore.
- Mark the ununsed methods as deprecated.

**How I did**
- Earlier to this changes, in at_secondary_server, the InboundConnection is an instance variable in MonitorVerbHandler.dart. For a new monitor request, a new MonitorVerbHandler.dart is created. If a connection is closed or terminated, to prevent the callback invoking the instance of MonitorVerbHander, we unregister from the callback. 
- With the new changes, Since, InboundConnection is not a instance variable in MonitorVerbHandler.dart, and there be only one instance of MonitorVerbHandler.dart. So, we do not have to unregister the callback, hence removed the method.
- Earlier to the changes, "NotificationFunction" is used to track if the connection associated with the MonitorVerbHandler.dart is vaild or not which is no longer required. Hene, removed "NotificationFunction" class.


packages/at_persistence_secondary_server/lib/src/notification/at_notification_callback.dart
```diff
@@ -12,22 +12,19 @@ class AtNotificationCallback {
...
-    functions ??= <NotificationFunction>[];
-    _removeUnregisteredFunctions(functions);
-    functions.add(nf);
-    callbackMethods[notificationType] = functions;
+    notificationCallback = callback;
   }
 
   /// Method to deregister callback function
+  @Deprecated('No longer in use')
   void unregisterNotificationCallback(
```
murali-shris: how do we unregister callback after your refactored code?

murali-shris (changes requested):

packages/at_persistence_secondary_server/lib/src/notification/at_notification_callback.dart
sitaram-kalluri: With the latest code changes, we fetch the list of connections for every notifications and write to the connections whose `isMonitor` is set to true. So, we will not have stale connections and does not require to unregisterNotificationCallback on closing the connection.

gkc:
@sitaram-kalluri I've changed my mind, we should keep the methods but mark them deprecated, as you had done before

gkc:
@sitaram-kalluri I'm unsure now as to whether we need this PR at all.

sitaram-kalluri:
The changes are deferred. Hence closing the PR.

PR #1594, opened 2023-10-04
fix: otp expires after use

srieteja:
**- What I did**
- Ensure that OTP expires after use and cannot be reused

**- How I did it**
- Remove the OTP from cache after it is validated

**- How to verify it**
- unit and functional tests added

fix: ensure otp expires after use


srieteja:
This fix has been merged to trunk through https://github.com/atsign-foundation/at_client_sdk/issues/1127

PR #1598, opened 2023-10-09
feat: Implement OTP Store and enable client to set OTP expiry

### LINKED ISSUE #1127: Allow client to configure the expiry on the OTP

Currently the `otp:get` returns an OTP from the server. The expiry of the OTP lies with the server. Enhance the code for the client to optionally set the expiry for the `opt:get`

Add an optional parameter to the "otp" verb to set the expiry limit.

sitaram-kalluri:
**- What I did**
- Enable the client to pass the OTP expiry duration, beyond which the OTP becomes invalid.

**- How I did it**
- Update the verb syntax for "OTP" to include a "TTL" parameter, which represents the duration in seconds.
- Introduce a new class called "OTPStore" to store OTPs along with their expiration durations.
- Implement a periodic job configured to run at specified intervals, responsible for removing expired OTPs from the "OTPStore"
- Configure the duration for this job using the "otp"-> "gcDuration" setting in the config.yaml file. By default, it runs every 1 hour.

- Changes in OtpVerbHandler.dart

  - Replace the "ExpireCache" - a third party library which is used to manage OTPs, with an instance of the "OTPStore" class. Since "ExpiryCache" is no longer used, removed it from the pubspec.yaml
  - In the constructor of the "OtpVerbHandler" class, initialize both the "OTPStore" and the "gcDuration." These are used to manage OTPs and determine how often the job for removing expired OTPs should run.
  - Introduce a new parameter called "otpExpiryDuration," which represents the duration beyond which an OTP expires. If the client does not specify a value for "TTL," it defaults to 5 minutes.
  - Implement the "isOTPValid" method, which returns a boolean value indicating whether an OTP is valid or not.
 
**- How to verify it**
Following are the unit tests:
 - A test to verify otp:get with TTL set is active before TTL is met
 - A test to verify otp:get with TTL set expires after the TTL is met
 - A test to verify gc removes the expired keys from the otp store

Following are the functional tests:
 - A test to generate OTP and returns valid before OTP is not expired
 - A test to generate OTP and returns invalid when TTL is met

 - Because the OTPs cannot be reused, updated the functional tests in enroll_verb_test.dart to fetch otp for each request

- Enable clients to configure OTPs expiry duration


gkc (changes requested):
Since OTP lifespans can extend to weeks rather than just minutes or seconds, the OTPs need to persist through server restarts.

sitaram-kalluri:
Closing this PR. The changes are pushed as a part of the following PR: https://github.com/atsign-foundation/at_server/pull/1609

PR #1609, opened 2023-10-12
feat: Persist OTPs in keystore and prevent reuse of OTPs

### LINKED ISSUE #1127: Allow client to configure the expiry on the OTP

sitaram-kalluri:
**- What I did**
- Store OTPs in the Hive keystore along with a "TTL" (Time To Live) value, representing the OTP's expiration time.
- The OTPs are stored in the format - private:< OTP >atsign. The "private:" ensure that keys are not synced to the client and keys are not returned in the scan verb.
- By default, the "TTL" is set to 5 minutes but can be overridden by the client by specifying the expiry time in milliseconds when using the "OTP" verb.
- In abstract_verb_handler.dart, "isOTPValid" function which verifies if the OTP is valid and removes the OTP from the keystore to prevent reuse. The otp_verb_handler and enroll_verb_handler will invoke the function.
- Replace the use deprecated constants with "AtConstants"

**- How to verify it**
- Following are the unit tests:
   - A test to verify otp:get with TTL set is active before TTL is met
   - A test to verify otp:get with TTL set expires after the TTL is met
   - A test to verify "isOTPValid" returns false if OTP is not available in the keystore.
   - A test to verify otp:validate invalidates an OTP after it is used

- Following are the functional tests:  
    - A test to generate OTP and returns valid before OTP is not expired
    - A test to generate OTP and returns invalid when TTL is met
    - Because the OTPs cannot be reused, updated the functional tests in enroll_verb_test.dart to fetch otp for each request

- Enable client to set the OTP expiry duration.


packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -116,14 +116,18 @@ class EnrollVerbHandler extends AbstractVerbHandler {
...
-    if (!atConnection.getMetaData().isAuthenticated) {
-      var otp = enrollParams.otp;
-      if (otp == null ||
-          (await OtpVerbHandler.cache.get(otp.toString()) == null)) {
+
+    // OTP is sent only in enrollment request which is submitted on
+    // unauthenticated connection.
+    if (atConnection.getMetaData().isAuthenticated == false) {
+      Response otpResponse = await OtpVerbHandler(keyStore)
+          .processInternal('otp:validate:${enrollParams.otp}', atConnection);
```
gkc: I don't like the idea of verb handlers using internal functionality of other verb handlers. Is there a better / cleaner way of making that validation functionality more explicitly available?

gkc (changes requested):
I don't like the idea of verb handlers using internal functionality of other verb handlers. Is there a better / cleaner way of making that validation functionality more explicitly available?

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
sitaram-kalluri: Moved the "isOTPValid" method to the abstract_verb_handler.dart which wil be invoked by otp_verb_handler and enroll_verb_handler.

packages/at_secondary_server/lib/src/verb/handler/abstract_verb_handler.dart
```diff
@@ -152,4 +152,27 @@ abstract class AbstractVerbHandler implements VerbHandler {
...
+    String otpKey =
+        'private:${otp.toLowerCase()}${AtSecondaryServerImpl.getInstance().currentAtSign}';
+    AtData otpAtData;
+    try {
+      otpAtData = await keyStore.get(otpKey);
+    } on KeyNotFoundException {
+      return false;
+    }
+    // Remove the key from keystore to prevent reuse of OTP.
+    await keyStore.remove(otpKey);
```
gkc, line 175: Is this safe?

This method is called in two places
1. EnrollVerbHandler._handleEnrollmentRequest : If the server is restarted after this method has been called, but before the handling of of the enrollment request has completed, then the OTP will have been deleted and when the client retries the OTP will no longer exist. Feels like the decision to remove the OTP ought to be made by the enrollment verb handler as its final piece of processing ...?
2. OtpVerbHandler for the `validate` operation : What is the purpose of the `validate` operation? Who calls it in what circumstances and why?

gkc: Is this safe?

This method is called in two places
1. EnrollVerbHandler._handleEnrollmentRequest : If the server is restarted after this method has been called, but before the handling of of the enrollment request has completed, then the OTP will have been deleted and when the client retries the OTP will no longer exist. Feels like the decision to remove the OTP ought to be made by the enrollment verb handler as its final piece of processing ...?
2. OtpVerbHandler for the `validate` operation : What is the purpose of the `validate` operation? Who calls it in what circumstances and why?

gkc (changes requested):

packages/at_secondary_server/lib/src/verb/handler/abstract_verb_handler.dart
sitaram-kalluri: On #‎2. The original purpose of the "validate" operation within the OTP verb was to validate the OTP. However, at present, we invoke "isOTPValid" in abstract_verb_handler.dart which retrives the OTP from the keystore for validation. So, the "validate" operation within OTPVerbHandler is not being utilized.

gkc: @sitaram-kalluri 
1. What is your response re this point?
2. So, should the `validate` operation be removed?

sitaram-kalluri: @gkc: On #‎1, will make the code changes to remove the OTP at the end of processing the enrollment.
On #‎2 Removing validate OTP operation make sense to me. @murali-shris : Please share your thoughts on this.

murali-shris: @sitaram-kalluri yes.we can remove validate operation since it is not used anywhere.

packages/at_secondary_server/pubspec.yaml
```diff
@@ -35,6 +34,13 @@ dependencies:
   yaml: 3.1.2
   logging: 1.2.0
 
+dependency_overrides:
+  at_commons:
+    git:
+      url: https://github.com/atsign-foundation/at_libraries.git
+      path: packages/at_commons
+      ref: expire_otp_changes
```
gkc: I've merged the at_commons PR so please publish at_commons 3.0.57 and remove this dependency override

gkc (changes requested):

packages/at_secondary_server/pubspec.yaml
sitaram-kalluri: Sure Gary, Thank you.

sitaram-kalluri:
Updated the at_commons version to 3.0.57.

packages/at_secondary_server/lib/src/verb/handler/otp_verb_handler.dart
```diff
@@ -1,24 +1,24 @@
...
+
+  @visibleForTesting
+  int otpExpiryInMills = Duration(minutes: 5).inMilliseconds;
 
   OtpVerbHandler(SecondaryKeyStore keyStore) : super(keyStore);
 
   @override
-  bool accept(String command) =>
-      command == 'otp:get' || command.startsWith('otp:validate');
+  bool accept(String command) => command.startsWith('otp');
```
gkc: Should be `command.startsWith('otp:get')`

gkc: or `command == 'otp:get'`

packages/at_secondary_server/lib/src/verb/handler/otp_verb_handler.dart
```diff
@@ -29,6 +29,10 @@ class OtpVerbHandler extends AbstractVerbHandler {
       HashMap<String, String?> verbParams,
       InboundConnection atConnection) async {
     final operation = verbParams['operation'];
+    if (verbParams[AtConstants.ttl] != null &&
+        verbParams[AtConstants.ttl]!.isNotEmpty) {
+      otpExpiryInMills = int.parse(verbParams[AtConstants.ttl]!);
+    }
     switch (operation) {
       case 'get':
```
gkc, line 37: Should have a `default` block which will throw an exception for any operation other than `get`. Even though currently we only have one operation, switch statements should always cover all cases (which in this case means we need a default handler.)

gkc (changes requested):

gkc (approved):

PR #1616, opened 2023-10-14
feat: at_root_server: rate limiting for invalid atSigns; misc logging and other cleanup

### LINKED ISSUE #1590: Noise in logs from monitoring connections
### Describe the bug

Originally noted in https://github.com/atsign-foundation/operations/issues/81

The logs are full of errors like:

```
SEVERE|2023-09-04 14:25:54.091540|RootServerImpl|rootServer Socket Error :HandshakeException: Connection terminated during handshakeconnecting to InternetAddress('0.0.0.0', IPv4)
```

These appear to be due to Uptime robot checks, and now (much more frequently) liveness checks within the k8s cluster.

### Steps to reproduce

1. First I make a TCP connection
2. Then I don't proceed with TLS, but just close the connection
3. And then the atDirectory throws a `SEVERE` exception to the logs

### Expected behavior

TCP connections that are closed before TLS should be treated as normal and not result in an exception.

### Additional context

@gkc I'd appreciate your help on this, as I think you've been through the code fairly recently (it's been mostly untouched for a long time now).

### LINKED ISSUE #1599: Rate limiting on root

Pen test item 4.4.2 identified that the atDirectory can be enumerated

Initially:

1. Drop connection on an invalid atSign query

Down the line:

2. Rate limit clients that make repeated invalid connections
3. Tarpit clients that continue

### LINKED ISSUE #1615: root server has unnecessary 10-second delay while starting up
### Describe the bug

- During root server startup there is a loop which waits for certificates to become available
- If the certificates are present on the file system then we break out of the loop
- The intent is that if the certificates are not present then we sleep for 10 seconds before trying again
- However the code as written results in a 10-second delay even if the certificates were present

### Steps to reproduce

- Start the root server with certificates present. Observe the 10-second delay

### Expected behavior

- When the certificates are present, there should be no delay while starting up

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

_No response_

gkc:
**- What I did**
- Resolve https://github.com/atsign-foundation/at_server/issues/1590
- Resolve https://github.com/atsign-foundation/at_server/issues/1615
- Partially resolve https://github.com/atsign-foundation/at_server/issues/1599
  - @cpswan This PR implements a variation on item 1 in #1599 - see "How to verify it" below
- Miscellaneous other logging cleanup
- update and pin dependencies
- bump version to 2.1.0 and update changelog

**- How I did it**
- See this PR's commits

**- How to verify it**
- https://github.com/atsign-foundation/at_server/issues/1615
  - Run local root server (see `at_redis` and `at_root` scripts in `tools/run_locally` directory in this repo)
    - Server should start up without delay if the certificates are present
- https://github.com/atsign-foundation/at_server/issues/1590
  - Open a plain socket connection to it then disconnect (e.g. by `curl http://vip.ve.atsign.zone:64`)
    - No log message should be generated
- https://github.com/atsign-foundation/at_server/issues/1599
  - Open a TLS connection (e.g. via `openssl s_client -connect vip.ve.atsign.zone:64`)
    - Look up a non-existent atSign: the client will receive a 'null' response
    - Do the lookup again: after a 3-second delay, the client will receive a 'null' response
    - Do the lookup again: after a 3-second delay, the client connection will be terminated


cpswan (approved):

gkc:
@cpswan @cconstab I can see that [a 'build and deploy to dev k8s' workflow](https://github.com/atsign-foundation/at_server/actions/runs/6525842261) ran after I merged this PR. The workflow appeared to run successfully but the new code is not running at root.atsign.wtf - is there something else I need to do?

PR #1633, opened 2023-10-20
test: Refactor end2end tests into functional tests

### LINKED ISSUE #1593: End2end tests that pass without being able to reach another atServer
### Describe the bug

`35 tests passed, 31 failed.` when cicd3 couldn't connect to cicd4 (and similarly cicd5 & 6)

### Steps to reproduce

1. First I configure cicd atServers to connect to wrong atDirectory
2. Then I run end2end tests
3. And then many of them pass even though remote atServers are unreachable

https://github.com/atsign-foundation/at_server/actions/runs/6402399843/job/17385865736 provides an example

### Expected behavior

End2end tests should only pass when another atServer is involved.

Tests that don't need another atServer should be refactored into functional tests.

If the run time for functional tests gets too long we can break them into batches.

### Additional context

This happened whilst migrating the cicd atServers, and a bunch of them were pointed at the wrong root server and thus unable to connect to their opposite atServer
[at_server_logs_6651.zip](https://github.com/atsign-foundation/at_server/files/12803282/at_server_logs_6651.zip)

purnimavenkatasubbu:
**- What I did**
Refactored end2end tests into functional tests
**- How I did it**
Moved tests that need only a single atsign from end2end tests to functional tests.
**- How to verify it**
All tests should pass


gkc (approved):
LGTM. Thank you @purnimavenkatasubbu

PR #1634, opened 2023-10-20
fix: root deployment issue

kumarnarendra701:
**- What I did**
- Added GitHub runner number with dev_env tag
- Added rollout section for prod deployment

**- How I did it**
- Added git runner number and rollout command 

**- How to verify it**
- GitHub action logs


.github/workflows/at_server_dev_deploy.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
           context: packages/at_root_server
           push: true
           tags: |
-            atsigncompany/root:dev_env
+            atsigncompany/root:dev_env-gha${{ github.run_number }}
```
cpswan, line 41: We still need a `:dev_env` tag (without a run number), and there's a tag with the (branch and) run number right there on the next line that we could use for the kubectl update.

cpswan (changes requested):

.github/workflows/at_server_dev_deploy.yaml
kumarnarendra701: @cpswan - We added this for keeping the latest code with the "dev_env" tag, even though we don't require it.

cpswan: @kumarnarendra701 *you* might not need the `:dev_env` tag to stay up to date, but it could be used elsewhere.

I see two choices:

1. Add another tag for `:dev_env-gha${{ github.run_number }}`
2. Use `:dev_env-${{ env.BRANCH }}-gha${{ github.run_number }}` for the deployment

kumarnarendra701: Modified as first choice.

cpswan: @kumarnarendra701 There should be 3 tags here. The 2 that we started with, and the new one that you've chosen to add.

kumarnarendra701: @cpswan  - Can you please review if my understanding is correct?

cpswan (approved):

PR #1713, opened 2023-12-20
feat: do NOT add delete entries in commit log when expired keys are deleted

### LINKED ISSUE #1660: Improve commit log handling for expired (`expiresAt` has passed) records

Summary
- We are managing deletions of expired records on the client side via server creating delete entries in the commit log, and the sync process. This is unnecessary as clients also have the ability to detect expired records and delete them.
- As a result, the commit log grows every time a record expires. So if you create a million records over a period of a week, all of which expire within an hour, you will have a commit log with a million entries in it, which will stay there forever

Detail
- `@alice` sends notification to `@bob` with ttr:-1 (cache indefinitely on bob's atServer) and ttl:3600000 (1 hour)
  - or does a put with ttr:-1 and ttl:36000000
- when bob's (and alice's, in the case of a `put` rather than a `notify`) atServers finds the expired records during a sweep, currently a record is added to the commit log
- as a result, the commit log grows every time a record with a new key is shared / notified
- this isn't necessary, since client-side storage can detect locally stored (e.g. by sync) expired records and delete them

- client-side, when a record is fetched, if it has expired then it should be deleted (and not added into local uncommitted entries)
- client-side, there should be regular background sweeps for expired records - or, at a minimum, at startup time (i.e. creation of LocalSecondary object)
- server-side, when a record is fetched, if it has expired then it should be deleted, and the existing commit entry should be deleted, and no new commit entry should be added

_No response_

### Additional context

_No response_

srieteja:
closes https://github.com/atsign-foundation/at_server/issues/1660
**- What I did**
- Do not add CommitEntry for Delete Op on expired keys (disabled for now, will be enabled later)
- Introduced an optional param in scheduleKeyExpireTask() called runInterval which takes Duration and schedules the expiryTask at this timeInterval. (This is an improvement as previously this method would accept only time in mins and in type String)

**- How I did it**
- Introduce an optional named param in deleteExpiredKeys() called skipCommit, this decides whether a commit should be skipped when an expired key is deleted. [defaults to false; meaning commit will be created]
- Based on the method param above, set the skipCommit flag while calling remove() from deleteExpiredKeys() [in hiveKeyStore and notificationKeyStore]

**- How to verify it**
- new unit tests added should verify the behaviour of expired keys w.r.t. the changes

feat: do NOT add delete entries in commit log when expired keys are deleted


murali-shris (dismissed):

gkc:
@srieteja We have a chicken and egg scenario here; until we can be sure that most apps are using the latest AtClient, those clients will still be dependent on the commit log and sync responses to trigger deletion on the client side.

Therefore for now we need to
- add a flag to the atServer's config for this feature, defaulted to false (not enabled) for now
- and add a way of setting that flag dynamically for testing purposes (add it to `ModifiableConfigs`, add a subscriber somewhere, etc etc)
- and modify tests so they verify both the current behaviour with the flag not enabled, and the new behaviour with the flag enabled
- and add a test to assert that the flag is by default not enabled

At some time in the future, we can change the config to enable this feature by default. I'd suggest create a ticket for that and reference that new ticket in the description of #1660 and we'll keep assigning the ticket to the "sprint after next" until we think we're ready to go.

Make sense? cc @murali-shris @sitaram-kalluri

gkc (changes requested):
See comment I left

packages/at_persistence_secondary_server/CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ## 3.0.60
+- feat: delete entries for expired keys are not committed to the commitLog [feature not enabled yet]
```
sitaram-kalluri, line 2: FYI, The package with 3.0.60 is published. Please upgrade the version

sitaram-kalluri (dismissed):

packages/at_persistence_secondary_server/CHANGELOG.md
srieteja: addressed in [b1c9493](https://github.com/atsign-foundation/at_server/pull/1713/commits/b1c949376913a6f090efed26cf6099ce7bde78bd)

srieteja:
Added flags in hive_manages/scheduleKeyExpiryTask() and hive_keystore/deleteExpiredKeys() to enable the feature being written in this PR. Also introduced config in AtSecondaryServer/AtSecondaryConfig and passed the same to the hive_manager/keystore to enable the feature in the PR (see https://github.com/atsign-foundation/at_server/pull/1727)

srieteja:
I have tested the functionality and how this feature interacts with the client. @purnimavenkatasubbu has tested this along with https://github.com/atsign-foundation/at_server/pull/1727 and https://github.com/atsign-foundation/at_client_sdk/pull/1187 with buzz and wavi. Everything seems to be working fine, Ready to be reviewed

gkc:
@srieteja in the PR description you say that you 
> - set the skipCommit flag to true in deleteExpiredKeys() [in hiveKeyStore and notificationKeyStore]

that is only true for notificationKeyStore. In hiveKeyStore deleteExpiredKeys takes an "optimizeCommits" flag

packages/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
```diff
@@ -341,7 +340,9 @@ class HiveKeystore implements SecondaryKeyStore<String, AtData?, AtMetaData?> {
 
   @override
   @server
-  Future<bool> deleteExpiredKeys() async {
+  @client
+  Future<bool> deleteExpiredKeys({bool? optimizeCommits = false}) async {
```
gkc: Why not call it 'skipCommit' same as the existing flag name?

gkc (changes requested):

packages/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
srieteja: makes more sense, let me get that changed

srieteja: change made as part of [fefa7d4](https://github.com/atsign-foundation/at_server/pull/1713/commits/fefa7d41067053944799aff620786d1a84b19694)

gkc (approved):
LGTM. Nice work on the tests

gkc:
> @srieteja in the PR description you say that you
> 
> > * set the skipCommit flag to true in deleteExpiredKeys() [in hiveKeyStore and notificationKeyStore]
> 
> that is only true for notificationKeyStore. In hiveKeyStore deleteExpiredKeys takes an "optimizeCommits" flag

@srieteja Please can you correct the PR description so it is accurate

srieteja:
Updated the PR description @gkc

PR #1721, opened 2023-12-27
fix: Refactor functional tests

### LINKED ISSUE #1649: Refactor at_server functional tests
Currently the "read" method in the at_functional tests in at_server package is not reliable. It fails when the server responses are sent in multiple packets. Hence refactor the functional tests to use the "read" method in the at_lookup package.

sitaram-kalluri:
**- What I did**
- In instances where the server holds numerous keys, typically during end-of-test scenarios, the scan responses become large, leading to the transmission of keys in multiple packets. The current "read" method fails in capturing the entirety of the server response.
- There are situations where the same "atKey" is utilized across multiple tests, potentially resulting in test failures due to inconsistent values. Especially in scenarios related to sync and caching of keys.
- Additional atSigns are appended to the list for testing enrollment scenarios. However, the ports are omitted in the "docker run" command leading to the atSigns unusable.

**- How I did**
- Replace the "read" method in functional test to capture the server response with the "read" method in at_lookup_impl to handle the partial responses from the server.
- Removed at_demo_credentials and used at_demo_data package. Also add APKAM private and public key for Alice atSign to test enrollment scenario's. 
- Moved the utils classes from "test" folder to "src" folder.
- Add unique-Id for each key to prevent same key being reused which perhaps might cause inconsistent behaviour in test.
- Add regex to the scan verb to prevent large numbers of keys being from the server because of which the intended key might be missed due to overloading of buffer capacity.
- Updated the mounting of ports in docker run command for the new atSigns added.


gkc:
@sitaram-kalluri Please add an explanation to the PR description of its purpose (i.e. what problems are being addressed by the PR)

gkc (commented):
LGTM but please explain the purpose of the PR

gkc (approved):

PR #1731, opened 2024-01-08
feat: Enable OTP verb handler to store Semi permanent passcodes

### LINKED ISSUE #1661: APKAM: Increase enrollment usability by adding ability to create multi-use 'semi-permanent' enrollment passcodes

In order to prevent spam, APKAM enrollment requests require a valid 'OTP' (one time passcode) to be provided in order for the AtServer to accept them and notify approving apps of the request. If we did not require a valid OTP then any software anywhere would be able to bombard anyone's atSign with spurious enrollment requests which would all be delivered to approving apps

However the requirement for a passcode reduces the usability of the enrollment workflow - e.g.
  - The user installs and launches an app 'NewApp'
  - Is asked to switch to another app 'ApprovingApp' to get a passcode
  - Switches back to 'NewApp' to type in the passcode
  - Is asked to switch back to 'ApprovingApp' to approve the enrollment request
  - And then switches back to 'NewApp'

Extend the otp verb and otp record handling so that 
  - One 'special' otp (let's call it an `spp` semi-permanent passcode) which user can provide and can be stored to AtServer by apps which have full access to the __manage namespace
  - which does not expire
  - and which can be used an arbitrary number of times without being deleted
  - and can be updated by the user whenever they like

This both preserves anti-spam protections AND enables a simpler enrollment flow where
- user types in their memorable passcode
- and has two context switches (switch to 'ApprovingApp', approve, then switch back to NewApp) instead of four

In the event that someone guesses the user's passcode and they start to be bombarded with spurious enrollment requests then the user should have the ability to change their spp

In order to limit the likelihood of the spp being guessed thus allowing enrollment spamming to commence, the spp should be case-insensitive alphanumeric 6 characters minimum

We should also implement detection of spp-guessing brute-force attacks, e.g.
- when an enrollment request is received with an invalid passcode
- then the server should wait for $delayTime before responding and closing the connection
- and $delayTime should increase (e.g. from 1 to 2 to 3 to 5 to 8 to 13 to 21 seconds) when subsequent invalid enrollment requests are received within a certain period (e.g. 60 seconds) or decrease back to 1 second if no invalid enrollment requests have been received for some period (e.g. 60 seconds)

_No response_

### Additional context

_No response_

sitaram-kalluri:
**- What I did**
- Store the Semi-Permanent Pass Codes(SPP) into the Key-Store. Only the connections which has access to the "__manage" namespace are allowed to store the SPP.

**- How I did it**
- Enhance the OTPVerbHandler.dart to accommodate the processing of "otp:put" requests, containing the Secure Processing Payload (SPP) transmitted from the client to the server.
- When the connection has access to the "__manage" namespace, extract the SPP and securely store it in the Key-Store under the key "private:spp@alice".
- In AbstractVerbHandler.dart, update the "isOTPValid" method responsible for validating OTP/SPP received during enrollment requests. Now, it first checks if the received OTP is a valid SPP; if not, it verifies if it is a valid OTP. If both checks fail, the method returns false.
- Within EnrollVerbHandler.dart, implement a mechanism to counter brute-force attacks. Upon receiving an invalid OTP, introduce a response delay. The delay is progressively increased following the Fibonacci sequence until it reaches the specified threshold value. Once the threshold is reached, reset the delay to 1 second 

**- How to verify it**
- Add Unit tests to assert the above functionality

- APKAM: Increase enrollment usability by adding ability to create multi-use 'semi-permanent' enrollment passcodes


packages/at_secondary_server/config/config.yaml
```diff
@@ -158,4 +158,9 @@ enrollment:
   # The maximum number of requests allowed within the time window.
   maxRequestsPerTimeFrame: 5
   # The duration of the time window in hours.
-  timeFrameInHours: 1
\ No newline at end of file
+  timeFrameInHours: 1
+  # The threshold value for the delay interval in seconds.
+  # When the last delay in '_delayForInvalidOTPSeries' surpasses this threshold,
+  # the delay in responding for Invalid OTP is reset to 1 second to prevent
+  #  excessively long delay intervals.
```
gkc: - The delay should not be reset if there continue to be enroll requests with invalid OTPs, but should stay at the maximum (55 seconds)
- To put it another way - the delay should only be reset if <some duration, e.g. 89 seconds> has passed since the last enroll request which had an invalid OTP

gkc (changes requested):

packages/at_secondary_server/config/config.yaml
sitaram-kalluri: This is addressed in https://github.com/atsign-foundation/at_server/pull/1731/commits/fe650b3783b996b824dbe445d758026f31248eef . Please review the changes.

gkc: Thanks @sitaram-kalluri. Please can you make changes now to enable all of these delay-testing tests to run much more quickly (order of tens of milliseconds, not tens of seconds). You'll need to adjust various aspects of the implementation to do so.

gkc: PS The code comments above are no longer accurate, please update them to reflect the actual behaviour.

gkc (changes requested):

gkc (approved):

PR #1776, opened 2024-02-03
feat: at_server_spec: BREAKING: make AtConnection generic; make it more Dart-idiomatic

gkc:
**Note to reader**
This PR is primarily about some breaking changes to at_server_spec and then a lot of changes in at_secondary_server to take up the at_server_spec changes.

The PR's purpose is to pave the way for other types of InboundConnections to an atServer (most notable, a WebSocket connection) and provide maximum reuse of atServer's connection management and verb handler code - the objective is to make it so that the verb handlers just see a different type of InboundConnection - i.e. an InboundConnection where the underlying is a WebSocket rather than a Socket

To review:
- Review the at_server_spec changes first.
- Most of the changes in at_secondary_server are to deal with the newly generic nature of AtConnection and InboundConnection, and the renaming of `getMetaData()` and `getSocket()` to getters `metaData` and `underlying` respectively
- Also renamed BaseConnection to BaseSocketConnection as it is Socket-specific, and made its Socket non-nullable. This had a ripple effect through the test pack to ensure that a mockSocket is always passed to the constructor (there were multiple cases where a null was being passed)

After approval of this PR, I will
- publish v4.0.0 of at_server_spec
- and then create a followup PR which removes at_secondary_server's dependency override for at_server_spec and has it use the published v4.0.0

## What I did
**at_server_spec**
- feat: at_server_spec: BREAKING: make AtConnection generic (i.e. not socket-specific) This is in order to enable reuse of the atServer's InboundConnectionManager class and InboundConnection interface for WebSocket connections as well as Socket connections
- feat(style): at_server_spec: BREAKING: rename AtConnection's `getSocket()` method to a getter called `underlying` and have it return the generic type for more idiomatic Dart code
- feat(style): at_server_spec: BREAKING: rename AtConnection's `getMetaData()` method to a getter called `metaData` for more idiomatic Dart code
- build: at_server_spec: update pubspec version to 4.0.0
- build: at_server_spec: upgrade dependencies

**at_secondary_server**
- feat: at_secondary_server: implement changes stemming from making AtConnection generic
- feat: atServer: replace `getMetaData()` and `getSocket()` with `metaData` and `underlying` respectively in all code related to AtConnection implementations or their usage


packages/at_server_spec/lib/src/connection/inbound_connection.dart
```diff
@@ -13,6 +12,4 @@ abstract class InboundConnection extends AtConnection implements AtRateLimiter {
 
   void acceptRequests(Function(String, InboundConnection) callback,
       Function(List<int>, InboundConnection) streamCallback);
-
-  Socket? receiverSocket;
```
gkc, line 17: I removed this because it is unused (and didn't belong here anyway)

murali-shris (approved):

sitaram-kalluri (approved):

PR #1789, opened 2024-02-09
build(deps): Bump at_commons to 3.0.58

cpswan:
Dependabot still fails to raise PRs for pubspec.yaml

**- What I did**

Bump at_commons to 3.0.58

**- Description for the changelog**

build(deps): Bump at_commons to 3.0.58

PR #1791, opened 2024-02-13
build(deps): Bump at_commons to 4.0.1

cpswan:
Dependabot still fails to raise PRs for pubspec.yaml

**- What I did**

Bump at_commons to 4.0.1 per @gkc note at #1789 

build(deps): Bump at_commons to 4.0.1

### DIFF
--- a/packages/at_persistence_root_server/CHANGELOG.md
+++ b/packages/at_persistence_root_server/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 2.0.6
+- changed at_commons dependency version 
 ## 2.0.5
 - upgrade persistence spec version
 ## 2.0.4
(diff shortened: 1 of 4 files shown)


gkc:
@cpswan The dependency chain is a bit tangled here; we will need to do something like the following (I'm happy to do it):
- PR 1: Modify at_persistence_root_server pubspec to use at_commons ^4.0.0 and merge to trunk
- PR 2: Modify at_root_server pubspec to use at_commons 4.0.1 and at_server_spec 4.0.1

cpswan:
@gkc ack that we're going to need a two stage process here. For now I have the pubspec.yaml pointed to this branch, so once we merge to trunk it will need to be repointed.

gkc (approved):

PR #1855, opened 2024-03-21
fix: In enrollment notification notify only enrollmentId instead of e…

sitaram-kalluri:
**- What I did**
- When an enrollment request is submitted, the notification that is sent to the approving apps contains the entire enrollment key. For example "123..new.enrollments.__manage" where 123 is enrollment id. When an app wants to approve or deny an enrollment, the apps have to do a substring to get the enrollment id -123. 
To remove the overhead on apps, send only the enrollment id in the notification instead of sending the entire enrollment key.

**- How to verify it**
- Updated the unit and functional tests

- Send enrollment id instead of enrollment key in notification

### DIFF
--- a/packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
+++ b/packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
@@ -196,7 +196,7 @@ class EnrollVerbHandler extends AbstractVerbHandler {
       enrollmentValue.encryptedAPKAMSymmetricKey =
           enrollParams.encryptedAPKAMSymmetricKey;
       enrollmentValue.approval = EnrollApproval(EnrollmentStatus.pending.name);
-      await _storeNotification(key, enrollParams, currentAtSign);
+      await _storeNotification(newEnrollmentId, enrollParams, currentAtSign);
       responseJson['status'] = 'pending';
       enrollData = AtData()
         ..data = jsonEncode(enrollmentValue.toJson())
--- a/tests/at_functional_test/test/enroll_verb_test.dart
+++ b/tests/at_functional_test/test/enroll_verb_test.dart
@@ -318,8 +318,8 @@ void main() {
           'pixel-$randomNumber');
       expect(
           enrollListResponseMap[
-          '$enrollmentId.new.enrollments.__manage$firstAtSign']
-          ['namespace']['wavi'],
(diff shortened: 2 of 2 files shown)


gkc (changes requested):
@sitaram-kalluri 

Unless you can demonstrate that the current (client side) code will function, unchanged, correctly against this change, 
please instead add a new field to the notification rather than changing what's already there; we can do another server change in future once we have updated all client-side code

gkc:
> @sitaram-kalluri 
> 
> Unless you can demonstrate that the current (client side) code will function, unchanged, correctly against this change, 
> please instead add a new field to the notification rather than changing what's already there; we can do another server change in future once we have updated all client-side code 

Also - the notification's key has to include the full namespace; if it doesn't then all apps would receive the enrollment request notifications

Please add a field to the notification

Or better still, just change the client side library so that it provides some convenience code to get the enrollment id

sitaram-kalluri:
@gkc : The notification already contains the "enrollmentId" field, which I overlooked. I will inform @sachins-geekyants about this. Closing the PR.

Attaching the sample notification

```notification: {"id":"ef909549-499b-4355-bc27-a049c7c7082e","from":"@sitaram","to":"@sitaram","key":"2f351053-7515-4bad-b188-87aa66a210cd.new.enrollments.__manage","value":"{\"encryptedApkamSymmetricKey\":\"a/79nCji+E1UvpOvY7bGEtcMrH1RlW0KoCyDFyuESRRORaIDbBMsYESBzlCzilTlijyWA4lvocPhEsuava68VrP+d14xbNEPnQGUAslf/OE3I20I/MMsNBuDd9w/Km+ISEyR6W1ksXepREewcNpm6tcEpr+XKYlZic9kZEdc4AAn3od/uktGa5WpdioNqB5jQHJ/UHKT73vk38U8DFFOLj6/y6ot9IiN93jxy2USMwTjDu6sX8YCtPyw0rCMQ7dgYvkazEle2BjZt1UqSFULBqJMWVPluuF+QoC1F2Geh0svBwD1b2GmEOp1WfuGWTJisvaO0TO9KvEqq9pv+CWY9w==\",\"appName\":\"wavi\",\"deviceName\":\"mydevice\",\"namespace\":null}","operation":"update","epochMillis":1711009514695,"messageType":"MessageType.key","isEncrypted":false,"metadata":{"encKeyName":null,"encAlgo":null,"ivNonce":null,"skeEncKeyName":null,"skeEncAlgo":null}}```

PR #1870, opened 2024-03-28
fix: isProtectedKey() check interfering with deletion of cached keys

### LINKED ISSUE #1856: Some cached public publickey atkeys cannot be deleted (atkeys that are like:`cached:public:publickey@`)
Initial question (March 21)
`How come I am able to delete some cached public keys in my atServer (cached:public:publickey@soccer99), but not able to delete some like this one (cached:public:publickey@jeremy_0). The atSign I'm authenticated as is @soccer0`

image

Reply from Murali (March 21)
`@Jeremy Tubongbanua I think it's a bug. cached public keys are not protected keys, so deletion must be allowed.`

Expected behaviour is to be able to delete cached public keys, but some are able to and some are not able to.

Sprint points allocated to this ticket involve fixing this bug and achieving expected behaviour.

srieteja:
Closes https://github.com/atsign-foundation/at_server/issues/1856

**- What I did**
- ensure cached keys like 'cached:public:publicKey' are not considered protected keys

**- How I did it**
- isProtectedKey() in DeleteVerbHandler now accepts a named parameter '**isCached**'
- isProtectedKey check returns false if 'isCached' is true
 
**- How to verify it**
- unit tests have been added

fix: isProtectedKey() check interfering with deletion of cached keys


gkc (approved):
LGTM

gkc:
Thank you @srieteja

srieteja:
@gkc could you please merge this too. I don't have merge access

PR #1911, opened 2024-04-24
fix: otp:put requires an authenticated connection

gkc:
**- What I did**
- fix: otp:put requires an authenticated connection
- test: addded unit tests for unauthenticated otp: requests

**- How I did it**
- Moved the auth check in the otp verb handler
- Added missing unit tests.
- Modified the Otp verb in at_server_spec so it also returns true from requiresAuth (for consistency and clarity only, as the otp verb handler has its own check for authentication)
  - updated at_server_spec CHANGELOG
  - updated at_server_spec package version

**- How to verify it**
Tests pass

### DIFF
--- a/packages/at_secondary_server/CHANGELOG.md
+++ b/packages/at_secondary_server/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 3.0.44
+fix: otp authentication check
+
 ## 3.0.43
 - fix: ensure all connection writes are awaited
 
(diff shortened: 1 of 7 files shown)


cconstab (approved):

gkc:
@sitaram-kalluri @murali-shris please review as soon as is convenient. If it looks good then please can you also (1) merge this PR to trunk and (2) once the merge actions run has completed, please do a canary release c3.0.44

sitaram-kalluri (approved):

murali-shris (approved):

PR #1938, opened 2024-05-01
fix: ignore commit log cache update, if existing entry has higher commitId

### LINKED ISSUE #1291: sync_multiple_client functional test is too flaky and fails often
### Describe the bug

https://github.com/atsign-foundation/at_client_sdk/actions/runs/8644815412/attempts/1
Most of the time, this test fails during first run and passes on re-run
In this sample run, the test failed twice and passed on 3rd run

Update:
After Investigation, it was found that the flakiness was due to the test being unable to complete the initial sync. The test is attempting to complete the initial sync repeatedly until the test times out. 

### Steps to reproduce

* Raise a PR on at_client_sdk
* Or run the at_client_sdk functional test 'sync_multiple_client_test' multiple times, likely that one of the attempts fails.

### Expected behavior

* test should pass consistently

### Additional context 

The server-side behaviour was analyzed to identify the cause of sync failure. It was found that the scenario causing this issue is: 'public encryption being updated by both the isolates in the test within milliseconds of each other'. Two updates of the same key happening concurrently. Let me explain this through an example:

e.g. there are two updates of the public enc key by two different clients (call them update1 and update2). keystote.put() is called twice within milliseconds of each other, and both updates run parallelly. Assume 'update1 has created internal key 10' in commitLog/commitLogCacheMap and 'update2 has internal key 11'. Now in some cases internal key 11 is being committed first and when update2 reaches the stage of updating the commitLogCacheMap; the entry with commitId 11 is already in the cache map and update2 is overwriting that commitEntry with another commitEntry that has commitId 10.
                                                    

Now, when the latest commitEntry for public encryption key should be 11 but commitLogCache only has an entry with commitId 10. And in this case the client would repeatedly try to sync commitEntry 11, but the cache map would not have that. This caused the client to get stuck.

### Possible solution
- Add a check in the commitLogKeyStore, when updating the commitLogCache with an entry; ensure that only the latest entry stays in the cache map.

srieteja:
Closes https://github.com/atsign-foundation/at_client_sdk/issues/1291

**- What I did**
- fix: ignore commit log cache update, if the existing entry has a higher commitId

**- How I did it**
- add a check in CommitLogCache.update() that checks if there is already present for the current key, if YES does not add the current entry to the commitLogCache.

**- How to verify it**
- unit test added to validate the changes
- funtional test mentioned in https://github.com/atsign-foundation/at_client_sdk/issues/1291 should consistently pass. Which has been manually verified.

fix: ignore commit log cache update, if existing entry has higher commitId

### DIFF
--- a/packages/at_secondary_server/pubspec.yaml
+++ b/packages/at_secondary_server/pubspec.yaml
@@ -34,6 +34,13 @@ dependencies:
   yaml: 3.1.2
   logging: 1.2.0
 
+dependency_overrides:
+  at_persistence_secondary_server:
+    git:
+      url: https://github.com/atsign-foundation/at_server.git
+      path: packages/at_persistence_secondary_server
+      ref: commit_id_inconsistency_fix
+
 dev_dependencies:
   test: ^1.24.4
   coverage: ^1.6.1
--- a/packages/at_secondary_server/pubspec.yaml
+++ b/packages/at_secondary_server/pubspec.yaml
@@ -34,13 +34,6 @@ dependencies:
   yaml: 3.1.2
   logging: 1.2.0
 
-dependency_overrides:
-  at_persistence_secondary_server:
-    git:
-      url: https://github.com/atsign-foundation/at_server.git
-      path: packages/at_persistence_secondary_server
-      ref: commit_id_inconsistency_fix
-
 dev_dependencies:
   test: ^1.24.4
   coverage: ^1.6.1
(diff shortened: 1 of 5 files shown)


packages/at_secondary_server/pubspec.yaml
```diff
@@ -34,6 +34,13 @@ dependencies:
   yaml: 3.1.2
   logging: 1.2.0
 
+dependency_overrides:
```
murali-shris: can you raise a separate PR for persistence

murali-shris (changes requested):

packages/at_secondary_server/pubspec.yaml
srieteja: since most of the changes in this PR are in at_persistence, I have removed one change that belonged to at_secondary, now this PR only contains changes for at_persistence.

murali-shris (approved):

gkc (approved):
LGTM

PR #1939, opened 2024-05-02
feat: Enroll drop revoked clients

### LINKED ISSUE #1927: When an enrollment is revoked, any clients connected via that enrollment should be immediately disconnected

Without this safeguard, a revoked app could remain connected until the network connection is lost or the server is restarted etc

When an enrollment is revoked, any clients connected via that enrollment should be immediately disconnected

_No response_

### Additional context

_No response_

murali-shris:
Closes https://github.com/atsign-foundation/at_server/issues/1927
**- What I did**
- When an enrollment is revoked and if that enrollment has any active client, close the connection to the client on the server
**- How I did it**
- Implemented a method _dropRevokedClientConnection which drop connection on the server when enrollment is revoked
    - If the connection to drop is not the current connection, directly close the connection without any warning
    - if the connection to drop is current connection and force flag is set..send a message to client in the response and drop the connection after 10 seconds.
- Modified enroll verb functional tests to check for inbound connection count before and after enroll:revoke 
**- How to verify it**
- unit and functional tests should pass

### DIFF
--- a/packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
+++ b/packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
@@ -108,8 +108,8 @@ class EnrollVerbHandler extends AbstractVerbHandler {
           if (responseJson['status'] == EnrollmentStatus.revoked.name) {
             logger.finer(
                 'Dropping connection for enrollmentId: $enrollmentIdFromParams');
-            await _dropRevokedClientConnection(
-                enrollmentIdFromParams!, forceFlag, atConnection, responseJson);
+            await _dropRevokedClientConnection(enrollmentIdFromParams!,
+                forceFlag != null, atConnection, responseJson);
           }
           break;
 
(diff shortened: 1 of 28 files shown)


murali-shris:
Resolved the functional test failure for enroll:revoke:force
working on functional test for revoking enrollment and verify using stats verb whether connection got closed

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -331,6 +338,43 @@ class EnrollVerbHandler extends AbstractVerbHandler {
     responseJson['enrollmentId'] = enrollmentIdFromParams;
   }
 
+  Future<void> _dropRevokedClientConnection(
+      String enrollmentId,
+      String? forceFlag,
```
gkc: boolean instead

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -104,6 +105,12 @@ class EnrollVerbHandler extends AbstractVerbHandler {
           }
           await _handleEnrollmentPermissions(enrollVerbParams, currentAtSign,
               operation, responseJson, response);
+          if (responseJson['status'] == EnrollmentStatus.revoked.name) {
+            logger.finer(
+                'Dropping connection for enrollmentId: $enrollmentIdFromParams');
+            await _dropRevokedClientConnection(
+                enrollmentIdFromParams!, forceFlag, atConnection, responseJson);
```
gkc: Suggest pass forceFlag as a boolean i.e. (forceFlag != null)

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
murali-shris: done

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
murali-shris: done

gkc (approved):

PR #1954, opened 2024-05-15
fix: outbound listener handle error response string

murali-shris:
**- What I did**
- fixed the error response from plookup for non existent key
**- How I did it**
- When error response is returned as string for non existent key, error code and error message is not parsed from the error response string. Incorrect AT0011 internal server error was returned to the client.
Implemented a method _throwAtExceptionFromErrorResponse which takes care of handling error code and error message from both json and string error response. Instead of AT0011 (internal server error), AT0015 (key not found) error will be returned as error response to the client
- added unit test for outbound message listener
**- How to verify it**
- plookup end2end tests should pass
- unit tests should pass

### DIFF
--- a/packages/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
+++ b/packages/at_secondary_server/lib/src/connection/outbound/outbound_message_listener.dart
@@ -98,17 +98,8 @@ class OutboundMessageListener {
           // Right now, all callers of this method only expect there ever to be a 'data:' response.
           // So right now, the right thing to do here is to throw an exception.
           // We can leave the connection open since an 'error:' response indicates normal functioning on the other end
-          try {
-            result = result.toString().replaceFirst('error:', '');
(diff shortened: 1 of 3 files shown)


gkc (commented):
Please add a couple of unit tests

murali-shris:
> Please add a couple of unit tests

Sure @gkc . I have to introduce unit test for outbound message listener.

Currently facing an issue in end2end test but they are passing in my local.

error response in local run which has the fix on local server.
plookup verb response error:{"errorCode":"AT0015","errorDescription":"key not found : Exception: public:no-key@alice does not exist in keystore"}

error response in github actions ( I guess trunk version of server is used)
 plookup verb response error:{"errorCode":"AT0011","errorDescription":"Internal server exception : Request to remote secondary @cicd5 at null:null received error response 'AT0015-Exception: public:no-key@cicd5 does not exist in keystore'"}

gkc (approved):

PR #1955, opened 2024-05-15
fix: OTP default expiry duration remain constant

sitaram-kalluri:
**- What I did**
- Fix the following issue - "otpExpiryInMillis" is an instance variable and hence if the TTL is set, then the value remains for all the subsequent otp: verbs instead of the default expiry value.

**- How I did it**
- Add "_defaultOtpExpiryInMills" and mark it as final to ensure its value is not changed.
- Add the following logic - If TTL value is set, then set "otpExpiryInMills" to the value of TTL, else set to "_defaultOtpExpiryInMills".

**- How to verify it**
- Added a unit test to ensure the default expiry of the OTP value not changed for the subsequent requests of the OTP.

### DIFF
--- a/packages/at_secondary_server/lib/src/verb/handler/otp_verb_handler.dart
+++ b/packages/at_secondary_server/lib/src/verb/handler/otp_verb_handler.dart
@@ -8,6 +8,7 @@ import 'package:at_secondary/src/constants/enroll_constants.dart';
 import 'package:at_secondary/src/server/at_secondary_impl.dart';
 import 'package:at_server_spec/at_server_spec.dart';
 import 'package:at_server_spec/at_verb_spec.dart';
+import 'package:meta/meta.dart';
 import 'package:uuid/uuid.dart';
 
 import 'abstract_verb_handler.dart';
@@ -15,7 +16,8 @@ import 'abstract_verb_handler.dart';
 class OtpVerbHandler extends AbstractVerbHandler {
   static Otp otpVerb = Otp();
 
-  final int _defaultOtpExpiryInMills = Duration(minutes: 5).inMilliseconds;
+  @visibleForTesting
+  static const Duration defaultOtpExpiry = Duration(minutes: 5);
 
   OtpVerbHandler(SecondaryKeyStore keyStore) : super(keyStore);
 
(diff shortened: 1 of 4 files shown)


packages/at_secondary_server/lib/src/verb/handler/otp_verb_handler.dart
```diff
@@ -31,10 +31,11 @@ class OtpVerbHandler extends AbstractVerbHandler {
       HashMap<String, String?> verbParams,
       InboundConnection atConnection) async {
     final operation = verbParams['operation'];
-    if (verbParams[AtConstants.ttl] != null &&
-        verbParams[AtConstants.ttl]!.isNotEmpty) {
-      otpExpiryInMills = int.parse(verbParams[AtConstants.ttl]!);
-    }
+    // int.tryParse() converts the string to an integer and sets to otpExpiryInMills.
+    // If the conversion fails, "tryParse()" returns null in which case defaultOtpExpiryInMills is set to otpExpiryInMills;
```
gkc: Simplify the code comment here e.g. "Extract the ttl from the verb parameters if supplied, or use the default value if not"

packages/at_secondary_server/lib/src/verb/handler/otp_verb_handler.dart
```diff
@@ -1,21 +1,21 @@
...
 import 'abstract_verb_handler.dart';
-import 'package:at_server_spec/at_verb_spec.dart';
-import 'package:at_persistence_secondary_server/at_persistence_secondary_server.dart';
 
 class OtpVerbHandler extends AbstractVerbHandler {
   static Otp otpVerb = Otp();
 
-  @visibleForTesting
-  int otpExpiryInMills = Duration(minutes: 5).inMilliseconds;
+  final int _defaultOtpExpiryInMills = Duration(minutes: 5).inMilliseconds;
```
gkc: Change to `static const defaultOtpExpiryInMillis`

gkc: note you have a typo 'inMills' should be 'inMillis'

packages/at_secondary_server/test/otp_verb_test.dart
```diff
@@ -149,16 +148,33 @@ void main() {
...
+      await otpVerbHandler.processVerb(response, verbParams, inboundConnection);
+      AtData? atData =
+          await secondaryKeyStore.get('private:${response.data}$atSign');
+      expect(atData?.metaData?.ttl, 1);
+
+      verbParams = getVerbParam(VerbSyntax.otp, 'otp:get');
+      inboundConnection.metaData.isAuthenticated = true;
+      await otpVerbHandler.processVerb(response, verbParams, inboundConnection);
+      atData = await secondaryKeyStore.get('private:${response.data}$atSign');
+      expect(atData?.metaData?.ttl, 300000);
```
gkc: Instead of the hard-coded 300000 use `OtpVerbHandler. defaultOtpExpiryInMillis`

gkc (changes requested):
Have requested some changes; please also update CHANGELOG and pubspec

gkc (approved):

Issue #1944, opened 2024-05-08
Setting `isEncrypted:false` in notify: verb doesn't actually set it to `false`.


JeremyTubongbanua:
![image](https://github.com/atsign-foundation/at_server/assets/79019866/3a5854da-82e0-4655-80ad-54168b5625ca)

In the above image, I sent notification with `isEncrypted:false`

Then in the receiving notification has `isEncrypted:true`

![image](https://github.com/atsign-foundation/at_server/assets/79019866/c3cd89c8-fc76-460d-97e0-99e65691884e)

Intended behaviour:

Running `notify:isEncrypted:false:<...>`

Should have a receiving notification of:

`notification:{..., "isEncrypted":false, ...}`

xavierchanth:
- P1 to look at it, then reevaluate the priority of the underlying issue.

srieteja:
Could not get to this in the previous sprint, will be taking this up first in PR88.

gkc:
Hi @srieteja I've been looking at this as it came up in another context. I'll pick it up this sprint.

Summary of the current behaviour for 'update' notifications
- at_secondary_server `NotifyVerbHandler._getIsEncrypted` ignores whatever it has been sent and always sets isEncrypted to 'true'
- at_client `NotificationServiceImpl.notify` always encrypts the value but does not set isEncrypted in the metadata. Thus, usually (unless app code explicitly sets metadata.isEncrypted to true), metadata.isEncrypted is apparently false (default value) but the value has in fact been encrypted
- at_commons NotifyVerbBuilder.buildCommand only includes isEncrypted in the generated command if metadata.isEncrypted is true (which it ~never is, right now)
- at_secondary_server ResourceManager.prepareNotifyCommandBody only includes isEncrypted in the generated command if isEncrypted is true


Summary of the changes I have in mind:
- at_secondary_server: these are the most important changes. Once implemented, the problem Jeremy encountered will be resolved
  - NotifyVerbHandler._getIsEncrypted
    - if a value for isEncrypted was set in the received command, respect it, otherwise default it to 'true' (current behaviour)
  - ResourceManager.prepareNotifyCommandBody
    - always include isEncrypted in the generated command
- at_client
  - NotificationServiceImpl.notify
    - add an optional bool parameter `encryptValue` default it to true
    - set the metadata.isEncrypted to the value of `encryptValue`
  - NotificationRequestTransformer.transform
[...]

PR #1983, opened 2024-06-10
fix: respect isEncrypted:false if supplied in notify: command, and ensure the correct value is always transmitted onwards

gkc:
**- What I did**
fix: respect isEncrypted:false if supplied in notify: command, and ensure the correct value is always transmitted onwards

Addresses at_server parts of #1944 

**- How I did it**
See commits

**- How to verify it**
- [x] at_server automated tests pass
- [x] manually verified that noports e2e tests pass when running against this branch's version of atServer
- [x] manually verified that at_client_sdk e2e tests pass when running against this branch's version of atServer
image


packages/at_secondary_server/lib/src/notification/resource_manager.dart
```diff
@@ -235,10 +235,11 @@ class ResourceManager {
...
       }
-      if (atNotification.atMetadata!.isEncrypted != null &&
-          atNotification.atMetadata!.isEncrypted == true) {
-        commandBody = '${AtConstants.isEncrypted}:true:$commandBody';
-      }
+
+      String? isEncryptedStr =
+          (atNotification.atMetadata!.isEncrypted ?? false) ? 'true' : 'false';
+      commandBody = '${AtConstants.isEncrypted}:$isEncryptedStr:$commandBody';
+
```
gkc, line 242: The above change is to ensure that the isEncrypted value in the notification's metadata is always passed on to the next atServer. Previously it was only passed on if the value was `true`

gkc:
Converted to draft as I need to add some unit tests to ensure this doesn't regress in the future

gkc:
> Converted to draft as I need to add some unit tests to ensure this doesn't regress in the future

@sitaram-kalluri Have added some unit tests. Please review when you have a moment, thank you.

sitaram-kalluri (approved):

PR #1998, opened 2024-06-15
chore: New certificates generated

atsignbot:
Fresh certificates generated.


cpswan:
Not needed, as we did certs a couple of days ago

PR #2020, opened 2024-07-10
feat: add enrollment details to public hidden key

murali-shris:
**- What I did**
- added enrollment details to public hidden key which can be used in enrollment widget to fetch list of apps that are eligible to approve enrollment request
**- How I did it**
- on enrollment approval,  add enrollment info (enrollmentid, app name,device name, namespaces) to a public key public:_enrollments<atsign>
- on enrollment revoke, remove the enrollment info from the public hidden key
**- How to verify it**
- added checks to existing unit tests

### DIFF
--- a/packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
+++ b/packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
@@ -242,6 +242,9 @@ class EnrollVerbHandler extends AbstractVerbHandler {
       // Store default encryption private key and self encryption key(both encrypted)
       // for future retrieval
       await _storeEncryptionKeys(newEnrollmentId, enrollParams, currentAtSign);
+      // Store the enrollment in a hidden for the user to retrieve active enrollments from client side when required
+      await _storeEnrollmentInPublicHiddenKey(
+          newEnrollmentId, enrollmentValue, currentAtSign);
       // store this apkam as default pkam public key for old clients
       // The keys with AT_PKAM_PUBLIC_KEY does not sync to client.
       await keyStore.put(AtConstants.atPkamPublicKey,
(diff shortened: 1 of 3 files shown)


packages/at_secondary_server/test/enroll_verb_test.dart
```diff
@@ -847,6 +867,14 @@ void main() {
           response, approveEnrollVerbParams, inboundConnection);
       expect(jsonDecode(response.data!)['enrollmentId'], enrollmentId);
       expect(jsonDecode(response.data!)['status'], 'approved');
+      // verify enrollmentId is present in hidden public key before enroll:revoke
```
gkc, line 870: This should be made into its own unit test rather than added to an existing one

packages/at_secondary_server/test/enroll_verb_test.dart
```diff
@@ -856,6 +884,12 @@ void main() {
           response, enrollVerbParams, inboundConnection);
       expect(jsonDecode(response.data!)['enrollmentId'], enrollmentId);
       expect(jsonDecode(response.data!)['status'], 'revoked');
+      // verify enrollmentId is removed from hidden public key after enroll:revoke
+      enrollmentAtData = await secondaryKeyStore
```
gkc, line 888: This should be made into its own unit test rather than added to an existing one

packages/at_secondary_server/test/enroll_verb_test.dart
```diff
@@ -744,6 +755,15 @@ void main() {
       enrollmentData = await secondaryKeyStore.get(enrollmentKey);
       expect(enrollmentData!.metaData!.expiresAt, null);
       expect(enrollmentData.metaData!.ttl, 0);
+      var enrollmentAtData = await secondaryKeyStore
+          .get('public:$enrollmentPublicHiddenKey$alice');
```
gkc, line 759: This should be made into its own unit test rather than added to an existing one

gkc:
Have added @cconstab and @XavierChanth as reviewers as I'd like this change to have more eyes on it before we merge it

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -268,6 +271,33 @@ class EnrollVerbHandler extends AbstractVerbHandler {
         'private:${enrollParams.otp?.toLowerCase()}${AtSecondaryServerImpl.getInstance().currentAtSign}');
   }
 
+  // Store enrollment info after approval in a public hidden key
+  Future<void> _storeEnrollmentInPublicHiddenKey(
+      String newEnrollmentId,
+      EnrollDataStoreValue enrollmentDataStoreValue,
+      String currentAtSign) async {
+    var key = 'public:$enrollmentPublicHiddenKey';
```
xavierchanth, line 279: Wait does this actually work as intended? A public key with `_` is also hidden? I was under the impression that this was only for self-keys, but it makes sense that it works for all key types.

If so, the changes I proposed in https://github.com/atsign-foundation/at_protocol/pull/166 are not quite right.

No action items for this PR intended out of this comment, but rather clarification for the changes I am making to the protocol spec.

xavierchanth (approved):

gkc (changes requested):
As discussed let's park this while we revisit the problem we're trying to solve and come up with the desired user experience first. As it stands, this would leak information which a user might not at all want

PR #2076, opened 2024-08-27
fix: remove UPDATE_ALL commitEntry of the key when removing expired keys

### LINKED ISSUE #2071: When skipCommitsForExpiredKeys is enabled, original commit entries for each of the keys might be getting left behind
### Describe the bug

skipCommitsForExpiredKeys has been implemented (https://github.com/atsign-foundation/at_server/issues/1660), which skips writing commit entries when deleting expired keys. When enabled, the original commit entry created while creating the key might be left behind.

Previously, when an expired key was deleted the latest commitEntry for that key would be updated to DELETE which is valid. But with this change, the latest commitEntry would indicate CommitOp.UPDATE but the key actually has been deleted.

### Expected behaviour

The commitLog should contain accurate and up-to-date details for the status of each key in the Keystore

### Possible solutions
- Delete all commitEntries for an expired key when it's being removed from the Keystore.

### Additional context
- Although https://github.com/atsign-foundation/at_server/issues/1660 has been implemented, this has NOT been enabled yet.
- Also, this only affects the deletion of expired keys. Manual deletion of keys using the delete verb is unaffected.

srieteja:
Closes https://github.com/atsign-foundation/at_server/issues/2071

**- What I did**
[fix: delete UPDATE commit entry when removing an expired key](https://github.com/atsign-foundation/at_server/commit/547c54fdafb5a4f207f66473443712008d80d23f)

**- How I did it**
- when removing an expired key from the Keystore and optimized commits is set to true, remove any commit entry for that corresponding key from the commit log

**- How to verify it**
- Existing unit tests have been updated to match current changes. All tests passing should indicate functioning as expected

fix: remove UPDATE_ALL commitEntry of the key when removing expired keys


packages/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
```diff
@@ -328,6 +328,11 @@ class HiveKeystore implements SecondaryKeyStore<String, AtData?, AtMetaData?> {
       // On deleting the key, remove it from the expiryKeyCache.
       _expiryKeysCache.remove(key);
       if (skipCommit) {
+        logger.finer('Skipping commit log entry for expired key deletion: $key');
```
VJag: So when a key expires remove() must be getting called which deletes latest commit entry if skipCommit is enabled.

sitaram-kalluri (changes requested):
Please add a unit test for the following scenario:
- Insert 3 distinct keys with Commit.Op. Update.
- Delete one of the key
- Then call, "getEntries" method and verify the list of the commit entries returned and their commit-ids

packages/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
```diff
@@ -328,6 +328,10 @@ class HiveKeystore implements SecondaryKeyStore<String, AtData?, AtMetaData?> {
       // On deleting the key, remove it from the expiryKeyCache.
       _expiryKeysCache.remove(key);
       if (skipCommit) {
+        CommitEntry? commitEntry = _commitLog.getLatestCommitEntry(key);
```
VJag, line 333: Please add docs on why we are deleting a key from commit log

VJag (dismissed):
Made a minor comment, otherwise LGTM.

packages/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
srieteja: Updated in [531e0e1](https://github.com/atsign-foundation/at_server/pull/2076/commits/531e0e16eb5ee3488ca44e65308baf1baf9d3df1)[531e0e1](https://github.com/atsign-foundation/at_server/pull/2076/commits/531e0e16eb5ee3488ca44e65308baf1baf9d3df1)

packages/at_persistence_secondary_server/lib/src/keystore/hive_keystore.dart
srieteja: Updated in [531e0e1](https://github.com/atsign-foundation/at_server/pull/2076/commits/531e0e16eb5ee3488ca44e65308baf1baf9d3df1)[531e0e1](https://github.com/atsign-foundation/at_server/pull/2076/commits/531e0e16eb5ee3488ca44e65308baf1baf9d3df1)

VJag (approved):

sitaram-kalluri (approved):

PR #2085, opened 2024-09-09
fix: Introduce time duration for apkam keys to auto expire

### LINKED ISSUE #2074: Introducing Auto-Expiry and Time-to-Birth Features for APKAM Keys

Currently, the APKAM keys do not have an auto-expiry feature. Enhance the APKAM keys by allowing users to set an auto-expiry, after which the keys will no longer be valid for authentication.

Additionally, introduce a time-to-birth mechanism, enabling users to specify when the atKeys will become active, alongside the existing time-to-live functionality.

- In the enrollment request, allow user to specify "TTL" and "TTB" which represents the time duration in minutes. When TTL represents "time to live", beyond which the APKAM keys will be inactive for usage.
- When TTL is met, the APKAM keys cannot be used for authentication and subsequently the enrollment data will be deleted from the keystore.

sitaram-kalluri:
**- What I did**
- Introduce APKAM keys to auto expire after specified time duration. 

**- How I did it**
- In at_commons, introduced `Duration apkamKeysExpiryDuration` to capture the time duration set by the user.  This time duration is propagated to the secondary server.
- In the secondary server, once enrollment is approved, set the TTL value to the `Duration apkamKeysExpiryDuration` in milliseconds. This will ensure that the key expires and is subsequently deleted from the keystore.
- In "AbstractVerbHandler.dart", add "_verifyIfEnrollmentIsActive", which will check if the enrollment is active on the existing connections. If the enrollment expires, closes the connection with error message.
- For the new connections, in the "PkamVerbHandler.dart->handleApkamVerification", there is call to "getEnrollDataStoreValue" which will check if the enrollment is active. If not active, sets enrollment state to expired and authentication fails.
- Introduce EnrollmentManager class which contains "get", "put" and "buildEnrollmentKey" methods which will return the enrollment details from the keystore, create/update enrollment details to the keystore and build the enrollment key for the enrollment id respectively.
- Refactor the code in the verb handler which related to fetch / creating enrollment records with happen via the enrollment manager.
- Update the unit tests to align with the EnrollmentManager class.

**- How to verify it**
- Added a unit test to verify the changes.
  - A test to verify apkam expiry is set for approved enrollment
  - A test to verify pkam verb fails when apkam keys are expired
  - A test to verify update verb fails when apkam keys are expired
  - A test to verify delete verb fails when apkam keys are expired

- Added below functional tests:
  -  A test to verify apkam authentication fails with expired apkam keys
  - A test to verify connection closes when apkam keys expire

- fix: Introduce time duration for apkam keys to auto expire


gkc (changes requested):
A few things missing here:
1. Enforcement to prevent a client from authenticating with an enrollment that has expired
2. Test(s) to verify the same
3. Enforcement in the verb handler(s) to terminate connections whose enrollment has expired
4. Test(s) to verify the same

sitaram-kalluri:
> A few things missing here:
> 
>     1. Enforcement to prevent a client from authenticating with an enrollment that has expired
> 
>     2. Test(s) to verify the same
> 
>     3. Enforcement in the verb handler(s) to terminate connections whose enrollment has expired
> 
>     4. Test(s) to verify the same

@gkc : Addressed the review comments and updated the PR description. Please review.

gkc (commented):
Looks good. However, we need to be smarter in terms of dealing with enrollment records - checking every time is necessary, but means an extra fetch from the data store every time. We should create an EnrollmentManager class through which all interaction with enrollment records happens, and which can therefore make appropriate decisions regarding in-memory caching of enrollment records. Can you check the effort involved in doing that - if it's not huge then we should make those changes in this PR also.

sitaram-kalluri:
> Looks good. However, we need to be smarter in terms of dealing with enrollment records - checking every time is necessary, but means an extra fetch from the data store every time. We should create an EnrollmentManager class through which all interaction with enrollment records happens, and which can therefore make appropriate decisions regarding in-memory caching of enrollment records. Can you check the effort involved in doing that - if it's not huge then we should make those changes in this PR also.

@gkc : Introduce the "EnrollmentManager" class through which all interactions with enrollment records happens. Updated the PR description. Please review.

gkc (commented):
@sitaram-kalluri looks good! can you resolve the conflicts please?

gkc (approved):
LGTM

PR #2117, opened 2024-10-09
fix: Update the error message and add skip commit for enrollment key deletion

sitaram-kalluri:
**- What I did**
- In enroll_verb_handler.dart, modify the error message to specify that enrollment keys can only be deleted for enrollments that are denied or revoked.
- Add "remove" method to the enrollment_manager.dart class which will delete the enrollment key and set "skipCommit" to true to prevent the entry getting logged into the commit log and to prevent enrollment data getting sync'ed between the client and server, therefore skip it.
- Since client does not need to the apkamPublicKey to be sync'ed, set skipCommit to true when storing them into the keystore and update the tests accordingly.
- Use the "remove" method from enrollment_manager.dart in enroll_verb_handler.dart to delete the enrollment key.
- Update the unit test accordingly to match the error message.

**- How to verify it**
- Added the following unit tests to verify the commit log state during various enrollment operations:
  - A test to verify commit log state during create approve revoke and delete an enrollment request
  - A test to verify commit log state during create deny and delete an enrollment request

### DIFF
--- a/packages/at_secondary_server/test/enroll_verb_test.dart
+++ b/packages/at_secondary_server/test/enroll_verb_test.dart
@@ -6,6 +6,7 @@ import 'package:at_persistence_secondary_server/at_persistence_secondary_server.
 import 'package:at_secondary/src/connection/inbound/inbound_connection_metadata.dart';
 import 'package:at_secondary/src/constants/enroll_constants.dart';
 import 'package:at_secondary/src/enroll/enroll_datastore_value.dart';
+import 'package:at_secondary/src/enroll/enrollment_manager.dart';
 import 'package:at_secondary/src/server/at_secondary_config.dart';
 import 'package:at_secondary/src/utils/handler_util.dart';
 import 'package:at_secondary/src/verb/handler/delete_verb_handler.dart';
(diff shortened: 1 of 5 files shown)


gkc:
1. Is there a commit skip when enrollments are added?

2. In the PR description you say enrollment info "is synced", but I think you mean "is not synced"

sitaram-kalluri:
> 1. Is there a commit skip when enrollments are added?
> 
> 2. In the PR description you say enrollment info "is synced", but I think you mean "is not synced"

@gkc : Yes, when we add the enrollments we have a skipCommit set to true. Attaching the code snippet below from the "EnrollmentManager.dart" . As for point # 2, Correct, I meant that there should be no syncing between the client and server

```dart
Future<void> put(String enrollmentId, AtData atData) async {
    String enrollmentKey = buildEnrollmentKey(enrollmentId);
    await _keyStore.put(enrollmentKey, atData, skipCommit: true);
  }
```

gkc (changes requested):
Thanks @sitaram-kalluri 

Please add unit test(s) which check commit log state after the various enrollment operations

sitaram-kalluri:
> Thanks @sitaram-kalluri
> 
> Please add unit test(s) which check commit log state after the various enrollment operations

Sure, @gkc

packages/at_secondary_server/test/enroll_verb_test.dart
```diff
@@ -2040,4 +2041,236 @@ void main() {
...
+
+      atCommitLog =
+          await AtCommitLogManagerImpl.getInstance().getCommitLog(alice);
+      itr = atCommitLog?.getEntries(-1);
+      while (itr!.moveNext()) {
+        // When approving an enrollment, stores the public key with
+        // public:appName.deviceName.pkam.__pkams.__public_keys@atSign key. Therefore,
+        // commit log has an entry.
+        expect(
+            itr.current.key.contains('pkam.__pkams.__public_keys$alice'), true);
```
gkc: please do a more precise check including the appName and deviceName being used

packages/at_secondary_server/test/enroll_verb_test.dart
```diff
@@ -2040,4 +2041,236 @@ void main() {
...
+      atCommitLog =
+          await AtCommitLogManagerImpl.getInstance().getCommitLog(alice);
+      itr = atCommitLog?.getEntries(-1);
+      // Since there are no entries in commit log, iterator.moveNext() returns false.
+      while (itr!.moveNext()) {
+        // When approving an enrollment, stores the public key with
+        // public:appName.deviceName.pkam.__pkams.__public_keys@atSign key. Therefore,
+        // commit log has an entry.
+        expect(
+            itr.current.key.contains('pkam.__pkams.__public_keys$alice'), true);
```
gkc: please do a more precise check including the appName and deviceName being used

packages/at_secondary_server/test/enroll_verb_test.dart
```diff
@@ -2040,4 +2041,236 @@ void main() {
...
```
gkc: please do a more precise check including the appName and deviceName being used

packages/at_secondary_server/test/enroll_verb_test.dart
```diff
@@ -2040,4 +2041,236 @@ void main() {
...
```
gkc: please do a more precise check including the appName and deviceName being used

packages/at_secondary_server/test/enroll_verb_test.dart
```diff
@@ -2040,4 +2041,236 @@ void main() {
...
+
+      atCommitLog =
+          await AtCommitLogManagerImpl.getInstance().getCommitLog(alice);
+      itr = atCommitLog?.getEntries(-1);
+      // Since there are no entries in commit log, iterator.moveNext() returns false.
+      while (itr!.moveNext()) {
+        // When approving an enrollment, stores the public key with
+        // public:appName.deviceName.pkam.__pkams.__public_keys@atSign key. Therefore,
+        // commit log has an entry.
+        expect(
```
gkc: please do a more precise check including the appName and deviceName being used

packages/at_secondary_server/test/enroll_verb_test.dart
```diff
@@ -2214,31 +2214,23 @@ void main() {
...
-      while (itr!.moveNext()) {
-        // When approving an enrollment, stores the public key with
-        // public:appName.deviceName.pkam.__pkams.__public_keys@atSign key. Therefore,
-        // commit log has an entry.
-        expect(
-            itr.current.key.contains('pkam.__pkams.__public_keys$alice'), true);
-      }
-      // Ensure there are no other keys in the commit log.
-      expect(itr.moveNext(), false);
+      expect(itr!.moveNext(), false);
```
gkc, line 2219: The EnrollVerbHandler is not requesting to skip the commit when storing this key. So in theory this test is required - but I believe the commit log is not currently being updated because of this https://github.com/atsign-foundation/at_server/pull/2115

So - we should probably explicitly change the EnrollVerbHandler so that it skips the commit log when storing this public key also, since these are not things that need to be synced to clients and, indeed, currently are not synced - but would start syncing if #2115 were merged

gkc (changes requested):

packages/at_secondary_server/test/enroll_verb_test.dart
gkc: Specifically add `skipCommit: true` to the calls to keyStore.put which are being made on lines 257 and 362 in enroll_verb_handler.dart

packages/at_secondary_server/test/enroll_verb_test.dart
sitaram-kalluri: Added validations for the appName, deviceName and namespaces.

packages/at_secondary_server/test/enroll_verb_test.dart
sitaram-kalluri: Added validations for the appName, deviceName and namespaces.

packages/at_secondary_server/test/enroll_verb_test.dart
sitaram-kalluri: - Set skipCommit to true and modified unit and functional tests accordingly. 
- Also did a round of manual testing with at_onboarding_cli to approve the enrollment request and use the new apkam keys authenticate the connection.

tests/at_functional_test/test/sync_verb_test.dart
```diff
@@ -121,7 +121,7 @@ void main() {
       expect(
           int.parse(
               jsonDecode(statsResponse.replaceAll('data:', ''))[0]['value']),
-          lastCommitIdBeforeUpdate + 4);
+          lastCommitIdBeforeUpdate + 3);
```
sitaram-kalluri, line 124: Since akamPublicKeys are not logged into the commit log, reduced value to 3.

gkc (approved):

PR #2131, opened 2024-10-17
feat: Sync skip deletes until

murali-shris:
**- What I did**
- added ability to skip delete commit entries in sync response. If skipDeletes flag is set to true, delete commits will not be returned in sync response from server to client.
**- How I did it**
- introduced skipDeletes flag in sync:from syntax
- added optional flag skipDeletes=false in commitLog getEntries. If set to true, delete commits will be skipped. if not passed, then existing behaviour applies. Refactored the code to introduce a strategy pattern to fetch keys.
- added unit tests
**- How to verify it**
- tests should pass

persistence change in this PR:
https://github.com/atsign-foundation/at_server/pull/2166


packages/at_persistence_secondary_server/lib/src/log/commitlog/at_commit_log.dart
```diff
@@ -127,12 +127,13 @@ class AtCommitLog extends BaseAtCommitLog {
   /// Returns the Iterator of [_commitLogCacheMap] from the commitId specified.
   @server
   Iterator<MapEntry<String, CommitEntry>> getEntries(int commitId,
-      {String? regex, int limit = 25}) {
+      {String? regex, int limit = 25, bool skipDeletes = false}) {
```
VJag: I think we can make this a strategy.. where FetchAllKeysStrategy is to get all and SkipDeletesStrategy is to skip deletes, instead of if-else code might look more readable.

packages/at_secondary_server/lib/src/verb/handler/sync_progressive_verb_handler.dart
```diff
@@ -43,7 +43,8 @@ class SyncProgressiveVerbHandler extends AbstractVerbHandler {
     // Get entries to sync
     var commitEntryIterator = atCommitLog!.getEntries(
         int.parse(verbParams[AtConstants.fromCommitSequence]!) + 1,
-        regex: verbParams['regex']);
+        regex: verbParams['regex'],
+        skipDeletes: verbParams['skipDeletes'] == 'true');
```
VJag: VerbSyntax also needs this parameter, otherwise regex validation would fail.

murali-shris: yes. those changes are in at_commons
https://github.com/atsign-foundation/at_libraries/pull/696/files

packages/at_persistence_secondary_server/lib/src/log/commitlog/sync/skip_deletes_strategy.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:at_persistence_secondary_server/src/log/commitlog/commit_entry.dart';
+import 'package:at_persistence_secondary_server/src/log/commitlog/sync/sync_keys_fetch_strategy.dart';
+
+class SkipDeleteStrategy extends SyncKeysFetchStrategy {
+  @override
+  bool shouldIncludeEntryInSyncResponse(
+      CommitEntry commitEntry, int commitId, String regex,
+      {List<String>? enrolledNamespace}) {
```
VJag: Dont we need call to super.shouldIncludeEntryInSyncResponse()?

murali-shris: i kept the method in base class abstract

packages/at_persistence_secondary_server/lib/src/log/commitlog/sync/sync_keys_fetch_strategy.dart
```diff
@@ -0,0 +1,59 @@
+import 'package:at_commons/at_commons.dart';
+import 'package:at_persistence_secondary_server/at_persistence_secondary_server.dart';
+import 'package:at_utils/at_utils.dart';
+
+abstract class SyncKeysFetchStrategy {
```
VJag: Should we call it NamespaceHonoringKeyFetchStrategy?

murali-shris: Honoring namespace is one of the checks done. regex match and skipping delete commits are the other two checks. In the future if we have any other checks not related to namespace we may have to rename this class

packages/at_persistence_secondary_server/test/commit_log_test.dart
```diff
@@ -840,6 +840,33 @@ void main() async {
         expect(commitEntriesMap.containsKey('public:phone.wavi@alice'), false);
         expect(commitEntriesMap.containsKey('public:location@alice'), true);
       });
```
gkc: Please add another test which both uses a regex and sets skipDeletes: true

packages/at_secondary_server/test/sync_unit_test.dart
```diff
@@ -469,6 +469,39 @@ void main() {
         expect(syncResponse[3]['operation'], '*');
       });
```
gkc, line 471: Please add another test which both uses a regex and sets skipDeletes: true

gkc (changes requested):
LGTM but have suggested adding couple more tests

packages/at_persistence_secondary_server/test/commit_log_test.dart
murali-shris: Done

packages/at_secondary_server/test/sync_unit_test.dart
murali-shris: Done

sitaram-kalluri (approved):

murali-shris:
@gkc I have published persistence secondary and this PR is ready to be merged

gkc (approved):

PR #2170, opened 2024-12-05
fix: apkam - add IV to self encryption key and encryption private key

murali-shris:
**- What I did**
- changes to store IV along with self encryption and encryption private key data in keystore for apkam.
**- How I did it**
- in enroll verb handler, add IV to keystore data if IV is sent in enroll:approve params
- added check in functional test to verify whether IV is sent in keys:get response
- in enroll verb handler, removed storing of encryption private key and self encryption key for first enrollment. Removed corresponding check in functional test

**- How to verify it**
- tests should pass


gkc (dismissed):
Need two IVs

gkc (commented):
Sorry I misread the code. Looks good but I'd like to talk through both this PR and the other ones tomorrow before we merge

gkc (commented):
.

gkc (approved):

PR #2193, opened 2025-01-09
ci: Update workflow to use uv to install python deps

cpswan:
As Actions moves `ubuntu-latest` to `ubuntu-24.04` we need to have a workflow that doesn't try to `pip install` into the default namespace, as that's now protected.

**- What I did**

Rather than having a statically generated `requirements.txt` it's now created from `poetry.lock` and then installed into a `uv` created venv using `uv`.

**- How I did it**

Based on https://github.com/atsign-foundation/at_protocol/pull/217

**- How to verify it**

 - [x] Manually run the Refreshcerts workflow

ci: Use uv based workflow


cpswan:
@kumarnarendra701 I've added you as a reviewer so that you're aware of what's going on here.

https://blog.thestateofme.com/2024/10/03/python-script-dependencies/ provides additional background.

NB we made the call to use `uv` in GitHub Actions workflows despite it not yet being at 1.0 as they're ephemeral rather than becoming part of a deployed production environment.

cpswan:
[Test run of the modified action](https://github.com/atsign-foundation/at_server/actions/runs/12688446352) failed:

```log
error: In `--require-hashes` mode, all requirements must be pinned upfront with `==`, but found: `urllib3`
```

Making this draft whilst I find the problem/solution

cpswan:
Clean [workflow run](https://github.com/atsign-foundation/at_server/actions/runs/12688916847) after lifting minimal Python to 3.9

xavierchanth (approved):

PR #2218, opened 2025-02-12
feat: dynamicvirtualenv image

xavierchanth:
**- What I did**

Why: I need hundreds of atServers that can be setup portably for a demo.

Created a version of virtualenv called dynamicvirtualenv that allows you to mount a config file for the atServers.

The file needs to be mounted at `/mnt/setup/atservers`.
example file format:

```
@alice 25000 <cramkey>
@bob   25001 <cramkey>
```

The other main difference to this image is that I've removed the pkamLoad service and the install_PKAM_keys binary since that depends on demo data. All of the pkam keys must be manually setup, but can be easily automated in a simple script (which I plan to develop as part of my demo first).

**- How I did it**

**- How to verify it**

feat: dynamicvirtualenv image


.github/workflows/refreshcerts.yaml
```diff
@@ -48,11 +48,19 @@ jobs:
...
           cp fullchain.pem ../../tools/build_virtual_environment/ve_base/contents/atsign/root/certs/fullchain.pem
           cp ../../tools/build_virtual_environment/ve_base/contents/atsign/root/certs/*.pem \
             ../../tools/build_virtual_environment/ve_base/contents/atsign/secondary/base/certs/
+          # dynamic ve
+          cp cert.pem ../../tools/build_virtual_environment/dynamic_ve_base/contents/atsign/root/certs/cert.pem
+          cp privkey.pem ../../tools/build_virtual_environment/dynamic_ve_base/contents/atsign/root/certs/privkey.pem
+          cp fullchain.pem ../../tools/build_virtual_environment/dynamic_ve_base/contents/atsign/root/certs/fullchain.pem
+          cp ../../tools/build_virtual_environment/dynamic_ve_base/contents/atsign/root/certs/*.pem \
+            ../../tools/build_virtual_environment/dynamic_ve_base/contents/atsign/secondary/base/certs/
+
```
xavierchanth, line 63: Also add vip.ve.atsign.zone certs to dynamic_ve image

.github/workflows/ve_base.yaml
```diff
@@ -2,14 +2,21 @@ name: VE_base
...
 jobs:
   ve_base:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - dir: "ve_base"
+            name: "vebase"
+          - dir: "dynamic_ve_base"
+            name: "dynamicvebase"
```
xavierchanth, line 19: Using a matrix for ve_base on scheduled workflow

.github/workflows/vip_dynamic_rebuild.yaml
```diff
@@ -0,0 +1,108 @@
+name: dynamic_vip_rebuild
```
xavierchanth, line 1: separate dynamic vip build because this one is workflow dispatch, so we don't want it to be a matrix with vip_rebuild.

tools/build_virtual_environment/dynamic_ve_base/contents/tmp/setup/create_demo_accounts.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+cd /tmp/setup
+
+config_file="/mnt/setup/atservers"
+
+while read -r line; do
+  echo "line: $line"
+  ./createConf $line # intentional word splitting
+done <"$config_file"
```
xavierchanth: instead of making several inline calls to ./createConf, we read the mounted config file and call createConf for each one.

xavierchanth:
Other than the removed files that are specific to the demo data, if there isn't a comment on the file changes tab, then I've not changed anything (it's a direct copy of the original ve). The only exception is an update to some comments in the Dockerfiles in https://github.com/atsign-foundation/at_server/pull/2218/commits/ea51a93a5d21efd1feacf6c832b46e4908a5e594

Edit to this comment, see: https://github.com/atsign-foundation/at_server/pull/2218#issuecomment-2654990539

xavierchanth:
Setting to draft, needs more work

xavierchanth:
More changes introduced [97c8e62](https://github.com/atsign-foundation/at_server/pull/2218/commits/97c8e62e009b88e3596d7fa6db4088e2a45b664f)

To fix the lifecycle of setting up the atServers, which can no longer be done during build time, but must be done before starting supervisord.

tools/build_virtual_environment/dynamic_ve/Dockerfile.canary_to_vip
```diff
@@ -0,0 +1,11 @@
+FROM atsigncompany/dynamicvirtualenv:canary
```
cpswan, line 1: Do we expect to have canaries for this, or has this just been copied over?

tools/build_virtual_environment/dynamic_ve/Dockerfile.vip
```diff
@@ -0,0 +1,32 @@
+FROM dart:3.6.1@sha256:a071f0322c7c80469842f7c8258de51d1946ba9b79d09bb8fc4dc969487a68ca AS buildimage
```
cpswan, line 1: Need to update hash, but also need to add this Dockerfile to Dependabot config

tools/build_virtual_environment/dynamic_ve_base/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+## sudo docker build -t atsigncompany/dynamicvebase \
+## -f tools/build_virtual_environment/dynamic_ve_base/Dockerfile .
+COPY . .
+RUN \
+  cd /app/packages/at_root_server ; \
+  dart pub get ; \
+  dart pub update ; \
+  dart compile exe bin/main.dart -o root ;
+# Disabled for dynamic_ve_base
+# cd /app/tools/build_virtual_environment/install_PKAM_Keys ; \
```
cpswan, line 16: Could we avoid a lot of duplication by having a Dockerfile like this as the new VE base (for both variants) and then have another base image that we build on top of that with this stuff as an extra layer?

cpswan (changes requested):

tools/build_virtual_environment/dynamic_ve/Dockerfile.canary_to_vip
xavierchanth: copied over, I can remove it

tools/build_virtual_environment/dynamic_ve_base/Dockerfile
xavierchanth: Sure, I can do that. I've put this PR as an item in the next sprint for me to make these changes. For now, this is enough to continue onwards with the demo I'm working on.

PR #2304, opened 2025-06-09
test: increase timeouts on some tests

gkc:
**- What I did**
- test: increased timeout on a notify e2e test from 1s to 5s, to be more resilient to highly loaded staging environment.
  - also reworked the test to remove a bug in the test itself
- test: increased delay on an enrollment denial unit test from 1ms to 2ms, to remove possibility of test failing because the enrollment (with an expiry of 1ms) has not yet expired
- chore: updated lints dependency in e2e tests, cleaned up lint

**- How I did it**
See commits

**- How to verify it**
Tests pass

### DIFF
--- a/tests/at_end2end_test/test/at_demo_data.dart
+++ b/tests/at_end2end_test/test/at_demo_data.dart
@@ -1,5 +1,5 @@
-///This is a dummy file which gets overwritten by the github actions with the
-/// atsign's keys map.
+// This is a dummy file which gets overwritten by the github actions with the
+//  atsign's keys map.
 
 /// The dummy pkamPrivateKeyMap which holds the pkam private keys of the atSign's. The values
 /// are populated by the github actions
--- a/tests/at_end2end_test/test/notify_verb_test.dart
+++ b/tests/at_end2end_test/test/notify_verb_test.dart
@@ -2,10 +2,13 @@ import 'dart:convert';
 import 'dart:math';
 
 import 'package:at_commons/at_commons.dart';
+import 'package:at_utils/at_logger.dart';
 import 'package:test/test.dart';
 import 'package:version/version.dart';
 
 import 'e2e_test_utils.dart' as e2e;
+
+AtSignLogger logger = AtSignLogger(' notify_verb_test ');
 
 void main() {
   late String atSign_1;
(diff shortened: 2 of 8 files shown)


cpswan (dismissed):

cpswan:
Change LGTM, but the e2e tests on staging suggest that this isn't having the desired effect :/

gkc:
> Change LGTM, but the e2e tests on staging suggest that this isn't having the desired effect :/

yeah found a bug in the test itself, have fixed

tests/at_end2end_test/test/notify_verb_test.dart
```diff
@@ -37,7 +40,7 @@ void main() {
   });
 
   setUp(() async {
-    print("Clearing socket response queues");
+    logger.info("Clearing socket response queues");
```
gkc, line 43: replaced all prints in this file with logger.info so can see timestamps

tests/at_end2end_test/test/notify_verb_test.dart
```diff
@@ -92,18 +95,22 @@ void main() {
...
+            log: true, timeoutMillis: 5000, throwTimeoutException: true);
+        final notifJson =
+            jsonDecode(receivedNotif.replaceFirst('notification: ', ''));
+        var notifId = notifJson['id'];
+        if (toCheck.contains(notifId)) {
+          logger.info('Received $notifId');
+          toCheck.remove(notifId);
+        }
       }
+      expect(toCheck, isEmpty);
```
gkc, line 113: The rework above makes the test resilient to receiving "stats" notifications with id -1 - this caused a test run to fail earlier.

tests/at_end2end_test/test/at_demo_data.dart
```diff
@@ -1,5 +1,5 @@
-///This is a dummy file which gets overwritten by the github actions with the
-/// atsign's keys map.
+// This is a dummy file which gets overwritten by the github actions with the
+//  atsign's keys map.
```
cpswan, line 2: why?

cpswan (approved):

tests/at_end2end_test/test/at_demo_data.dart
gkc: That's the way these tests were implemented, and changing the impl has never been prioritized.

cpswan: @gkc My question was about the change from `///` to `// `

gkc: @cpswan  Linter correctly picked it up as a dangling doc comment

PR #2320, opened 2025-06-30
fix: Prevent OutboundClient from unnecessarily creating new socket connections

gkc:
**- What I did**
fix: Prevent OutboundClient from unnecessarily creating new socket connections because of OutboundClient.connect being called unnecessarily.

**- How I did it**
- Made some implicit things explicit by making some optional parameters mandatory
- Added a field (`handshakeRequired`) to OutboundClient to show its intended usage. Note that it is never the case that an OutboundClient is used both for unauthenticated requests (e.g. looking up public data) and authenticated requests.
- In turn this allows for OutboundClient to strongly assert state and prevent any bugs related to handshake being required or not.
- OutboundClient.connect also now checks if it is already connected and, if so, just returns.
- refactor: OutboundClientManager.getClient and NotificationConnectionPool.getOutboundClient now will call `connect` on new OutboundClients which they create, before adding them to their respective pools. This eliminates multiple bug vectors and removes possibility of races around the `connect` part since other parts of the codebase just get clients from the pools maintained by OutboundClientManager / NotificationConnectionPool

**- How to verify it**
Tests pass


packages/at_secondary_server/lib/src/caching/cache_manager.dart
```diff
@@ -27,6 +27,8 @@ class AtCacheManager {
 
   final logger = AtSignLogger('AtCacheManager');
 
+  final _dummyInboundConnection = DummyInboundConnection();
```
gkc, line 31: No need to make a new one every time

packages/at_secondary_server/lib/src/connection/outbound/outbound_client.dart
```diff
@@ -67,8 +68,25 @@ class OutboundClient {
...
   /// Throws a [SocketException] when a socket connection to secondary cannot be established
   /// Throws a [HandShakeException] for any exception in the handshake process
-  Future<bool> connect({bool handshake = true}) async {
-    logger.finer('connect(handshake:$handshake) called for $toAtSign');
+  Future<bool> connect({required bool handshake}) async {
+    // TODO There is the potential for races here under heavy load.
+    // TODO Need to make isHandshakeDone and isConnectionCreated into futures
+    // which return the futures of underlying completers which are completed
+    // when the appropriate thing has happened.
+    if (handshake != handshakeRequired) {
```
gkc: I have reviewed the full codebase and this is never the case right now. This check has been added for the safety of future maintainers.

packages/at_secondary_server/lib/src/connection/outbound/outbound_client_manager.dart
```diff
@@ -10,7 +10,7 @@ import 'package:meta/meta.dart';
 class OutboundClientManager {
   var logger = AtSignLogger('OutboundClientManager');
 
-  static const int defaultPoolSize = 10;
+  static const int defaultPoolSize = 200;
```
gkc, line 13: You don't need it ... until you need it

packages/at_secondary_server/lib/src/connection/outbound/outbound_client.dart
```diff
@@ -51,8 +52,8 @@ class OutboundClient {
 
   late OutboundConnectionFactory _outboundConnectionFactory;
 
-  OutboundClient(
-      this.inboundConnection, this.toAtSign, this.secondaryAddressFinder,
+  OutboundClient(this.inboundConnection, this.toAtSign,
+      this.secondaryAddressFinder, this.handshakeRequired,
```
gkc, line 56: Added final field `handshakeRequired` so there is a strong and explicit relationship between how the client was created and how it is subsequently used in `connect`

gkc: As I wrote the above comment I realized that in fact we can now get rid of the `handshake` parameter in `connect`

gkc: ... which I've done

packages/at_secondary_server/lib/src/connection/outbound/outbound_client.dart
```diff
@@ -67,8 +68,12 @@ class OutboundClient {
   /// Throws a [SecondaryNotFoundException] if secondary is url is not found for atsign
   /// Throws a [SocketException] when a socket connection to secondary cannot be established
   /// Throws a [HandShakeException] for any exception in the handshake process
-  Future<bool> connect({bool handshake = true}) async {
-    logger.finer('connect(handshake:$handshake) called for $toAtSign');
+  Future<bool> connect() async {
```
gkc, line 71: Removed the `handshake` parameter since OutboundClients are always created for either unauthenticated or pol-authenticated connections, but not both,. Thus, handshakeRequired is now a final field in the class, required in constructor.

packages/at_secondary_server/lib/src/connection/outbound/outbound_client.dart
```diff
@@ -67,8 +68,12 @@ class OutboundClient {
...
+    if (isConnectionCreated) {
+      logger.warning('connect called for $toAtSign but is already connected');
+      logger.warning(StackTrace.current);
+      return isHandShakeDone;
+    }
```
gkc, line 76: The above situation doesn't happen but if it does it will be nice to have a stack trace.

gkc:
Dang it, went too far in the last few commits. Reverting to Draft until I figure it out.

gkc:
> Dang it, went too far in the last few commits. Reverting to Draft until I figure it out.

Fixed by [this commit](https://github.com/atsign-foundation/at_server/pull/2320/commits/dd80451503a4242d25d6e8e27f61f8a57915a46e)

cpswan (approved):

PR #2346, opened 2025-07-30
fix: set the trusted cacert path for AtSecondaryFinder

xavierchanth:
**- What I did**

- Ensure trusted ca bundle is used when talking atServer to root server.

**- How I did it**

**- How to verify it**

- Needs to be deployed with a trusted ca bundle in a non-standard location and an application which triggers the atServer to lookup another atSign needs to be called (such as npt).

@cpswan I think you're most equipped to test this, would you mind lending a hand?

fix: set the trusted cacert path for AtSecondaryFinder

### DIFF
--- a/packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
+++ b/packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
@@ -80,7 +80,7 @@ class AtSecondaryServerImpl implements AtSecondaryServer {
 
   AtSecondaryServerImpl._internal() {
     final socketConfig = SecureSocketConfig()
-      ..decryptPackets = true
+      ..decryptPackets = false
       ..pathToCerts = AtSecondaryConfig.trustedCertificateLocation
       ..tlsKeysSavePath = null;
 
--- a/packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
+++ b/packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
@@ -79,8 +79,16 @@ class AtSecondaryServerImpl implements AtSecondaryServer {
   }
 
   AtSecondaryServerImpl._internal() {
+    final socketConfig = SecureSocketConfig()
+      ..decryptPackets = true
+      ..pathToCerts = AtSecondaryConfig.trustedCertificateLocation
+      ..tlsKeysSavePath = null;
+
     secondaryAddressFinder = CacheableSecondaryAddressFinder(
-        AtSecondaryConfig.rootServerUrl, AtSecondaryConfig.rootServerPort);
+      AtSecondaryConfig.rootServerUrl,
+      AtSecondaryConfig.rootServerPort,
+      socketConfig: socketConfig,
+    );
     outboundClientManager = OutboundClientManager(secondaryAddressFinder);
   }
 
--- a/packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
+++ b/packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
@@ -577,7 +577,7 @@ class AtSecondaryServerImpl implements AtSecondaryServer {
 
   /// Starts the secondary server in secure mode and calls the listen method of server socket.
   Future<void> _startSecuredServer() async {
-    var secCon = SecurityContext();
+    var secCon = SecurityContext.defaultContext;
     var retryCount = 0;
     var certsAvailable = false;
     // if certs are unavailable then retry max 10 minutes
--- a/packages/at_secondary_server/CHANGELOG.md
+++ b/packages/at_secondary_server/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 3.5.3
+- fix: Set the trusted cacert path for AtSecondaryFinder
+
 # 3.5.2
 - fix: Prevent OutboundClient from creating new socket connections unnecessarily
 
--- a/packages/at_secondary_server/lib/src/connection/outbound/outbound_client.dart
+++ b/packages/at_secondary_server/lib/src/connection/outbound/outbound_client.dart
@@ -380,7 +380,7 @@ class DefaultOutboundConnectionFactory implements OutboundConnectionFactory {
   Future<OutboundSocketConnection> createOutboundConnection(
       String host, int port, String toAtSign) async {
     AtSecurityContextImpl securityContext = AtSecurityContextImpl();
-    SecurityContext secConConnect = SecurityContext();
+    SecurityContext secConConnect = SecurityContext.defaultContext;
     secConConnect.useCertificateChain(securityContext.publicKeyPath());
     secConConnect.usePrivateKey(securityContext.privateKeyPath());
     secConConnect
--- a/packages/at_secondary_server/pubspec.yaml
+++ b/packages/at_secondary_server/pubspec.yaml
@@ -1,12 +1,12 @@
 name: at_secondary
 description: Implementation of secondary server.
-version: 3.5.2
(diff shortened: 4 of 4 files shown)


xavierchanth:
Some odds and ends in at_lookup to be discussed tomorrow.

Upon resolution of those items and approval of this PR, then I will remove the dependency override.

packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
```diff
@@ -79,8 +79,16 @@ class AtSecondaryServerImpl implements AtSecondaryServer {
   }
 
   AtSecondaryServerImpl._internal() {
+    final socketConfig = SecureSocketConfig()
+      ..decryptPackets = true
+      ..pathToCerts = AtSecondaryConfig.trustedCertificateLocation
+      ..tlsKeysSavePath = null;
```
xavierchanth, line 85: Create a SecureSocketConfig for the secondaryAddressFinder.

packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
```diff
@@ -79,8 +79,16 @@ class AtSecondaryServerImpl implements AtSecondaryServer {
...
+
     secondaryAddressFinder = CacheableSecondaryAddressFinder(
-        AtSecondaryConfig.rootServerUrl, AtSecondaryConfig.rootServerPort);
+      AtSecondaryConfig.rootServerUrl,
+      AtSecondaryConfig.rootServerPort,
+      socketConfig: socketConfig,
```
xavierchanth, line 90: Add the config.

xavierchanth:
Related PR: https://github.com/atsign-foundation/at_libraries/pull/814

packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
gkc: Should be setting decryptPackets to `false`

gkc:
@XavierChanth As we discussed last night the creation of socket connections to other atServers will also be needed. I'll start working on that while @cpswan is testing what has been done so far

xavierchanth:
> @XavierChanth As we discussed last night the creation of socket connections to other atServers will also be needed. I'll start working on that while @cpswan is testing what has been done so far

atServer to atServer socket creation already loads the ca from the atServer config via DefaultOutboundConnectionFactory and 
AtSecurityContextImpl.

packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
xavierchanth: Right… i may need to do more work in at_lookup then. I don’t think ca bundle is loaded at all without this being set to true.

xavierchanth: At least, i didn’t see a code path that did yesterday, I’ll take another look.

gkc: Ah yes I see what you mean. Let's leave as is for now here; I'll create another ticket in at_libraries to clean up that code in at_lookup so we always use the certs bundle regardless of the value of `decryptPackets`

gkc: Actually, since https://github.com/atsign-foundation/at_libraries/pull/814 is still draft, I'll make the change there now

gkc: Done, and accordingly I've added [this commit](https://github.com/atsign-foundation/at_server/pull/2346/commits/5db52b40b3cf0933787e0a693823169803085275) here

gkc:
> > @XavierChanth As we discussed last night the creation of socket connections to other atServers will also be needed. I'll start working on that while @cpswan is testing what has been done so far
> 
> atServer to atServer socket creation already loads the ca from the atServer config via DefaultOutboundConnectionFactory and AtSecurityContextImpl.

Ah yes so it does. One thing I notice there is that we're constructing a SecurityContext rather than using `SecurityContext.defaultContext`. I've already changed this in the at_lookup PR; I'm going to change it here also.

packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
xavierchanth: Thanks @gkc

gkc:
Manual testing completed. I did the following:
- created my own CA for self-signing, added its cert to my local trusted ca cert store, generated a new self-signing private key, generated a self-signed cert with that key, make a certificate signing request with that cert, and signed the cert with my CA
- successfully spun up atDirectory and four atServers (from the trunk version of the code), using my self-signed certs instead of those for vip.ve.atsign.zone
- verified I can connect to atDirectory and atServer with `openssl s_client` without any cert verification error
- successfully run at_activate onboard for my four atServers (`@ssc_client, @ssc_service, @ssc_policy, @ssc_relay`)
- verified that we have atServer-to-atServer connectivity failures using the *_trunk_* branch
- restarted the four atServers from this branch of the code
- successfully started various NoPorts services
  - NoPorts daemon: `sshnpd --root-domain services.atsign.gkc -a @ssc_service -p @ssc_policy -d mbp --permit-open '*:*'`
  - NoPorts relay: `sudo srvd -a @ssc_relay --root-domain services.atsign.gkc -i 127.0.0.1`
  - NoPorts policy service: `npp_atserver --atsign @ssc_policy -v --root-domain services.atsign.gkc`
  - NoPorts policy admin app: `np_admin -a @ssc_policy -a @ssc_policy --root-domain services.atsign.gkc`
- and successfully created and used an npt session
  - `npt --root-domain services.atsign.gkc -f @ssc_client -t @ssc_service -r @ssc_relay --rh 127.0.0.1 --rp 3000 --lp 3001 -d mbp --ram escr --no-pss`
    - curl to the policy admin api to port 3001 which is npt-tunnelled to port 3000
    ```
    gary@gkc2025 ~ % curl --silent http://localhost:3001/api/policy/group/1 | jq
    {
      "id": "1",
      "name": "@ssc_client",
      "description": "Just this @client",
      "daemonAtSigns": [
        "@ssc_service"
      ],
      "devices": [
        {
          "name": "mbp",
          "permitOpens": [
            "localhost:3000",
            "127.0.0.1:3000"
          ]
        }
      ],
      "deviceGroups": [],
      "userAtSigns": [
        "@ssc_client"
      ]
    }
    ```
  - No warnings or error logged by any of the atServers


I've also verified that everything still looks fine when using atServers from this branch which are using the vip.ve.atsign.zone certs

I think therefore this branch is good to be merged and released to canary once I've published the new `at_lookup` and updated the at_lookup dependency here

xavierchanth:
Thanks for your help @gkc !

gkc:
Hmmm. Not so fast ... if I tell macos not to trust my ca, then the atServer fails to talk to my atDirectory

gkc:
> Hmmm. Not so fast ... if I tell macos not to trust my ca, then the atServer fails to talk to my atDirectory

I'll need to repeat this test on a Linux VM to be certain

gkc:
Have verified on a clean Linux vm
- fails if the pem isn't in /etc/cacert/cacert.pem
- succeeds if it is

Another MacOS special ... basically, [SecurityContext.setTrustedCertificates](https://api.flutter.dev/flutter/dart-io/SecurityContext/setTrustedCertificates.html) has no effect on MacOS

cpswan (approved):

cconstab (approved):

PR #2363, opened 2025-08-17
feat: more http stuff, more tests

gkc:
**- What I did**
- feat: support query param `at_rt` in http GET requests to atServer. `…at_rt=meta` means get the metadata
- feat: root transmits query parameters to atServer in redirect URLs
- feat: support path-style or dot-style http requests - so GET /index.html.foo.bar.baz will work the same as GET /baz/bar/foo/index.html or GET /baz/bar/foo/html/index
- feat: infer the response content type based on the lookup key (previous example will set `text/html; charset=utf-8`). Default to `text/plain; charset=utf-8` for text data, `application/octet-stream` for binary data
- feat: support fetching of binary data. (Still TODO: functional tests for this feature)
- also support client specifying the content type via the `at_ct` (ct is for content-type) - `at_ct=application/json; charset=utf-8` for example
- test: lots of unit tests in atServer
- test: added redirect tests with query parameters to at_functional_tests
- build: updated tools/build_ve and tools/build_ee to export port 444 and set `httpsEnabled` to true by default in the atDirectory config

**- How I did it**
- See commits
- There are a few minor lint cleanups in there as well

**- How to verify it**
Tests pass


cconstab (commented):
Probably a good idea to enable the 443 options in EE and VE

gkc:
> Probably a good idea to enable the 443 options in EE and VE

Done in [this commit](https://github.com/atsign-foundation/at_server/pull/2363/commits/22735727ce7127f8c84e6b957dc6f54d35245033)

cpswan (dismissed):

gkc:
Setting back to draft until I've stopped the tests from flaking.

cpswan (approved):

PR #2371, opened 2025-08-25
fix: initial monitor response (existing matching notifications) was not including all metadata

gkc:
**- What I did**
- fix: Fixed issue where notifications in the initial response from atServer were not including the (very important) `sharedKeyEnc` metadata attribute, while notifications received after the initial monitor command were correctly including the `sharedKeyEnc`.
- test: added tests for the same

**- How I did it**
See commits

**- How to verify it**
Tests pass


gkc:
@cconstab @XavierChanth if you have a few minutes today please review; it fixes a bug which is blocking [this](https://github.com/atsign-foundation/at_client_sdk/pull/1635) at_client PR

xavierchanth (approved):

PR #2380, opened 2025-08-29
build(deps): rollup 2377 2378 2379

gkc:
**- What I did**
Rollup of #2377 #2378 #2379


cpswan (changes requested):
Don't merge beta Dart

PR #2396, opened 2025-09-18
feat: better idle time handling for inbound connections

gkc:
**- What I did**
- feat: change default inbound idle time for authenticated connections to 10 minutes (same as unauthenticated connections)
- feat: revert stats:1 to previous behaviour; add stats:16 (summary of inbound connections by type) and stats:17 (detail on all inbound connections)
- refactor: removed singleton nature of InboundConnectionManager and InboundConnectionPool
- feat: added a `closed` guard to InboundConnectionManager (and renamed its `clearInvalidConnections` function to `close`) to prevent new connections being established when the manager is closed (which happens when the atServer is doing a soft restart)
- refactor: removed the `ConnectionUtil` class and moved  its static methods to be instance methods of InboundConnectionPool
- refactor: renamed the `MetricNames` enum to just `Metric`; removed the `MetricClasses` extension and moved its functionality to StatsVerbHandler
- fix: removed a try-catch from StatsVerbHandler which was masking actual causes of exceptions
- refactor: removed singleton nature of all MetricProviders
- refactor: some other small refactorings (removing redundant variables etc)

**- How I did it**

**- How to verify it**
Tests pass

### DIFF
--- a/packages/at_secondary_server/CHANGELOG.md
+++ b/packages/at_secondary_server/CHANGELOG.md
@@ -1,7 +1,12 @@
-# 3.7.0
+# 3.7.1
 - feat : added `stats:16` for a summary of number of inbound connections by 
   type (self, other, anon) and `stats:17` for a detailed report on all 
   inbound connections including atSigns, time established, last accessed time.
+- fix: better idle time defaults for inbound and outbound connections, 
+  authenticated and unauthenticated 
+- refactor: removed a bunch of singletons
+
+# 3.7.0
 - fix: better idle time defaults for inbound and outbound connections, 
   authenticated and unauthenticated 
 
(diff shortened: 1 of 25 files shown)


packages/at_secondary_server/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # 3.7.0
```
cpswan: Shouldn't this be 3.7.1 given 3.7.0 was already released?

cpswan (changes requested):

packages/at_secondary_server/CHANGELOG.md
gkc: Only to canary

gkc: So per previous precedent next canary would be 3.7.0a

cpswan: Canary is still a release.

I don't know how we got into the habit of a,b,c suffixing versions rather than just bumping the patch. We've never done an ADR on it :/

Anyways... I think there's enough change here to justify the .1

cpswan (approved):

PR #2412, opened 2025-10-14
fix: improved memory usage and error handling in StatsNotificationService

gkc:
**- What I did**
- fix: improved memory usage and error handling in StatsNotificationService. Fixes a minor bug in StatsNotificationService which would only occur when running an atServer on a development machine which is put to "sleep" for a while.
- refactor: Removed unnecessary instance variable from StatsNotificationService

**- How I did it**

**- How to verify it**
Tests pass

**- Context**
The problem being fixed was as follows:
- Given
  - You are running an atServer on your dev machine
  - And you have a monitor connection from a client
- When
  - You put your machine to sleep for more than the idle timeout time (10 minutes) so the atServer has not received any `noop` heartbeats from the client
  - And you then resume from sleep
- Then
  - The atServer StatsNotificationService would attempt to write stats to the monitor connection
  - Which would be deemed invalid
  - And therefore cause the `write` on the InboundConnectionImpl to throw an exception back to `writeStatsToMonitor`
  - And therefore cause the atServer to exit with an uncaught exception because `writeStatsToMonitor` was not called with an `await`

### DIFF
--- a/packages/at_secondary_server/lib/src/connection/inbound/inbound_message_listener.dart
+++ b/packages/at_secondary_server/lib/src/connection/inbound/inbound_message_listener.dart
@@ -43,6 +43,7 @@ class InboundMessageListener {
   /// Handles messages on the inbound client's connection and calls the verb executor
   /// Closes the inbound connection in case of any error.
   Future<void> _messageHandler(streamData) async {
+    connection.metaData.lastAccessed = DateTime.now().toUtc();
     logger.finest('_messageHandler received ${streamData.runtimeType}'
         ' : $streamData ');
     List<int> data;
--- a/packages/at_secondary_server/lib/src/notification/stats_notification_service.dart
+++ b/packages/at_secondary_server/lib/src/notification/stats_notification_service.dart
@@ -58,9 +58,6 @@ class StatsNotificationService {
 
   static final Duration zeroDuration = Duration(microseconds: 0);
 
-  // Counter for number of active monitor connections. Used for logging purpose.
-  int numOfMonitorConn = 0;
-
   Notification notification = Notification.empty();
 
   @visibleForTesting
@@ -108,16 +105,13 @@ class StatsNotificationService {
         await AtCommitLogManagerImpl.getInstance().getCommitLog(currentAtSign);
 
     // Runs the _schedule method as long as server is up and running.
-    timer = Timer.periodic(interval, (timer) {
+    timer = Timer.periodic(interval, (timer) async {
       try {
         _logger.finer('Stats Notification Job triggered');
-        writeStatsToMonitor();
+        await writeStatsToMonitor();
         _logger.finer('Stats Notification Job completed');
-      } on Exception catch (exception) {
-        _logger.severe(
-            'Exception occurred when writing stats ${exception.toString()}');
-      } on Error catch (error) {
-        _logger.severe('Error occurred when writing stats ${error.toString()}');
+      } catch (e) {
+        _logger.severe('Exception while writing stats: $e');
       }
     });
     state = StatsNotificationServiceState.scheduled;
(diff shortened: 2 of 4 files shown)


packages/at_secondary_server/lib/src/connection/inbound/inbound_message_listener.dart
```diff
@@ -43,6 +43,7 @@ class InboundMessageListener {
   /// Handles messages on the inbound client's connection and calls the verb executor
   /// Closes the inbound connection in case of any error.
   Future<void> _messageHandler(streamData) async {
+    connection.metaData.lastAccessed = DateTime.now().toUtc();
```
gkc, line 46: Previously, lastAccessed was only being set when data was written to a connection, and not when the data was received on it. Not the cause of any particular problem right now but had the potential to be a problem in some future scenarios.

packages/at_secondary_server/lib/src/notification/stats_notification_service.dart
```diff
@@ -58,9 +58,6 @@ class StatsNotificationService {
 
   static final Duration zeroDuration = Duration(microseconds: 0);
 
-  // Counter for number of active monitor connections. Used for logging purpose.
-  int numOfMonitorConn = 0;
```
gkc, line 62: should not have been an instance variable

packages/at_secondary_server/lib/src/notification/stats_notification_service.dart
```diff
@@ -108,16 +105,13 @@ class StatsNotificationService {
         await AtCommitLogManagerImpl.getInstance().getCommitLog(currentAtSign);
 
     // Runs the _schedule method as long as server is up and running.
-    timer = Timer.periodic(interval, (timer) {
+    timer = Timer.periodic(interval, (timer) async {
```
gkc, line 108: added async so we could add `await` to the call to `writeStatsToMonitor` below

packages/at_secondary_server/lib/src/notification/stats_notification_service.dart
```diff
@@ -108,16 +105,13 @@ class StatsNotificationService {
...
       try {
         _logger.finer('Stats Notification Job triggered');
-        writeStatsToMonitor();
+        await writeStatsToMonitor();
```
gkc, line 111: Without the `await`, if `writeStatsToMonitor` threw an exception then it would be an uncaught exception and the atServer would exit

packages/at_secondary_server/lib/src/notification/stats_notification_service.dart
```diff
@@ -108,16 +105,13 @@ class StatsNotificationService {
...
         _logger.finer('Stats Notification Job completed');
-      } on Exception catch (exception) {
-        _logger.severe(
-            'Exception occurred when writing stats ${exception.toString()}');
-      } on Error catch (error) {
-        _logger.severe('Error occurred when writing stats ${error.toString()}');
+      } catch (e) {
+        _logger.severe('Exception while writing stats: $e');
```
gkc, line 114: only need one catch block

packages/at_secondary_server/lib/src/notification/stats_notification_service.dart
```diff
@@ -133,41 +127,48 @@ class StatsNotificationService {
...
+      if (connection.isMonitor ?? false) {
+        numOfMonitorConn++;
+        try {
           // Convert notification object to JSON and write to connection
           await connection.write('notification:'
               ' ${jsonEncode(notification.toJson())}\n');
+        } catch (e) {
+          _logger.severe('Exception occurred when writing stats to'
+              ' inbound connection session ${connection.metaData.sessionID}'
+              ' : ${e.toString()}');
```
gkc, line 164: Moved the setting of the `notification` fields to outside the loop so it only happens once rather than once per monitor connection.

JeremyTubongbanua (approved):

JeremyTubongbanua:
LGTM

PR #2553, opened 2026-03-01
feat: at_persistence_secondary_server 4.3.0

gkc:
**- What I did**
Cleanup as part of a larger overhaul of at_secondary_server's notifications subsystem
- feat: deprecated AtNotificationCallback class
- feat: deprecated use of AtNotificationKeystore singleton
- feat: deprecated NotificationManagerSpec
- fix: AtNotificationBuilder.build will update a null ttl to the default value
- fix: AtNotification.isExpired treats null expiry as expired, as the idea of notifications without expiration is a historical antipattern and is no longer possible

**- How I did it**
See commits

**- How to verify it**
Tests pass


packages/at_persistence_secondary_server/lib/src/notification/at_notification.dart
```diff
@@ -460,6 +461,7 @@ class AtNotificationBuilder {
   AtMetaData? atMetaData;
 
   AtNotification build() {
+    ttl ??= Duration(minutes: _defaultTTLInMins).inMilliseconds;
```
gkc, line 464: notifications should always have an expiration

packages/at_persistence_secondary_server/lib/src/notification/at_notification.dart
```diff
@@ -142,7 +142,8 @@ class AtNotification {
   }
 
   bool isExpired() {
-    return _expiresAt != null && _expiresAt!.isBefore(DateTime.now().toUtc());
+    // expiresAt == null never made sense and never now happens
+    return _expiresAt == null || _expiresAt!.isBefore(DateTime.now().toUtc());
```
gkc, line 146: Notifications should always have an expiry. Any which don't are legacy from the distant past and should be considered expired.

cpswan (approved):

PR #2566, opened 2026-03-04
fix: Fix the `expiresAt` and `ttl` rules in `AtNotificationBuilder.build`

gkc:
**- What I did**
- fix: Fix the `expiresAt` and `ttl` rules in `AtNotificationBuilder.build`
- test: added bunch more tests to verify behaviour of expiresAt and ttl rules
- chore: dart format
- refactor: renamed some ttl-related static vars for readability

The rules in `AtNotificationBuilder.build` are now as follows:
```
  /// Applies the following rules, then builds the [AtNotification]
  /// 1. if [expiresAt] is null
  ///   1. If [ttl] is null or non-positive, set to [defaultTtl]
  ///   2. Ensure that [ttl] is no greater than [AtNotification.maxTtl]
  ///   3. calculate [expiresAt] by adding [ttl] to [DateTime.now]
  /// 2. Ensure that [expiresAt] is no more than [AtNotification.maxTtl] in
  /// the future
  ///
  /// Explanation of the above:
  /// - [ttl] can be supplied by AtClients when sending a notification, and
  /// [expiresAt] is calculated at that time. (Note: In future, clients will be
  /// able to specify a precise [expiresAt], and the above rules will work
  /// unchanged.)
  /// - expiresAt should be sent as-is from sending atServer to receiving
  /// - if expiresAt is set, ttl is not relevant
```

**- How I did it**
See commits

**- How to verify it**
Tests pass

### DIFF
--- a/packages/at_persistence_secondary_server/lib/src/notification/at_notification.dart
+++ b/packages/at_persistence_secondary_server/lib/src/notification/at_notification.dart
@@ -1,3 +1,5 @@
+import 'dart:math';
+
 import 'package:at_persistence_secondary_server/at_persistence_secondary_server.dart';
 import 'package:at_persistence_secondary_server/src/utils/type_adapter_util.dart';
 import 'package:hive/hive.dart';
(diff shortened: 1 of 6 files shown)


packages/at_persistence_secondary_server/lib/src/notification/at_notification.dart
```diff
@@ -141,15 +143,20 @@ class AtNotification {
         'atValue:$atValue';
   }
 
-  /// If notificationDateTime is before now minus this duration,
+  /// 1. A notification may have an expiration no more than this duration into
+  /// the future.
+  /// 2. If notificationDateTime is before now minus this duration,
   /// then the notification is considered to be expired.
-  static Duration notificationDateTimeExpiryTimeout = Duration(days: 8);
+  static Duration maxTtl = Duration(days: 8);
```
gkc, line 150: Renamed to maxTtl which is kind of a better name, lol

packages/at_persistence_secondary_server/lib/src/notification/at_notification.dart
```diff
@@ -141,15 +143,20 @@ class AtNotification {
...
+
+  /// The default used by [AtNotificationBuilder.build] to calculate expiresAt
+  static Duration defaultTtl = Duration(minutes: 15);
```
gkc, line 153: Moved to here from the builder, for readability, and renamed and retyped, also for readability

akafredperry (approved):
looks good

PR #2567, opened 2026-03-04
fix: more notification-related fixes / improvements

gkc:
**- What I did**
- from #2564
  - fix: in NotificationManager.reEnqueueUndelivered, update the notification state to 'errored' if enqueuing fails
  - refactor: start using Atsign type in various places
- take up at_persistence_secondary_server 4.3.3 following completion of #2565 and #2566
- docs: correct a comment in atServer config.yaml
- fix: make default notification ttl in at_secondary_config.dart consistent with the value in config.yaml

The changes in the above PRs:
- fix: in NotificationManager.reEnqueueUndelivered, update the notification state to 'errored' if enqueuing fails
- refactor: start using Atsign type in various places
- fix: AtNotification.isExpired
- fix: AtNotification.reset was using defaultTTLInMins as an hours value, smh
- feat: make AtNotificationBuilder.defaultTTLInMins settable
- test: modified some tests to run faster
- test: added tests fully covering isExpired
- test: added tests of AtNotificationBuilder.reset
- fix: Fix the `expiresAt` and `ttl` rules in `AtNotificationBuilder.build`
- test: added bunch more tests to verify behaviour of expiresAt and ttl rules
- chore: dart format
- refactor: renamed some ttl-related static vars for readability

The rules in `AtNotificationBuilder.build` are now as follows:
```
  /// Applies the following rules, then builds the [AtNotification]
  /// 1. if [expiresAt] is null
  ///   1. If [ttl] is null or non-positive, set to [defaultTtl]
  ///   2. Ensure that [ttl] is no greater than [AtNotification.maxTtl]
  ///   3. calculate [expiresAt] by adding [ttl] to [DateTime.now]
  /// 2. Ensure that [expiresAt] is no more than [AtNotification.maxTtl] in
  /// the future
  ///
  /// Explanation of the above:
  /// - [ttl] can be supplied by AtClients when sending a notification, and
  /// [expiresAt] is calculated at that time. (Note: In future, clients will be
  /// able to specify a precise [expiresAt], and the above rules will work
  /// unchanged.)
  /// - expiresAt should be sent as-is from sending atServer to receiving
  /// - if expiresAt is set, ttl is not relevant
```

**- How I did it**
See commits

**- How to verify it**
Tests pass

### DIFF
--- a/packages/at_persistence_secondary_server/test/hive_keystore_impl_test.dart
+++ b/packages/at_persistence_secondary_server/test/hive_keystore_impl_test.dart
@@ -728,12 +728,12 @@ void main() async {
       HiveKeystore? keystore = keyStoreManager?.getSecondaryKeyStore();
       AtData atData = AtData()
         ..data = 'dummy_value'
-        ..metaData = (AtMetaData()..ttl = 20);
+        ..metaData = (AtMetaData()..ttl = 30);
       await keystore?.put('keyabouttoexpire.wavi@test_user_1', atData);
       var expiredKeysList = await keystore?.getExpiredKeys();
       expect(expiredKeysList?.contains('keyabouttoexpire.wavi@test_user_1'),
           false);
-      await Future.delayed(Duration(milliseconds: 21));
+      await Future.delayed(Duration(milliseconds: 31));
       expiredKeysList = await keystore?.getExpiredKeys();
       expect(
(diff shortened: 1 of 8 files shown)


cpswan (dismissed):

gkc:
Fixed flaky tests

cpswan (approved):

PR #2592, opened 2026-03-20
feat: make presentation of client certificates configurable for atServer-to-atServer communication

### LINKED ISSUE #2550: Add ability for atServers to disable mTLS check within the virtual environment

If customers wish to use a virtual environment as a completely private instance of the platform, but as "production", then for full security it will be important for customers to manage the certs for that environment. (Note: while it is feasible to use the environment out-of-the-box, this allows for the possibility of a bad actor on the local network pointing DNS to a different IP address, where they are running another copy of the VE)

Managing certs has two aspects:
- server certs - the certs which allow atClients to verify that the server to which they are connecting is presenting a valid cert
- client certs - the certs which atServers present to each other for mTLS as an additional layer of security protection

It is straightforward for customers to manage their own server certs.

However, managing the client certs for mTLS is no longer straightforward. Options for customers would be
1. Run their own certificate authority. Issue client certs using that CA. Both (i) the certs and (ii) the CA public key would then be mounted into the VE. This is a **lot** of work for customers - cf the work that @cpswan did over the last few months of 2025
2. Have Atsign issue client certs upon customer request. This is a lot of work for both the customers and Atsign
3. (Proposed): Add ability for atServers to disable mTLS check within the VE

We believe the proposed option 3 (Add ability for atServers to disable mTLS check within the VE) to be safe from a security perspective, since the atServers in the VE can only communicate out to other atServers which are also in the VE, as they look up the address of an atSign via the atDirectory in the VE. 

To work this through:
Given
- We have a VE running at `my-atplatform.acme.com`
- We have `@alice` and `@bob` atServers running in the VE
- `@alice` and `@bob` atServers resolve each other's atServer addresses by connecting to the atDirectory which is also in the VE
- the ve atDirectory will not contain any addresses other than addresses within the VE

When
- I spin up an evil `@alice` atServer elsewhere
- and the evil `@alice` atServer connects to the `@bob` atServer within the VE

Then
- `@bob` does not verify the client cert on the connection from `@alice` (This is the change being proposed)
- evil `@alice` sends `from` command
- `@bob` replies with the challenge
- evil `@alice` sends `pol` with the challenge response key for `@bob` to look up
- `@bob` looks up `@alice` address via the VE's atDirectory and receives, e.g., `my-atplatform.acme.com:29005`
- `@bob` connects to the ('good') `@alice` atServer within the VE to do the `pol` verification (fetch the challenge response, and verify the signature)
- `pol` verification fails because the `@alice` within the VE will not have the relevant challenge-response key

gkc:
**- What I did**
- resolves #2550 
- feat: make presentation of client certificates configurable for atServer-to-atServer communication
- fix: on startup, enqueue undelivered notifications AFTER opening the serverSocket, not before
- build: update pubspec and changelog for 3.11.0

**- How I did it**
See commits

**- How to verify it**
Tests pass

### DIFF
--- a/packages/at_secondary_server/lib/src/verb/handler/from_verb_handler.dart
+++ b/packages/at_secondary_server/lib/src/verb/handler/from_verb_handler.dart
@@ -19,8 +19,6 @@ class FromVerbHandler extends AbstractVerbHandler {
   static From from = From();
   static final _rootDomain = AtSecondaryConfig.rootServerUrl;
   static final _rootPort = AtSecondaryConfig.rootServerPort;
-  static final bool? clientCertificateRequired =
-      AtSecondaryConfig.clientCertificateRequired;
 
   FromVerbHandler(super.keyStore);
 
@@ -75,7 +73,7 @@ class FromVerbHandler extends AbstractVerbHandler {
     }
 
     if (fromAtSign != AtSecondaryServerImpl.getInstance().currentAtSign &&
-        clientCertificateRequired!) {
+        AtSecondaryConfig.clientCertificateRequired) {
       var result = await _verifyFromAtSign(fromAtSign, atConnection);
       logger.finer('_verifyFromAtSign result : $result');
       if (!result) {
@@ -127,7 +125,7 @@ class FromVerbHandler extends AbstractVerbHandler {
       return atConnection.metaData.isStream;
     }
 
-    if (clientCertificateRequired!) {
+    if (AtSecondaryConfig.clientCertificateRequired) {
       var result = _verifyClientCerts(cn, host);
       return result;
     }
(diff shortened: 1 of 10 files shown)


JeremyTubongbanua (approved):

packages/at_secondary_server/lib/src/verb/handler/from_verb_handler.dart
```diff
@@ -127,33 +125,34 @@ class FromVerbHandler extends AbstractVerbHandler {
...
+        'Connected from: $cn ${cn.subject} issued by ${cn.issuer} valid from ${cn.startValidity} to ${cn.endValidity}');
+
+    X509CertificateData certData = X509Utils.x509CertificateFromPem(cn.pem);
+    List<String> subjectAlternativeNames =
+        certData.tbsCertificate?.extensions?.subjectAlternativNames ?? [];
+    logger.info('SAN: $subjectAlternativeNames');
+
+    String commonName = certData.tbsCertificate?.subject['2.5.4.3'] ?? '';
+    logger.info('CN: $commonName');
+
```
JeremyTubongbanua, line 146: What does the 2.5.4.3 mean?

JeremyTubongbanua:
LGTM thanks @gkc

packages/at_secondary_server/lib/src/verb/handler/from_verb_handler.dart
gkc: part of the x509 spec

PR #2594, opened 2026-03-21
feat: dart runtime flags for more aggressive heap management

gkc:
**- What I did**
feat: dart runtime flags for more aggressive heap management

**- How I did it**
Modified dockerfiles

**- How to verify it**
These flags have been tested extensively for several weeks, including in some heavily used canary and prod atServers; no adverse side effects observed

### DIFF
--- a/tools/run_locally/scripts/macos/at_server
+++ b/tools/run_locally/scripts/macos/at_server
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 function usage {
-  echo "Usage: $scriptName -a <@sign> -p <port> -s <secret> [-r]"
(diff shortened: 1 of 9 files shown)


tools/run_locally/scripts/macos/at_server
gkc, line 1: This is just a local dev utility

cpswan (changes requested):
All these LGTM, but I think we need the ENV in other docker files to ensure that things don't get submerged during promotion and also when using VE.

cconstab (dismissed):

cpswan (approved):

PR #2597, opened 2026-03-22
feat: tweak http handling, log DART_VM_OPTIONS on startup

gkc:
**- What I did**
- feat: http handling tweak for more consistent behaviour for access via proxy and direct access. Basically we'd like to use the same URL path in both cases
  - i.e. both this URL
    - https://proxy0001.atsign.org/@garycasey/gary/demo/blog/index.html
  - and this one
    - https://85d0d210-2841-59bf-a0c4-3a9a63dd42eb.canary.atsign.zone:11823/@garycasey/gary/demo/blog/index.html
  - should both work.
- feat: log the values of DART_VM_OPTIONS during atServer startup

**- How I did it**

**- How to verify it**

### DIFF
--- a/packages/at_secondary_server/CHANGELOG.md
+++ b/packages/at_secondary_server/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - feat: http handling tweak for more consistent behaviour for access via proxy 
   and direct access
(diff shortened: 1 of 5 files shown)


akafredperry (approved):
looks good

PR #2600, opened 2026-03-23
fix: can only have max eight flags set in DART_VM_OPTIONS, who knew

gkc:
**- What I did**
Lots of sighing

**- How I did it**
Fred kept me company

**- How to verify it**


akafredperry (approved):
all good

PR #2639, opened 2026-05-13
feat(deps): Take up at_commons ^5.10.0 to pick up new command (verb) syntax

gkc:
**- What I did**
feat(deps): Take up at_commons ^5.10.0 to pick up new command (verb) syntax

**- How I did it**
Bumped the at_commons dependency, ran dart pub upgrade to upgrade lock files

**- How to verify it**
Tests pass


gkc:
DNS is being weird on staging, causing the staging check to fail. Not an AtServer issue

cpswan (approved):

PR #2668, opened 2026-06-11
feat: first-class min/max/floor query surface on the persistence interfaces

gkc:
**- What I did**

Added a first-class min/max/floor query surface to the persistence layer, and made the atServer's key-expiry sweep event-driven on top of it:

- `KeyValueStore.nextExpiresAt()` / `peekExpired({asOf, limit})` — expiry wake-up and bounded ascending-order drain, on both the main and notification keystores
- `AtKeyValueStore.nextAvailableAt({asOf})` / `peekNewlyAvailable({since, asOf, limit})` — the TTB analogues, with a caller-side `(since, asOf]` watermark window
- `AtCommitLog.firstCommittedSequenceNumber()` — the log's floor, pairing with `lastCommittedSequenceNumber()` so sync clients can detect "the server no longer retains my delta"
- `AtSecondaryServerImpl`'s key-expiry sweep now sleeps until the next key actually expires instead of polling on a fixed cadence
- Fixed a latent bug: `HiveCommitLogKeyStore.removeAll` deleted box entries before reading them back to clean the commit-log cache, so the cleanup never ran; eviction is now captured pre-delete and guarded on commitId match
- Updated MIGRATION.md: documented the new surface and re-verified the at_client migration section against the current at_client code (the new methods replace at_client's hand-rolled scans of the private `getExpiryKeysCache()`)

**- How I did it**

- The Hive main keystore answers the four keystore methods from its existing in-memory expiry cache.
- The Hive notification keystore gains an in-memory effective-expiry index: `min(expiresAt, notificationDateTime + maxTtl)`, with entries that are already expired by shape (status `expired`, null `expiresAt`, null `notificationDateTime`) pinned to the epoch — so `effectiveExpiry < now` is exactly `AtNotification.isExpired()`. The index is maintained on every mutation path and also backs `getExpiredKeys()` without a full-box deserialise.
- The commit-log floor reads `getBox().keys.first`: the box key is the commitId and Hive iterates keys in ascending order, so the read is correct independent of any cache state.
- The expiry timer clamps its sleep to `[10s, expiringRunFreqMins]` plus jitter, so a put with a sooner expiry that lands mid-sleep is swept at worst one old-style poll interval late; worst case equals the previous fixed-cadence behaviour.[^1]

**- How to verify it**

- `dart test --concurrency=1` in `packages/at_persistence_secondary_server` (242 tests, including the new `test/min_max_first_class_test.dart`) and `packages/at_secondary_server` (773 tests)
- `tests/at_functional_test/runLocal.sh` — 157 tests, all passing

feat: first-class `nextExpiresAt`/`peekExpired`, `nextAvailableAt`/`peekNewlyAvailable`, and `firstCommittedSequenceNumber` on the persistence interfaces; the atServer key-expiry sweep is now event-driven.

[^1]: This staleness bound affects **disk hygiene only, not visibility**. Expired keys are already invisible to reads regardless of sweep timing — the lookup path filters them through `SecondaryUtil.isActiveKey`, and scan filters them too. A late sweep just means the expired entry occupies storage a little longer before physical deletion. If tighter reclamation is ever wanted, the `KeyValueStore.changes` stream provides the hook: subscribe, coalesce write bursts, and re-arm the timer from `nextExpiresAt()` — the same pattern at_client uses with its `dataEvents`-driven expiry timer. Deliberately not done here: the added moving part isn't justified while reclamation latency is harmless.


gkc:
e2e tests using the staging environment are failing for staging env reasons not to do with this PR

Xlin123 (approved):

PR #2673, opened 2026-06-15
feat: appMetadata support on update, update:meta and notify

gkc:
**- What I did**

Added `appMetadata` (provider-owned, server-opaque plaintext metadata, from at_commons) as a first-class metadata field that round-trips through every write and read/delivery path of the atServer:

- written via `update`, `update:meta` and `notify`
- persisted on `AtMetaData` (new field) and on stored notifications
- returned by `llookup:all` / `llookup:meta`
- preserved on the `cached:` copy created by `lookup` / `plookup` of another atSign's data
- delivered in monitor notifications (`metadata.appMetadata`)
- carried in the server-to-server notify command body
- emitted in `sync` responses
- retained across a subsequent update that doesn't mention it; malformed values rejected as invalid syntax

**- How I did it**

- `AtMetaData` gets an `AppMetadata? appMetadata` field, mapped in `fromCommonsMetadata`/`toCommonsMetadata`, `toJson`/`fromJson` (the latter accepts both the Map and base64(JSON) wire forms), `==` and `hashCode`.
- `AtMetaDataAdapter` persists it as a JSON-encoded string at Hive field index 26 — no new typeId. The adapter's count-prefixed indexed-field format makes this backward compatible by construction: records written before the field existed have no entry at index 26 and read back with `appMetadata == null`.
- The update path (shared by `update` and `update:meta`) decodes the base64(JSON) `:appMetadata:` fragment and, like other typed metadata fields (e.g. `pubKeyHash`), merges retain-only — an update that omits it keeps the stored value.
- `notify` stores it on the notification; `prepareNotifyCommandBody` appends it to the outbound command (it is the last group in the metadata fragment, and the body is built back-to-front, so it is prepended first).
- The per-delivery-path encoding follows each consumer's parser: nested Map for `llookup`/monitor, base64(JSON) for the server-to-server notify command and for `sync` responses (the only String form `Metadata.decodeAppMetadata` accepts — a `toString()` form would not parse client-side). `lookup`/`plookup` needed no handler change: the remote-lookup path round-trips through `AtData.fromJson` → `AtMetaData.fromJson`, which already accepts the Map form.
- Requires at_commons from the `xl-pluggable` branch (which carries `Metadata.appMetadata`, `AtConstants.appMetadata`, and the `appMetadata` group in `VerbSyntax`) via a dependency override until it is published.

**- How to verify it**

- Unit: `dart test --concurrency=1` in `packages/at_persistence_secondary_server` (incl. `test/app_metadata_persistence_test.dart` — model/adapter round-trip + a 26-field binary fixture for the backward-compat read) and `packages/at_secondary_server` (incl. `test/app_metadata_verb_test.dart` — every read/delivery path).
- Functional: `tests/at_functional_test/runLocal.sh` — 164 tests pass, incl. `test/app_metadata_test.dart` (update/update:meta/notify round-trips and the sync round-trip).

feat: `appMetadata` provider metadata is parsed, persisted and returned across `update`, `update:meta`, `notify`, `llookup`, `lookup`/`plookup`, monitor and `sync`.


packages/at_secondary_server/test/app_metadata_verb_test.dart
```diff
@@ -0,0 +1,325 @@
...
+      final metadata = await llookupAllMetadata('@bob:phone.wavi$alice');
+      expect(metadata.appMetadata, sampleAppMetadata());
+
+      // And the value was untouched.
+      await llookupHandler.process(
+          'llookup:@bob:phone.wavi$alice', inboundConnection);
+      expect(inboundConnection.lastWrittenData, startsWith('data:hello'));
+    });
+
+    test('malformed appMetadata is rejected as invalid syntax', () async {
```
Xlin123, line 157: can we have a malformed appMetadata test for notify as well?

I want to pin down any functionality related to AppMetadata that _might break_ in the future.

packages/at_secondary_server/test/app_metadata_verb_test.dart
```diff
@@ -0,0 +1,325 @@
...
+    test('update with appMetadata; llookup:all returns it', () async {
+      await updateHandler.process(
+          'update:appMetadata:${encoded()}:@bob:phone.wavi$alice hello',
+          inboundConnection);
+
+      final metadata = await llookupAllMetadata('@bob:phone.wavi$alice');
+      expect(metadata.appMetadata, sampleAppMetadata());
+    });
+
+    test('a subsequent update without appMetadata retains it', () async {
```
Xlin123, line 104: can we add a 'subsequent update:meta' to ensure it retains when changing other metadata fields?

Xlin123 (changes requested):

packages/at_secondary_server/test/app_metadata_verb_test.dart
gkc: Added in 8bfb8953 — `a subsequent update:meta changing other fields retains appMetadata`: it establishes appMetadata via `update`, then runs an `update:meta` setting `ttl` and `encKeyName` (no appMetadata), and asserts both changed fields took effect AND appMetadata was retained.

packages/at_secondary_server/test/app_metadata_verb_test.dart
gkc: Added in 8bfb8953 — `notify with malformed appMetadata is rejected as invalid syntax`, mirroring the existing update-path test (a non-base64 value is rejected as `InvalidSyntaxException` at the decode step).

Xlin123 (approved):
thanks @gkc looks great

srieteja (approved):

PR #2683, opened 2026-06-25
feat: IS-1 post-quantum inter-server authentication

### LINKED ISSUE #2049: IS-1 — inter-server PQ authentication: swap FROM/POL challenge signature RSA → ML-DSA-65
Part of #1889. Plan project **IS-1**. Design in `design.md` section 8 on `gkc-pq-d1-spike`.

The server-to-server FROM/POL handshake signs a random per-session UUID challenge with RSA-2048. FROM/POL is authentication, not confidentiality: TLS secures the channel and the UUID gives freshness, so the signature is the only Shor-vulnerable part and swapping it is the whole change. Scope was pared back from a hybrid KEM plus certificate design on 2026-07-21, since a recorded signature is worthless once its key retires, and the adversary is an active quantum computer forging a live auth.

Not built, and not to be: X-Wing KEM or any encaps/decaps, an inter-server certificate, key expiry and rotation grace, an HKDF key-confirmation tag, a `PqKeyManager` lifecycle class.

## What was built

Public keys go to `public:signing_publickeys@<atSign>`. The name is plural and generic, naming neither an algorithm nor an era, so a new algorithm adds an entry rather than a record. Top-level keys are challenge types and entries are objects, leaving room for parameters later: `{"ml-dsa-65": {"publicKey": "<base64>"}}`. Readers ignore unknown keys and unknown fields.

The record carries key material only. Which types a verifier accepts travels in-band inside the signed pol challenge, where it's tamper-evident, so RSA needs no entry at all and stays at the singular `signing_publickey` on `PolVerbHandler`'s legacy branch.

Secret halves live at `local:signing_secretkey.<type>@<atSign>`, one per type so a migration can hold two. `local:` is never commit-logged, so never synced and not addressable.

The pol cookie is `<type>:<base64 signature>`. `FromVerbHandler` fetches the peer's record and `chooseNegotiatedAlgo` picks the strongest entry the verifier also holds, walking the verifier's own preference order so a peer can't steer the choice by reordering its record.

`pq.disablePqAuth` turns PQ auth off and withdraws the published keys. Self-auth is always UUID, and a peer with no PQ key falls back to UUID/RSA.

## What's left

Merge [at_server#2683](https://github.com/atsign-foundation/at_server/pull/2683), still open at 2026-08-18 and wanting a re-review after the pare-back.

## Acceptance

Both the PQ path and the RSA fallback. `signing_publickeys` looked up live every handshake. Pure-Dart fallback with `AT_CHOPS_LIBCRYPTO_PATH` unset. Mixed fleet. Existing delete and update verb tests still pass, with the protected-key count updated.

*(This description previously named a record `pq_signing_publickey@<atSign>` that was never built. Corrected against the code 2026-08-17.)*

srieteja:
Closes https://github.com/atsign-foundation/at_client_sdk/issues/2049

### What I did

Secondary servers now authenticate each other over FROM/POL using an ML-DSA-65 signature challenge-response, replacing the RSA-signature challenge — for any peer that has published a PQ signing key. Peers without one still get the legacy UUID/RSA challenge, so rollout is backward-compatible both ways, and a config kill switch can force everyone onto the legacy path.

### How I did it

- **Key material** (`pq_key_manager.dart`, `pq_constants.dart`)
   - `PqKeyManager` generates an ML-DSA-65 signing keypair on startup and publishes the public half at `public:pq_signing_publickey@atsign` (JSON, keyed by algorithm id for crypto agility). Secret material stays under `local:pq_*` (never synced, not verb-reachable).
   - Publishing is folded into `init()` — a keypair isn't "initialised" until it's also published — and skipped when an unchanged key is already published, so an ordinary restart doesn't push a no-op sync delta to every enrolled client.

- **The handshake** (`from_verb_handler.dart`, `pol_verb_handler.dart`, `outbound_client.dart`)
   - `FROM` issues the same verifier-bound pol challenge for all peer auth (`SecondaryUtil.buildBoundPolChallenge`, 3.15.1's "augmented pol challenge"), so PQ needs no separate binding step — the binding already applies uniformly to both the PQ and legacy RSA paths.
   - A prover signs with ML-DSA-65 only when its own PQ keypair is initialised *and* the peer has published a PQ key (`checkPeerPqSupport`, an unauthenticated `lookup`), else it falls back to legacy RSA — keeping a mixed-version fleet working during a rolling upgrade.
   - Malformed/oversized peer key material fails as an auth error.

- **Config & protection** (`at_secondary_config.dart`, `at_secondary_impl.dart`, `abstract_update_verb_handler.dart`, `delete_verb_handler.dart`, `handler_util.dart`)
   - `pq.disablePqAuth` in `config.yaml` (default `false`) is a global kill switch. `AtSecondaryServerImpl.initializePqAuth` withdraws any previously-published key on `disablePqAuth` or an init failure, since a peer that's already seen one has no mid-handshake fallback.
   - `update`/`delete` refuse to touch `public:pq_signing_publickey@atsign` (the only PQ record that's verb-addressable), via a shared `expandProtectedKeyTemplates` helper.

### How to verify it
- All the existing and newly added tests pass.

### Description for the changelog
feat: post-quantum (ML-DSA-65) inter-server FROM/POL authentication

---
**PQ plan conformance (§10(e)):** project **IS-1** — [implementation-plan.md](https://github.com/atsign-foundation/at_client_sdk/blob/trunk/docs/projects/pq/implementation-plan.md), tracking epic atsign-foundation/at_client_sdk#1889.


packages/at_secondary_server/lib/src/crypto/pq_key_fetch.dart
```diff
@@ -0,0 +1,40 @@
...
+/// Every handshake live-fetches fresh key material so a peer's key rotation is
+/// always honoured and there is no stale-cert failure mode.
+///
+/// Returns the fetched material, or null if the peer publishes no PQ keys
+/// (caller falls back to the legacy RSA/UUID path). Never throws — failures are
+/// logged and surface as null.
+Future<FetchedPqKeys?> fetchPqKeys(
+    OutboundClient oc, String fromAtSign) async {
+  try {
+    final certRaw = (await oc.plookUp('pq_xwing_cert$fromAtSign'))
```
srieteja: should not explicitly refer to key names e.g. pq_xwing_cert@atsign.

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
```diff
@@ -0,0 +1,234 @@
...
+import 'package:at_utils/at_logger.dart';
+import 'package:meta/meta.dart';
+
+/// Manages PQ keypairs for inter-server authentication.
+///
+/// Two-tier hierarchy:
+///   ML-DSA-65 root key  → long-lived trust anchor, signs X-Wing certs
+///   X-Wing operational  → used for KEM-based proof-of-possession in FROM/POL
+class PqKeyManager {
+  static final PqKeyManager _instance = PqKeyManager._();
```
srieteja: why singleton?

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
```diff
@@ -0,0 +1,234 @@
...
+
+  /// Returns a fresh, uninitialised instance — for use in tests only.
+  @visibleForTesting
+  static PqKeyManager testInstance() => PqKeyManager._();
+
+  static const _certValidityDays = 365;
+  static const _certGraceDays = 30;
+  static const _pqSigningPrivKeyName = 'pq_signing_secretkey';
+  static const _pqSigningPubKeyName = 'pq_signing_publickey';
+  static const _pqXwingSecKeyName = 'pq_xwing_secretkey';
```
srieteja: Unless these are server only, they should reside in at_commons.

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
```diff
@@ -0,0 +1,234 @@
...
+  static final PqKeyManager _instance = PqKeyManager._();
+  static PqKeyManager get instance => _instance;
+  PqKeyManager._();
+
+  /// Returns a fresh, uninitialised instance — for use in tests only.
+  @visibleForTesting
+  static PqKeyManager testInstance() => PqKeyManager._();
+
+  static const _certValidityDays = 365;
+  static const _certGraceDays = 30;
```
srieteja: the two configs above should ideally be in config.yaml and should be read through AtSecondaryConfig

gkc:
Per the PQ plan's §10(e) conformance rule, this PR is project **IS-1** (see [implementation-plan.md](https://github.com/atsign-foundation/at_client_sdk/blob/trunk/docs/projects/pq/implementation-plan.md) and tracking epic atsign-foundation/at_client_sdk#1889).

packages/at_secondary_server/lib/src/crypto/pq_key_fetch.dart
```diff
@@ -0,0 +1,75 @@
...
+
+/// A remote atSign's published PQ key material.
+class FetchedPqKeys {
+  /// Raw JSON of the remote's X-Wing cert (`pq_xwing_cert@<atSign>`), which
+  /// also carries the signing ML-DSA-65 public key — see [XWingCert].
+  final String certRaw;
+
+  FetchedPqKeys(this.certRaw);
+}
+
```
srieteja: the negative cache feels like overkill.

srieteja: Why is this even needed?

packages/at_secondary_server/lib/src/crypto/pq_key_fetch.dart
srieteja: These should either be put in at_commons if they're used outside at_server. Or someplace in at_server if only limtied to at_server. Follow the current convention.

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
```diff
@@ -0,0 +1,311 @@
...
+        _mlDsaPublicKey = base64.decode(existingMlDsaPub);
+        _xwingSecretKey = base64.decode(existingXwingSec);
+        _xwingPublicKey = base64.decode(existingXwingPub);
+        final existingXwingPrev = await _tryGet(keyStore, xwingPrevSecKeyId);
+        if (existingXwingPrev != null) {
+          _xwingPrevSecretKey = base64.decode(existingXwingPrev);
+        }
+        _log.info('Loaded existing PQ keypairs for $atSign');
+      } else {
+        if (existingMlDsaSec != null || existingXwingSec != null) {
```
srieteja: heal/remove previous X-Wing key too if only public/private survives

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
```diff
@@ -0,0 +1,311 @@
...
+    _assertInitialized();
+    final certKeyId = 'public:$_pqXwingCertName$atSign';
+    final certPrevKeyId = 'public:$_pqXwingCertPrevName$atSign';
+    final xwingPrevSecKeyId = '$atSign:$_pqXwingSecKeyPrevName$atSign';
+
+    final prevCertRaw = await _tryGet(keyStore, certPrevKeyId);
+    if (prevCertRaw != null) {
+      final parsed = XWingCert.tryParse(prevCertRaw);
+      if (parsed != null &&
+          parsed.validUntil.isBefore(DateTime.now()
```
srieteja: Is this the right place to do this ?

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    final existingCert = await _tryGet(keyStore, certKeyId);
+    if (existingCert != null) {
+      final parsed = XWingCert.tryParse(existingCert);
+      if (parsed != null &&
+          parsed.validUntil.isAfter(DateTime.now()
+              .toUtc()
+              .add(const Duration(days: _certGraceDays))) &&
+          _bytesEqual(parsed.xwingPublicKey, _xwingPublicKey)) {
```
srieteja: This needs to be moved to the prevCert key as the new one is going to replace thisl.

srieteja: I see some overlap between this method and the rotateCert().

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
```diff
@@ -0,0 +1,311 @@
...
+      await _put(keyStore, certPrevKeyId, currentCert, null);
+    }
+
+    _xwingPrevSecretKey = _xwingSecretKey;
+    await _put(
+        keyStore, xwingPrevSecKeyId, base64.encode(_xwingPrevSecretKey!), null);
+
+    final newKp = await XWingKeyPair.generate();
+    _xwingSecretKey = base64.decode(newKp.atPrivateKey.privateKey);
+    _xwingPublicKey = base64.decode(newKp.atPublicKey.publicKey);
```
srieteja: base64 decoded keys can now be directly fetched from the key instance.

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
```diff
@@ -0,0 +1,311 @@
...
+    final cert = await _buildCert();
+    await _put(keyStore, certKeyId, cert.toJson(), null);
+    _log.info('Rotated X-Wing cert for $atSign');
+  }
+
+  /// Encapsulate against an arbitrary X-Wing public key.
+  ///
+  /// Returns the encapsulation record with [ciphertext] and [sharedSecret].
+  static Future<({Uint8List ciphertext, Uint8List sharedSecret})> encapsulate(
+          Uint8List xwingPublicKey) async =>
```
srieteja: This is not needed, the callers can directly call AtPqc

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
```diff
@@ -0,0 +1,311 @@
...
+        .signBytes(tmp.signedBytes, secretKey: _mlDsaSecretKey);
+    return XWingCert(
+        xwingPublicKey: _xwingPublicKey,
+        validUntil: validUntil,
+        signature: sig,
+        mlDsaPublicKey: _mlDsaPublicKey);
+  }
+
+  /// Decapsulate an X-Wing [ciphertext] → shared secret (32 bytes).
+  Future<Uint8List> decaps(Uint8List ciphertext) async {
```
srieteja: same as encaps

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  /// Verify [certJson]. The ML-DSA-65 public key is carried inside the cert
+  /// itself (see [XWingCert.mlDsaPublicKey]) — a single `plookUp` of the cert
+  /// record is sufficient; there is no separate signing-key fetch. Returns
+  /// the parsed cert if the signature verifies and it is not expired; null
+  /// otherwise.
+  Future<XWingCert?> verifyCert(String certJson) async {
+    final cert = XWingCert.tryParse(certJson);
+    if (cert == null) return null;
```
srieteja: This can't be null as the method param accepts String and not String?, so very unlikely we'll need a nullable return type.

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
```diff
@@ -0,0 +1,311 @@
...
+    if (cert.validUntil.isBefore(DateTime.now().toUtc())) return null;
+    final ok = await AtPqc.mlDsa65.verifyBytes(cert.signedBytes,
+        signature: cert.signature, publicKey: cert.mlDsaPublicKey);
```
srieteja: The grace period is not taken into consideration.

srieteja: The grace period might need to be included in the cert itself.

packages/at_secondary_server/lib/src/crypto/x_wing_cert.dart
```diff
@@ -0,0 +1,58 @@
...
+  const XWingCert({
+    required this.xwingPublicKey,
+    required this.validUntil,
+    required this.signature,
+    required this.mlDsaPublicKey,
+  });
+
+  /// Canonical bytes the signer covers:
+  /// `xwingPublicKey || utf8(validUntil ISO-8601 UTC) || mlDsaPublicKey`.
+  Uint8List get signedBytes => Uint8List.fromList([
```
srieteja: signedBytes? What is signed here?

packages/at_secondary_server/lib/src/server/at_secondary_config.dart
```diff
@@ -118,6 +123,27 @@ class AtSecondaryConfig {
 
   static final Map<String, String> _envVars = Platform.environment;
 
+  // ignore: prefer_final_fields — mutated by setTestEnvVar
+  static Map<String, String>? _testEnvOverrides;
+
```
srieteja: I don't like this.

packages/at_secondary_server/lib/src/verb/handler/from_verb_handler.dart
```diff
@@ -103,12 +145,79 @@ class FromVerbHandler extends AbstractVerbHandler {
...
+  Future<String?> _tryBuildPqProof(
+      String fromAtSign, String sessionID, String storedSecretId) async {
+    try {
+      // Always fetch the peer's published PQ key material live (unauthenticated
+      // plookUp — the same channel the legacy RSA path uses). Nothing is
+      // cached: each handshake honours the peer's current keys, so there is no
+      // stale-cert failure mode. Falls through to the UUID/RSA path when no
+      // outbound client is wired (e.g. unit tests) or the peer publishes no
+      // PQ cert.
+      if (outboundClientManager == null) return null;
```
srieteja: Fallback if outboundClient is null? Also needs to be logged.

packages/at_secondary_server/lib/src/verb/handler/from_verb_handler.dart
```diff
@@ -103,12 +145,79 @@ class FromVerbHandler extends AbstractVerbHandler {
...
+      final fetched = await _fetchPeerPqKeys(fromAtSign);
+      if (fetched == null) return null;
+      final String certRaw = fetched.certRaw;
+
```
srieteja: How come fetch is done here, but the verification is done in PqKeyManager? Needs symmetry.

packages/at_secondary_server/lib/src/verb/handler/from_verb_handler.dart
```diff
@@ -103,12 +145,79 @@ class FromVerbHandler extends AbstractVerbHandler {
...
+      final cert = await PqKeyManager.instance.verifyCert(certRaw);
+      if (cert == null) return null;
+
+      final result = await PqKeyManager.encapsulate(cert.xwingPublicKey);
```
srieteja: Call AtPqc.xwing.encaps directly.

packages/at_secondary_server/lib/src/verb/handler/pol_verb_handler.dart
```diff
@@ -22,11 +22,11 @@ class PolVerbHandler extends AbstractVerbHandler {
 
   final OutboundClientManager outboundClientManager;
   final AtCacheManager cacheManager;
-  final AtAccessLog accessLog;
+  final AtAccessLog? accessLog;
   final _dummyInboundConnection = DummyInboundConnection();
 
   PolVerbHandler(super.keyStore, this.outboundClientManager, this.cacheManager,
-      {required this.accessLog});
+      {this.accessLog});
```
srieteja: why has the required been removed?

packages/at_secondary_server/test/crypto/pq_key_manager_test.dart
```diff
@@ -0,0 +1,474 @@
...
+      final cert = XWingCert(
+          xwingPublicKey: xwingPub,
+          validUntil: validUntil,
+          signature: sig,
+          mlDsaPublicKey: mlDsaPub);
+
+      final result =
+          await PqKeyManager.instance.verifyCert(cert.toJson());
+      expect(result, isNotNull);
+      expect(result!.xwingPublicKey, equals(xwingPub));
```
srieteja: validate the rest of the params in the cert

packages/at_secondary_server/test/crypto/pq_key_manager_test.dart
```diff
@@ -0,0 +1,474 @@
...
+      final mlDsaSk = base64.decode(mlDsaKp.atPrivateKey.privateKey);
+      final xwingKp = await XWingKeyPair.generate();
+      final xwingPub = base64.decode(xwingKp.atPublicKey.publicKey);
+      final validUntil =
+          DateTime.now().toUtc().subtract(const Duration(hours: 1));
+
+      final draft = XWingCert(
+          xwingPublicKey: xwingPub,
+          validUntil: validUntil,
+          signature: Uint8List(0),
```
srieteja: buildCert in PqKeyManager should be made public and this test should re-use that instead of doing this dance over and over again.

packages/at_secondary_server/test/crypto/pq_key_manager_test.dart
```diff
@@ -0,0 +1,474 @@
...
+      final sig = await AtPqc.mlDsa65
+          .signBytes(draft.signedBytes, secretKey: mlDsaSk);
+      final cert = XWingCert(
+          xwingPublicKey: xwingPub,
+          validUntil: validUntil,
+          signature: sig,
+          mlDsaPublicKey: mlDsaPub);
+
+      final result =
+          await PqKeyManager.instance.verifyCert(cert.toJson());
```
srieteja: Also why does verifyCert return the cert itself rather than returning true or false?

packages/at_secondary_server/test/crypto/pq_key_manager_test.dart
```diff
@@ -0,0 +1,474 @@
...
+      expect(result, isNull);
+    });
+  });
+
+  group('PqKeyManager KEM round-trip', () {
+    test('encapsulate/decapsulate produces matching shared secret', () async {
+      final kp = await XWingKeyPair.generate();
+      final pub = base64.decode(kp.atPublicKey.publicKey);
+      final sk = base64.decode(kp.atPrivateKey.privateKey);
```
srieteja: This is an atChops test, remove from here.

packages/at_secondary_server/test/crypto/pq_key_manager_test.dart
```diff
@@ -0,0 +1,474 @@
...
+        () async {
+      final mgr = PqKeyManager.testInstance();
+      await mgr.init(atSign, keyStore);
+      final xwingPubBefore = mgr.xwingPublicKey;
+
+      // Simulate the public X-Wing key record being deleted (owner-deletable)
+      // while the secret key survives.
+      await keyStore.remove('public:pq_xwing_secretkey_pub$atSign');
+
+      final mgr2 = PqKeyManager.testInstance();
```
srieteja: The test instance architecture is really weird.

srieteja: The test itself is fine tho.

packages/at_secondary_server/test/crypto/pq_key_manager_test.dart
```diff
@@ -0,0 +1,474 @@
...
+
+      // Encapsulate against current X-Wing public key (before rotation).
+      final kemResult =
+          await PqKeyManager.encapsulate(mgr.xwingPublicKey);
+
+      await mgr.rotateCert(atSign, keyStore);
+
+      final prevSecretKey = mgr.xwingPrevSecretKey;
+      expect(prevSecretKey, isNotNull,
+          reason: 'Manager must retain prev secret key after rotation');
```
srieteja: Also assert that the key before rotateCert and the previous key after are the same.

packages/at_secondary_server/test/crypto/pq_key_manager_test.dart
```diff
@@ -0,0 +1,474 @@
...
+      await keyStore.put('public:pq_xwing_cert_prev$atSign', prevData);
+
+      await mgr.publishKeys(atSign, keyStore);
+
+      expect(await keyStore.exists('public:pq_xwing_cert_prev$atSign'), isFalse,
+          reason: 'Expired prev cert must be deleted after grace period');
+      expect(
+          await keyStore.exists('$atSign:pq_xwing_secretkey_prev$atSign'),
+          isFalse,
+          reason: 'Expired prev secret key must be deleted alongside '
```
srieteja: assert that a new cert is generated.

packages/at_secondary_server/test/crypto/pq_key_manager_test.dart
```diff
@@ -0,0 +1,474 @@
...
+        signature: Uint8List(0),
+        mlDsaPublicKey: mlDsaPub,
+      );
+      final prevData = AtData()..data = recentCert.toJson();
+      await keyStore.put('public:pq_xwing_cert_prev$atSign', prevData);
+
+      await mgr.publishKeys(atSign, keyStore);
+
+      expect(await keyStore.exists('public:pq_xwing_cert_prev$atSign'), isTrue,
+          reason: 'Prev cert within grace period must NOT be deleted');
```
srieteja: Assert that new cert is generated.

packages/at_secondary_server/test/from_verb_handler_pq_test.dart
```diff
@@ -0,0 +1,432 @@
...
+      final sig = await AtPqc.mlDsa65
+          .signBytes(tmp.signedBytes, secretKey: mlDsaSk);
+      final cert = XWingCert(
+          xwingPublicKey: xwingPub,
+          validUntil: validUntil,
+          signature: sig,
+          mlDsaPublicKey: mlDsaPub);
+
+      // Mock the outbound client so plookUp returns Bob's published PQ cert.
+      final mockMgr = _pqMockMgr(cert.toJson());
```
srieteja: _pqMockMgr name is confusing. The name should describe whjat it actually is

packages/at_secondary_server/lib/src/crypto/x_wing_cert.dart
```diff
@@ -1,12 +1,13 @@
...
 
+import 'package:at_chops/at_chops_ffi.dart';
+
 /// An ML-DSA-65-signed binding of an X-Wing public key to an expiry timestamp.
 ///
-/// Pure data class — all crypto (sign/verify) is performed by the caller via a
-/// resolved [AtSigningAlgorithm]. Lives here (not in at_chops) because the cert
-/// is an atServer↔atServer protocol artifact with a single consumer:
-/// [PqKeyManager]. See docs/inter-atserver-comms.md.
+/// Lives here (not in at_chops) because the cert is an atServer↔atServer
```
srieteja: should not refer/reason about at_chops here

packages/at_secondary_server/lib/src/connection/outbound/outbound_client.dart
```diff
@@ -188,14 +190,54 @@ class OutboundClient {
...
+          // Two candidate tags are stored (current + prev key) when a rotation
+          // grace-period key is retained, since X-Wing implicit rejection gives
+          // no signal for which key the peer actually encapsulated against.
+          final tag =
+              deriveConfirmationTag(decapsResult.current, sessionIdWithAtSign);
+          var tagsToStore = base64Url.encode(tag);
+          if (decapsResult.prev != null) {
+            final prevTag =
+                deriveConfirmationTag(decapsResult.prev!, sessionIdWithAtSign);
+            tagsToStore = '$tagsToStore,${base64Url.encode(prevTag)}';
```
srieteja: Stores the tag(s) generated from the current x-wing keypair and the previous x-wing keypair (if any) that qualify under the grace period.

packages/at_secondary_server/lib/src/crypto/x_wing_cert.dart
srieteja: renamed to tbsBytes (to be signed bytes)

packages/at_secondary_server/test/cram_verb_test.dart
```diff
@@ -20,6 +20,7 @@ import 'test_utils.dart';
 void main() {
   late AtKeyValueStore<String, AtData, AtMetaData?> mockKeyStore;
   late FakeSocket mockSocket;
+  late MockOutboundClientManager mockOcm;
```
srieteja, line 23: This file only changes existing FromVerbHandler instantiations to match the new signature.

packages/at_secondary_server/lib/src/crypto/pq_key_fetch.dart
```diff
@@ -0,0 +1,38 @@
...
+    // remote side prepends 'public:' itself. pqXwingCertName() returns the
+    // public:-prefixed *storage* name used for keyStore.put/get, which is
+    // the wrong shape here.
+    return (await oc.plookUp('$pqXwingCertNamePart$fromAtSign'))
+        ?.replaceFirst(_dataPrefix, '');
+  } catch (e) {
+    _log.finer('Live PQ cert fetch failed for $fromAtSign: $e');
+    return null;
+  }
+}
```
gkc: It is worth noting that talking to a legacy atServer will result every time in a null so every connection will pay the cost of a round trip plookup before falling back to fetching the legacy key. But we are going to accept that cost because it only applies during rollout so it's not worth the effort building a ttl cache or whatever.

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
```diff
@@ -0,0 +1,324 @@
...
+import 'package:at_utils/at_logger.dart';
+import 'package:meta/meta.dart';
+
+/// Manages PQ keypairs for inter-server authentication.
+///
+/// Two-tier hierarchy:
+///   ML-DSA-65 root key  → long-lived trust anchor, signs X-Wing certs
+///   X-Wing operational  → used for KEM-based proof-of-possession in FROM/POL
+class PqKeyManager {
+  static final PqKeyManager instance = PqKeyManager();
```
gkc: No more singletons please 🙏🏼

- remove this `instance`;
- AtSecondaryServerImpl should
  - construct imediately (add a `final PqKeyManager pqKeyManager = PqKeyManager();`)
  - modify OutboundClientManager and NotificationManager constructors, inject the pqKeyManager
  - at initialization time, use the `pqKeyManager` instead of `PqKeyManager.instance`
- OutboundClientManager and NotificationManager need to inject the pqKeyManager in the OutboundClients they create
- and the OutboundClient gets the pqKeyManager added to its constructor

None of the pqKeyManager constructor parameters should be optional, we need to drive it all the way down.


Why does this matter? We've still got global state - but it's now very obvious via constructor parameters what is going on (no hidden spooky-action-at-a-distance dependency on some rando singleton). Also gives us more flexibility in unit testing. (More on that later)

packages/at_secondary_server/lib/src/connection/outbound/outbound_client.dart
```diff
@@ -302,14 +303,54 @@ class OutboundClient {
...
+      }
       var sessionIdWithAtSign = cookieParams[2];
       var challenge = cookieParams[3];
 
       if (productionMode) {
-        var signedChallenge = SecondaryUtil.signChallenge(
-            challenge, AtSecondaryServerImpl.getInstance().signingKey);
-        await SecondaryUtil.saveCookie(sessionIdWithAtSign, signedChallenge,
-            AtSecondaryServerImpl.getInstance().keyValueStore);
+        if (challenge == 'pq') {
```
gkc: Please review what you've built bearing in mind that we need to have some crypto agility here; this isn't one and done - the algorithms we use will change over time in response to cryptographic developments; and we will likely in the future have a mixed fleet where many of the atServers are not hosted under our control and are therefore being upgraded at different rates.

What this means in practice:
- stuff that is stored (and fetched) needs to be self-describing to an extent, and again over time we may be storing many (not just one) set of stuff. Like, I might have to retain my legacy key until the atServer I am connecting to is "pq-capable". And later, I might have to retain the key-stuff in this PR to support atServers that only support this, while also writing new key-stuff required by new algos for atServers which understsand them

I'd suggest we assume ONE public record to store (and fetch), forever, and keep it generically named. The value should be json, and should hold many things. The verifying side would fetch it, see what's offered, and use the "best" according to its knowledge. (Just like now we're saying "oh, pq is available, we'll use that") So if "pq_567" stuff is available in the record as well as "pq" then the verifier can go "oh, pq_567 is stronger than pq, we'll use that.

For simplicity, I would suggest that the top-level keys in the json map should be the same as the 'challenge type'

Hope all of that makes sense, talk to me on gchat if not

gkc: In addition to the above, I would take this opportunity to extract all of the parse-validate-store-cookie stuff into their own functions now, marked visibleForTesting, so that we can unit test all of this logic more simply.

packages/at_secondary_server/lib/src/crypto/pq_key_fetch.dart
gkc: See also my comments in OutboundClient about giving us crypto agility so we can gracefully cope with evolution of the algos used over time

packages/at_secondary_server/lib/src/verb/handler/abstract_update_verb_handler.dart
```diff
@@ -70,6 +76,24 @@ abstract class AbstractUpdateVerbHandler extends ChangeVerbHandler {
...
       ' is not authorized to update key: $key';
 
+  /// Protected-key templates for the bare-update guard in
+  /// [preProcessAndNotify], expanded for [atSign] and lowercased, minus
+  /// `publickey`: unlike the other entries in
+  /// [AtSecondaryConfig.protectedKeys] (signing keys, the PKAM public key,
+  /// the PQ cert), the encryption public key is client-writable by design —
+  /// an at_client publishes/rotates its own `publickey` via update during
+  /// onboarding/key-rotation — so it's only protected from delete, not
+  /// update.
```
gkc: Have you added unit tests for the above?

packages/at_secondary_server/lib/src/verb/handler/delete_verb_handler.dart
```diff
@@ -184,19 +184,14 @@ class DeleteVerbHandler extends ChangeVerbHandler {
...
+        .toSet();
   }
 
   bool _isProtectedKey(String key, {String? isCached}) {
     isCached ??= 'false';
-    if (protectedKeys!.contains(key) && isCached == 'false') {
+    if (protectedKeys!.contains(key.toLowerCase()) && isCached == 'false') {
       logger.severe('Cannot delete key. \'$key\' is a protected key');
       return true;
     }
```
gkc, line 198: Good time for unit tests for this path also if not already covered

packages/at_secondary_server/lib/src/verb/handler/from_verb_handler.dart
```diff
@@ -86,29 +108,94 @@ class FromVerbHandler extends AbstractVerbHandler {
...
     } on DataStoreException catch (e) {
       logger.severe('Hive error adding to access log:${e.toString()}');
     }
   }
 
+  /// Returns `(wireProof, storedSecret)` — a `pq:<base64Url(ciphertext)>`
+  /// wire proof paired with the `pq:<key-confirmation-tag>` to persist — if
+  /// the remote's X-Wing cert can be live-fetched and passes ML-DSA-65
+  /// verification. Returns null to fall back to the legacy UUID flow — every
+  /// abandon branch is logged so a stuck-on-legacy peer is diagnosable.
```
gkc: again think about crypto agility in this function

packages/at_secondary_server/lib/src/crypto/pq_key_fetch.dart
gkc: Thinking about it though, maybe we should actually consider building a (very short) ttl cache so that when we DO fetch successfully, we don't have to refetch for, say, 3 minutes.

packages/at_secondary_server/pubspec.yaml
```diff
@@ -43,5 +43,14 @@ dev_dependencies:
...
 dependency_overrides:
+  at_chops:
+    git:
+      url: https://github.com/atsign-foundation/at_client_sdk.git
+      # PR #2030 (at_chops FFI / AtPqc) has merged to trunk, so st/at_chops_ffi
+      # is gone. Pinned to a commit, not the branch name — floating on trunk
+      # makes builds non-reproducible as it moves underneath us.
+      ref: 3a9302edda6df9a8f25c62edce8442ef34e590e7
+      path: packages/at_chops
+  pointycastle: ^4.0.0
```
gkc: Even though we don't directly depend on pointycastle, I'm guessing this is required because we do still directly depend on the encrypt and crypto packages. We should make a follow-up PR to remove dependencies on those packages and get everything we need from at_chops. (There aren't that many call sites, so it should be an easy PR because I think all of the stuff we need is already in at_chops.)

packages/at_secondary_server/pubspec.yaml
```diff
@@ -43,5 +43,14 @@ dev_dependencies:
   mocktail: ^1.0.5
 
 dependency_overrides:
+  at_chops:
+    git:
+      url: https://github.com/atsign-foundation/at_client_sdk.git
+      # PR #2030 (at_chops FFI / AtPqc) has merged to trunk, so st/at_chops_ffi
+      # is gone. Pinned to a commit, not the branch name — floating on trunk
+      # makes builds non-reproducible as it moves underneath us.
+      ref: 3a9302edda6df9a8f25c62edce8442ef34e590e7
```
gkc: What's holding us back from publishing at_chops 3.4.0?

packages/at_secondary_server/lib/src/crypto/x_wing_cert.dart
```diff
@@ -0,0 +1,67 @@
+import 'dart:convert';
+import 'dart:typed_data';
+
+import 'package:at_chops/at_chops_ffi.dart';
+
+/// An ML-DSA-65-signed binding of an X-Wing public key to an expiry timestamp.
+///
+/// Lives here because the cert is an atServer↔atServer
+/// protocol artifact with a single consumer: [PqKeyManager]. See
+/// docs/inter-atserver-comms.md.
```
gkc: No such doc.

Regardless of being currently used only here, I think this should live in at_chops according our stated package responsibility scoping

gkc (changes requested):
This is really good stuff @srieteja and at the "does it do the job" level it LGTM. But
- it introduces a new singleton
- there's no consideration of crypto agility
- and a few other nits which are also important

See my review comments

gkc:
@srieteja After staring at this for a while longer, I think that we have over-built, substantially.

All we really need is 
- Boot: generate an ML-DSA-65 keypair (the same key you're already generating — minus the X-Wing one). Publish pq_signing_publickey@A as a protected record. (And, for **crypto agility**, publish it as as JSON structure which initially will contain one field - 'ml-dsa-65')
- FROM: B issues the existing random UUID challenge, stores it — unchanged.
- A's outbound: sign the challenge with ML-DSA instead of RSA, save the cookie — a one-line algorithm swap in the existing branch.
- POL: fetch A's ML-DSA public key + signed challenge, AtPqc.mlDsa65.verify(...) instead of RSAPublicKey.verifySHA256Signature(...) — a one-line swap.

No X-Wing, no cert, no expiry, no rotation, no grace period, no confirmation tags, no decaps-fallback, no dual-candidate logic. Same PQ authentication guarantee. It even reuses the existing per-session UUID, so the freshness/replay property is already handled by code that's staying.

gkc:
@srieteja to be clear - this was on me ... the spec that I wrote for this was overbuilt. My apologies for not thinking this through properly before you started the work

packages/at_secondary_server/test/from_verb_test.dart
```diff
@@ -76,20 +76,20 @@ void main() async {
     test('test from accept', () {
       var command = 'from:$alice';
```
srieteja, line 77: This file only changes existing FromVerbHandler instantiations to match the new signature.

packages/at_secondary_server/test/local_lookup_verb_test.dart
```diff
@@ -156,7 +156,7 @@ void main() {
           'b26455a907582760ebf35bc4847de549bc41c24b25c8b1c58d5964f7b4f8a43bc55b0e9a601c9a9657d9a8b8bbc32f88b4e38ffaca03c8710ebae1b14ca9f364';
       await keyValueStore.put('privatekey:at_secret', secretData);
```
srieteja, line 157: This file only changes existing FromVerbHandler instantiations to match the new signature.

packages/at_secondary_server/lib/src/verb/handler/delete_verb_handler.dart
srieteja: Already covered

packages/at_secondary_server/lib/src/crypto/pq_key_manager.dart
srieteja: Addressed through [4fdfd05](https://github.com/atsign-foundation/at_server/pull/2683/commits/4fdfd056146dffd4a4dd974883f744e87aaf14f2)

packages/at_secondary_server/lib/src/verb/handler/abstract_update_verb_handler.dart
```diff
@@ -70,6 +75,19 @@ abstract class AbstractUpdateVerbHandler extends ChangeVerbHandler {
       'Connection with enrollment ID $enId'
       ' is not authorized to update key: $key';
 
+  /// Protected-key templates for the guard in [preProcessAndNotify], expanded
+  /// for [atSign] and lowercased, minus
+  /// [AtSecondaryConfig.encryptionPublicKeyTemplate] (delete-protected only).
+  Set<String> _updateProtectedKeys() {
```
gkc, line 81: This reads as a verb but is actually a noun - it's "keysProtectedForUpdate"

gkc (commented):
Mostly LGTM, small nit re a method name

I haven't verified that there have been no changes to flow-of-control caused by changes in if / catch / finally / etc Please can you do so with the help of your nearest friendly ai agent?

srieteja:
> Mostly LGTM, small nit re a method name
> 
> I haven't verified that there have been no changes to flow-of-control caused by changes in if / catch / finally / etc Please can you do so with the help of your nearest friendly ai agent?

Sure. But I have another refactor coming in that actually does algo negotiation. I will do so after that lands.

PR #2043, opened 2024-07-29
fix: Add "sharedKeyEnc" to the metadata to propagate from server to client on receiving end

sitaram-kalluri:
Resolves #1364

**- What I did**
- When running no-ports locally, an issue appear where the client waits for the `srvd` response and eventually times out. This occurs because the client receives the notification sent by `srvd` but fails to decrypt the message. The decryption fails because the "sharedEncryptedKey" received from the sender to the receiver is not being propagated from the receiver's server to the receiver's client via the notification.

Adding the log snippet:
```
SEVERE|2024-07-29 11:34:42.853820|NotificationServiceImpl (@alice🛠)|Failed to fetchData caused by
key not found : Exception: @alice🛠:shared_key@sitaram🛠 does not exist in keystore 

INFO|2024-07-29 11:34:44.149909| SrvdChannel |Still waiting for srvd response 
INFO|2024-07-29 11:49:03.865867| SrvdChannel |Still waiting for srvd response 
WARNING|2024-07-29 11:49:04.976872| SrvdChannel |Timed out waiting for srvd response 

Error : TimeoutException: Connection timeout to srvd @sitaram🛠 service

```
**- How I did it**
- Add "sharedEncKey" to the metadata in MonitorVerbHandler.dart

**- How to verify it**
- After the changes, tested no-ports manually and able to run successfully. Adding the log file.
[client-no-ports.log](https://github.com/user-attachments/files/16408450/client-no-ports.log)

**- Description for the changelog**
- In MonitorVerbHandler, add "sharedKeyEnc" to the metadata to propagate the sharedEncryptedKey in
  notifications from the server to the client.

There is a follow-up PR in at_client.

PR #2685, opened 2026-06-30
feat(at_server): `enroll:listns`

JeremyTubongbanua:
Related to https://github.com/atsign-foundation/at_client_sdk/issues/2008

  **- What I did**

  Implements the atServer half of the WP-SS secret-sharing substrate — the gated
  discovery verb the client's `EnrollmentDirectory` seam talks to, plus verbatim
  storage of per-enrollment key-package metadata.

  - **`enroll:listns:<namespace>`** — returns all *approved* enrollments that hold
    read-or-better access to the requested namespace, as a **flattened** list of
    `{enrollmentId, access, apkamPubKey, metadata}` (no nested `apkam[]` array).
  - **`metadata` on enrollment records** — an opaque JSON map stored verbatim from
    `EnrollParams.metadata` on `enroll:request` (the client's X-Wing key package
    rides here under `metadata['keyPackage']`; `signingAlgo` likewise). Read from
    the raw request JSON, so no at_commons grammar change is required. Surfaced in
    `enroll:fetch`, `enroll:list`, and `enroll:listns` responses.
  - **`_apsk` signing-key publication** — on new-enrollment and on `enroll:approve`,
    the enrollment's APKAM public key is published to
    `public:_apsk.<enrollmentId>.a.__e@<atSign>` so a same-atSign or peer-atSign
    verifier can always fetch it (via plookup) to verify APKAM signatures on
    `__ssenv` envelopes and advertised recipient keys. Idempotent re-publish.

  **- How I did it**

  - `enroll_verb_handler.dart` — new `listns` case (`_fetchEnrollmentsForNamespace`);
    reads `metadata`/`signingAlgo` from the raw decoded `enrollParams` JSON tail
    (the pinned at_commons `EnrollParams` model doesn't type these yet); adds
    `_publishApskSigningKey` on the new-enrollment and approve paths.
  - `enrollment_manager.dart` — `getEnrollmentsForNamespace()` with the suffix/`*`
    namespace match (both `r` and `rw` satisfy the ≥`r` bar).
  - `enroll_datastore_value.dart` (+ `.g.dart`) — `metadata` and `signingAlgo`
    fields on the record.
  - Access gate: the caller must be APKAM-authenticated with an approved
    enrollment that itself holds ≥`r` on the namespace; unauthenticated or
    under-privileged callers get `UnAuthorized`, and an empty namespace is rejected.
  - Bumps `at_secondary_server` to 3.14.0 and `at_server_spec` to 5.1.0, both on
    `at_commons ^5.12.0`.

  **- How to verify it**

  dart test packages/at_secondary_server/test/enroll_verb_test.dart
  Four new tests in the `enroll:listns discovery + _apsk + request metadata` group
  cover the flat roster shape + `apkamPubKey`/`metadata`, the ≥`r` gating (serves a
  caller with access, refuses one without), `_apsk` publication on approval, and
  `metadata`/`signingAlgo` persistence. Also verified live end-to-end: a
  virtualenv built from this branch drives an `enroll:listns` round-trip and passes
  the full at_client_sdk `at_functional_test` suite (see at_client_sdk#2037).

  `enroll:listns:<namespace>` discovery verb + opaque per-enrollment `metadata`
  for the WP-SS secret-sharing substrate.

---
**PQ plan conformance (§10(e)):** project **SS-1b** — [implementation-plan.md](https://github.com/atsign-foundation/at_client_sdk/blob/trunk/docs/projects/pq/implementation-plan.md), tracking epic atsign-foundation/at_client_sdk#1889.

### DIFF
--- a/packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
+++ b/packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
@@ -68,10 +68,12 @@ class EnrollVerbHandler extends AbstractVerbHandler {
     }
     EnrollParams? enrollVerbParams;
 
-    // Ensure that enrollParams are present for all enroll operation
-    // Exclude operation 'list' which does not have enrollParams
+    // Ensure that enrollParams are present for all enroll operation.
+    // 'list' and 'listfornamespace' carry no enrollParams JSON body.
     if (verbParams[AtConstants.enrollParams] == null) {
-      if (operation != 'list') {
+      if (operation != 'list' &&
+          operation != 'listfornamespace' &&
+          operation != 'metadata') {
         logger.severe(
             'Enroll params is empty | EnrollParams: ${verbParams[AtConstants.enrollParams]}');
         throw IllegalArgumentException('Enroll parameters not provided');
@@ -141,6 +143,23 @@ class EnrollVerbHandler extends AbstractVerbHandler {
           enrollVerbParams: enrollVerbParams,
         );
         return;
+      case 'listfornamespace':
+        response.data = await _fetchEnrollmentsForNamespace(
+          enMgr,
+          atConnection,
+          verbParams['namespace'] ?? '',
+        );
+        return;
+      case 'metadata':
+        await _storeEnrollmentMetadata(
+          enMgr,
+          atConnection,
+          verbParams['namespace'] ?? '',
+          verbParams[AtConstants.enrollParams] ?? '{}',
+          currentAtSign,
+        );
+        responseJson['status'] = 'success';
+        break;
       case 'fetch':
         response.data = await _fetchEnrollmentInfoById(
           enMgr,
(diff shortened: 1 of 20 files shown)


JeremyTubongbanua:
Need to wait for https://github.com/atsign-foundation/at_client_sdk/pull/2040

gkc:
**Aligning to the ratified WP-SS wire shape (`docs/projects/pq/`, rulings 2026-07-02)**

Thanks for getting the server side moving. The `metadata`-on-record + `getEnrollmentsForNamespace` shape is the right skeleton, but a few things need to change to match the rulings the client (atsign-foundation/at_client_sdk#2043) and at_commons (atsign-foundation/at_client_sdk#2044) now implement — the three sides need to agree exactly or the seam silently breaks:

1. **Verb name → `listns`.** The verb is `enroll:listns` (ruled 2026-07-02, superseding `listfornamespace`). Rename `case 'listfornamespace'` → `case 'listns'`, and read the namespace from `verbParams['listNamespace']` (the at_commons regex in #2044 renamed the capture group `namespace` → `listNamespace` to avoid colliding with `EnrollParams.namespaces`).
2. **Remove the `enroll:metadata` op entirely** (`case 'metadata'` + `_storeEnrollmentMetadata` + the enrollParams-guard exemption). Decision #B/#F retired it: the key package rides `EnrollParams.metadata` on **`enroll:request`**, and there is no post-enrollment metadata write. Instead, **persist `enrollParams.metadata` onto `EnrollDataStoreValue.metadata` in the `enroll:request` path** — the `metadata` field you added is correct; it just needs to be populated at request time, not via a separate verb.
3. **Add `apkamPubKey` to each response element.** Ratified flat shape is `{enrollmentId, access, apkamPubKey, metadata}` (1:1:1, no nested `apkam[]`). The client derives `KeyPackage.apkamId` from it; without it `apkamId` is null.
4. **Gate on the caller's namespace access, not just "approved."** `_fetchEnrollmentsForNamespace` currently only checks that the caller's enrollment is approved. The verb must require the caller to hold **≥`r` on the requested namespace** — mirror `_checkForNamespaceAuthorization` (suffix + `*`), and do **not** reuse `_isReadAllowed` (it is verb-type-gated and excludes `Enroll`).
5. **`signingAlgo` (DEP3).** Record `EnrollParams.signingAlgo` on the enrollment and make PKAM verification **record-authoritative** (read the stored algo, not the wire value); at_commons #2044 adds the `mldsa65` pkam literal. This can be a follow-up PR if you would rather keep this one scoped to discovery.

Refs: ratified rulings in `docs/projects/pq/decisions.md` (atsign-foundation/at_client_sdk#2042); client parser in #2043; grammar in #2044.

gkc:
Two follow-ups from today's additional rulings (docs updated on atsign-foundation/at_client_sdk#2042):

6. **Singular `metadata['keyPackage']` — no `keyPackages` map.** Under 1:1:1 an enrollment advertises exactly one key package, so it sits directly at `metadata['keyPackage']` (the payload), not a format-keyed `metadata['keyPackages']` map. Your opaque store/return is unaffected — just don't assume a `keyPackages` sub-map anywhere the server introspects `metadata`, and update the `enroll_datastore_value.dart` dartdoc that currently says "Key packages live under metadata['keyPackages'] by convention" → `metadata['keyPackage']`.

7. **Keep `_apsk` present (not just write-restricted).** The atServer should **populate** `public:_apsk.<eid>.a.__e@<atSign>` from the enrollment record's `apkamPublicKey` (on approval / first authenticated use), so a verifier can always fetch it — rather than leaving it to the client-side `ApkamSigning.publishPublicSigningKey` get-then-put (which races and can be missing). Keep the write-restriction too. Reason: advertised recipient keys (key packages, the `nskey` public half, `pqpublickey`) are now **APKAM-signed and verified against `_apsk`** the same way same-atSign and cross-atSign, so a missing `_apsk` breaks verification. See design.md §2.4 + the DEP list (#2042). e2e: an approved enrollment's `_apsk` is fetchable without a client publish, and a cross-enrollment overwrite is refused.

(Signing the advertised key package itself is **client-side** — the client wraps it in a `wrapAndSign` envelope before it rides `enroll:request`; the server stores/returns that signed envelope opaquely, so no server change for the signature itself.)

gkc:
I've implemented these review points as a PR into this branch: **#2687** (`gkc-wp-ss-server-align` → `jt-wp-ss`). It covers items 1–4, 6, 7 (verb rename to `listns`, drop the `enroll:metadata` op + persist `metadata`/`signingAlgo` at `enroll:request` time, flat `apkamPubKey` response, caller ≥r gate, singular `metadata.keyPackage`, and keeping `_apsk` present on approval), with 4 new unit tests and the full enroll suite green (86 tests). Item 5 (record-authoritative ML-DSA PKAM verify, DEP3) is left as a follow-up — `signingAlgo` is now stored on the record, the `pkam_verb_handler` wiring is the remaining piece. Merge #2687 in when it looks right.

gkc (commented):
What's your plan for publishing and releasing? at_server_spec needs a new version published; then we can do a follow on PR to bump atServer version to 3.15.0 (note for that PR, the entries in the changelog need to live under a 3.15.0 heading)

gkc:
> What's your plan for publishing and releasing? at_server_spec needs a new version published; then we can do a follow on PR to bump atServer version to 3.15.0 (note for that PR, the entries in the changelog need to live under a 3.15.0 heading) 

My usual approach is : PR with everything as it will be except for dependency overrides; merge; publish the upstream package; small follow up OR removing the dependency override; merge; canary release

gkc:
> > What's your plan for publishing and releasing? at_server_spec needs a new version published; then we can do a follow on PR to bump atServer version to 3.15.0 (note for that PR, the entries in the changelog need to live under a 3.15.0 heading)
> 
> My usual approach is : PR with everything as it will be except for dependency overrides; merge; publish the upstream package; small follow up OR removing the dependency override; merge; canary release

I misread the dependencies; looks like we'll be good to go after this PR (but a little cleanup first, I'll draft a PR)

gkc (changes requested):
@JeremyTubongbanua  All LGTM but I've made a PR onto your PR to clean up a few things before we can merge: https://github.com/atsign-foundation/at_server/pull/2696

gkc:
Per the PQ plan's §10(e) conformance rule, this PR is project **SS-1b** (see [implementation-plan.md](https://github.com/atsign-foundation/at_client_sdk/blob/trunk/docs/projects/pq/implementation-plan.md) and tracking epic atsign-foundation/at_client_sdk#1889).

gkc (approved):

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -511,6 +551,62 @@ class EnrollVerbHandler extends AbstractVerbHandler {
...
+    if (callerEnVal.approval?.state != EnrollmentStatus.approved.name) {
+      throw UnAuthorizedException('Caller enrollment is not in approved state');
+    }
+    if (_accessForNamespace(callerEnVal, namespace) == null) {
+      throw UnAuthorizedException(
+          'Caller enrollment is not authorised for namespace "$namespace"');
+    }
+
+    final members = await enMgr.getEnrollmentsForNamespace(namespace);
+    return jsonEncode(members);
```
Xlin123, line 591: can we base64 encode the return json? or is it too late to fix because of the regex?

Xlin123 (commented):
LGTM overall, worth a conversation around base64. I know in enroll we have some json, but encoding it in base64 allows for our cleaner `:` delimited wire format.

gkc (changes requested):
Needs functional tests. Will make a PR to add some

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
gkc: I think that ship has sailed; enroll is wired since a long time ago to take json on the wire. It would be nice at some point to clean everything up so there is never any json, always base64 with colons reserved for actual delimitation in command syntax but that's very much a nice-to-have for the dim and distant future imo

gkc:
> Needs functional tests. Will make a PR to add some

@JeremyTubongbanua I've made https://github.com/atsign-foundation/at_server/pull/2698 ... please review and merge if you;re comofrtable then I will re-approve this one and we can merge it to trunk

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
Xlin123: in agreement, no worries

gkc (approved):

PR #2690, opened 2026-07-04
feat(arch): cross-server to: verb (always-on inbound, outbound default off)

gkc:
## What I did

Added a cross-server `to:` verb to give us architectural flexibility for atServer endpoints — in particular, for atServers to be able to communicate with other atServers through proxies / gateways. As the first verb on a connection, `to:@x` declares the target tenant; the server replies with a `{publickey, signing_publickey}` envelope for `@x`.

Inbound and outbound are decoupled so the verb can roll out across the fleet safely:

- **Inbound** understanding of `to:@x` is **always on**. It is unauthenticated and serves only data that is already publicly readable via `lookup:`, so there is no reason to gate it — and having every atServer understand `to:` *before* any atServer emits it is what makes a staged rollout coherent.
- **Outbound** emission of `to:@target` is gated behind `AtSecondaryConfig.toVerbOutboundEnabled` (default **false** → outbound connections are byte-for-byte unchanged). When on, a peer that does not understand `to:` triggers an automatic fallback to the legacy lookup, so enabling it early is safe against any atServer in the fleet.

## How I did it

**Inbound (always on):** a local `To` verb + `ToVerbHandler` (unauth, first-verb; single-tenant, so `@x` must be this server's own atSign). `connection_util.validate()` special-cases `to:` (absent from the published `AtVerb` enum) as an unauthenticated first-verb. `InboundConnectionMetadata` gains `toAtSign`. Each value in the envelope is in the `{key, data, metaData}` shape of a `lookup:all:` response — built via the same `SecondaryUtil.prepareResponseData('all', …)` the lookup handlers use — so the caller gets the value's metadata (notably `ttr:-1`), not just the key string.

**Outbound (flag-gated, with fallback):** when `toVerbOutboundEnabled` is on, `OutboundClient.checkRemotePublicKey()` sends `to:@target` first and reads the peer's public key straight from the envelope (one fewer round trip than the separate `lookup:all:publickey@target`). If the peer rejects `to:` with an error, or **closes the connection** (atServers up to v3.0.28 close on an unknown verb rather than replying), it falls back to the legacy lookup — reconnecting first when the peer closed. `OutboundMessageListener.read()` now returns as soon as the connection is closed instead of waiting out the timeout, so a peer that closes on `to:` costs milliseconds, not seconds. The legacy lookup path is preserved verbatim as both the flag-off branch and the fallback. `AtRequestFormatter.createToRequest` added.

Outbound flag off ⇒ zero behaviour change (outbound uses the legacy `lookup:`/bare-`from:` path). Inbound is purely additive: a flag-off/legacy sender still emits the unchanged `lookup:` first verb, which this server handles exactly as before.

## How to verify it

- **Unit:** `dart test --concurrency=1 test/to_verb_test.dart test/at_request_formatter_test.dart test/secondary_config_test.dart test/outbound_to_verb_fallback_test.dart` — covers the `to:` grammar/handler, the `lookup:all:`-shaped envelope, the config default, and the outbound fallback (to: success caches with the peer's metadata incl. `ttr:-1`; error → fallback; peer-close → reconnect → fallback; null-publickey authoritative; outbound-off uses the legacy lookup only). Full package unit suite green; `dart analyze` clean.
- **Functional:** `tests/at_functional_test/test/to_verb_test.dart` asserts inbound `to:` works end to end by default — the server answers `to:@<self>` with the envelope in `lookup:all:` shape and rejects `to:` for another atSign. Full functional suite green via `runLocal.sh`.
- Once outbound is ready to roll out to the fleet, a follow-up PR will enable `toVerbOutboundEnabled` and cover it in e2e / functional tests.

### DIFF
--- a/packages/at_secondary_server/config/config.yaml
+++ b/packages/at_secondary_server/config/config.yaml
@@ -35,6 +35,14 @@ security:
   # atServer
   clientCertificateRequired: true
 
+# Cross-server 'to:' verb: the first verb on an outbound peer connection
+# declares the target tenant, giving architectural flexibility for atServer
+# endpoints — e.g. atServers reached through proxies / gateways.
+# Default false. Turn on (env crossServerToVerbEnabled=true) only once every
+# atServer in the fleet understands 'to:' — a staged rollout.
+protocol:
+  crossServerToVerbEnabled: false
+
 # The atProtocol storage configurations
 hive:
   # The storage path for secondary storage.
--- a/packages/at_secondary_server/config/config.yaml
+++ b/packages/at_secondary_server/config/config.yaml
@@ -38,10 +38,16 @@ security:
 # Cross-server 'to:' verb: the first verb on an outbound peer connection
 # declares the target tenant, giving architectural flexibility for atServer
 # endpoints — e.g. atServers reached through proxies / gateways.
-# Default false. Turn on (env crossServerToVerbEnabled=true) only once every
-# atServer in the fleet understands 'to:' — a staged rollout.
+# Understanding inbound 'to:@x' is always on (unauthenticated, serves only
+# data already publicly readable via lookup). Emitting 'to:@target' on
+# outbound peer connections is gated here:
+# - toVerbOutboundEnabled (default false): emit 'to:@target' as the first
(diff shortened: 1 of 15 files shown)


cconstab (approved):
LGTM

PR #2692, opened 2026-07-05
feat(persistence): SQLite persistence backend + Hive↔SQLite migration + dual-mode validation

### LINKED ISSUE #2697: feat: sqlite for atServer persistence

We continue to carry significant costs - memory, code complexity and perf - from our dependence on Hive.

Following the successful [refactoring](https://github.com/atsign-foundation/at_client_sdk/issues/2045) of the persistence API, let's now land the multiple performance benefits we can get from implementing sqlite as a persistence layer.

Why not a "proper" dbms? Too much of a break from right now. The sqlite approach allows us to use the existing atServer fleet's infrastructure architecture unchanged, as well as giving us the option to adopt in AtClient SDKs

### Additional context

_No response_

gkc:
Fixes #2697 

**- What I did**

Add a SQLite persistence backend for the atServer alongside Hive, behind the existing factory/bundle/config abstraction, plus Hive↔SQLite migration and a dual-mode validation gate.

**- How I did it**

- New SQLite backend (`lib/sqlite.dart`): keystore, commit log (dense counters-based ids), notifications, access log — with full-fidelity snapshots, path queries, and true transactions. One `atsign.db` per atSign.
- Verbatim-restore primitives + `PersistenceMigrator` + `PersistenceSnapshot` comparator; config-driven backend selection with an abort-on-failure startup migrate-verify-flip.
- Dual-write mode (Hive primary, every write mirrored into SQLite) + a DB-set compare CLI, gated by a conversion-integrity round-trip and a new `functional_dual_verify` CI workflow.

Per-commit detail is in the commit messages.

**- How to verify it**

- `dart test --concurrency=1` in `at_persistence_secondary_server` (includes the 6-DB conversion-integrity gate).
- Functional pack in dual mode → every atSign's Hive and SQLite DB sets are identical (`functional_dual_verify.yaml`; 40/40 locally and in CI).

Add a SQLite persistence backend for the atServer, with Hive↔SQLite migration and dual-mode DB-set validation.

---

### This is a capability release — nothing flips by default

The default persistence backend stays **Hive**. `config.yaml` ships `persistence.backend: 'hive'` and `AtSecondaryConfig.persistenceBackend` defaults to `hive`, so merging and deploying this changes no server's behaviour or on-disk storage. It only *adds* the ability to select SQLite (or the transient `dual` validation harness) per server, via the `persistenceBackend` env var or the `persistence.backend` yaml key. No atServer runs on SQLite until it is explicitly, individually flipped.

### Client-side (at_client) migration is a separate, later effort

This PR is **server-side only** — it changes how the atServer persists data. The `at_client` SDK keeps its own on-device local storage (Hive today), and migrating that to SQLite is a distinct piece of work. The `at_client` migration code will live in a **later PR in the `at_client` repo**, if and when we decide to do that migration; it is deliberately out of scope here.

### Release sequencing (this PR carries a temporary bridge)

`at_secondary_server` consumes new `at_persistence_secondary_server` **5.2.0** API that isn't on pub.dev yet, and CI resolves sibling packages from pub.dev. So this PR includes a committed `dependency_overrides` path, isolated in its own commit so it reverts cleanly. Landing sequence:

1. **Merge** this PR.
2. **Publish** `at_persistence_secondary_server` 5.2.0 to pub.dev.
3. Open a small **follow-up PR** that removes the `dependency_overrides` block and depends on the hosted `^5.2.0`.
4. **Merge** the follow-up.
5. **Release** `at_secondary_server`.

### What happens when a container is restarted with the backend flipped to SQLite

Setting `persistenceBackend=sqlite` (env) or `persistence.backend: 'sqlite'` (yaml) and restarting triggers a one-time, in-process migration during startup — **before the server opens any store or accepts a connection**:

1. Startup reads a single marker file, `<storageRoot>/.persistence_backend`. On a Hive-only server it is absent, so the current backend is taken as `hive`; the configured target is `sqlite`; they differ, so a migration runs.
2. Both bundles are opened — Hive (source) and SQLite (target, under `<storageRoot>/sqlite`). The SQLite target is `clear()`ed first, so a retry after an earlier failed attempt starts from a clean tree rather than stacking onto partial data.
3. `PersistenceMigrator` copies all four stores **verbatim** — keystore, commit log (preserving commit ids), access log (preserving timestamps), and notifications — reading the source and never mutating it.
4. It then captures a backend-neutral snapshot of both stores and **verifies they are byte-identical**, while both are still open and *before* anything is committed.
5. **Only on a clean verify** does it write the marker (`activeBackend: sqlite`, `previousBackend: hive`, the retained Hive paths, and `migratedAt`). Bootstrap then opens SQLite, and the server comes up on SQLite.

The old Hive data is **retained on disk** — never deleted at startup. It is reclaimed only out-of-band by `bin/cleanup_stale_persistence.dart`, and only after a configurable retention window.

**Abort-on-failure.** If the migration or the verify fails, the exception propagates out of `start()` and the process exits without serving a single request. Because the marker is written only *after* a clean migration, and the migration only *reads* the Hive source (the `clear()` touches only the SQLite target), nothing is committed: a restart reverted to `hive` comes straight back up on the original, untouched data, and a restart still pointed at `sqlite` retries the whole migration from a clean target.

**Steady state & rollback.** Once a server is on SQLite, every subsequent restart sees the marker and config agree and skips migration entirely — it just opens SQLite. Flipping the config back to `hive` and restarting runs the reverse migration (SQLite→Hive) with the same clear → migrate → verify → flip, rebuilding Hive from the *current* SQLite so any writes made while on SQLite are preserved.

### Rollout: atSign by atSign on canaries — never a fleet flip

Each secondary hosts a **single atSign** with its own storage root and its own marker, so the backend is chosen per-server / per-atSign. The actual move to SQLite will be rolled out **one atSign at a time on canary servers** — flip, migrate, verify, observe — before the fleet is even considered. There is no fleet-wide switch; the default remains Hive, and each canary atSign is individually validated on SQLite before any broader rollout is contemplated.

### DIFF
--- a/tools/build_secondary/Dockerfile
+++ b/tools/build_secondary/Dockerfile
@@ -34,16 +34,33 @@ RUN \
   cp config/config.yaml $HOMEDIR/config/ ; \
   cp cacert/cacert.pem /etc/cacert ; \
   cp pubspec.yaml $HOMEDIR/
+
+# Stage the architecture-appropriate libsqlite3 (+ its libz dependency) into
+# /runtime, alongside the Dart AOT runtime, so the scratch stage's static
+# `COPY --from=buildimage /runtime/ /` picks up the correct library for the
+# build architecture (a scratch-stage COPY cannot be arch-conditional).
+# `--dereference` resolves the SONAME symlink so a real file lands in the
+# scratch image (a copied symlink would dangle). Only dlopen'd at runtime when
+# persistence.backend=sqlite (the default hive image never loads it). NOTE:
+# confirm the full transitive .so set with a Docker build before the first
+# SQLite production deploy (this scratch image has no package manager).
+RUN \
+  case "$(dpkg --print-architecture)" in \
+    amd64) TRIPLET="x86_64-linux-gnu" ;; \
+    armhf) TRIPLET="arm-linux-gnueabihf" ;; \
+    arm64) TRIPLET="aarch64-linux-gnu" ;; \
+    riscv64) TRIPLET="riscv64-linux-gnu" ;; \
+    *) echo "Unsupported architecture: $(dpkg --print-architecture)" ; exit 5 ;; \
+  esac ; \
+  FILES="/usr/lib/$TRIPLET/libsqlite3.so.0 /usr/lib/$TRIPLET/libz.so.1" ; \
+  for f in $FILES ; do \
+    mkdir -p "/runtime$(dirname "$f")" ; \
+    cp --archive --link --dereference --no-target-directory "$f" "/runtime$f" ; \
+  done
+
 # Second stage of build FROM scratch
 FROM scratch
 COPY --from=buildimage /runtime/ /
-# libsqlite3 (+ libz) for the SQLite persistence backend's FFI. Only
-# dlopen'd when persistence.backend=sqlite; the default (hive) image never
-# loads it. NOTE: the exact transitive .so set must be confirmed with a
-# Docker build before the first SQLite production deploy (this scratch
-# image has no package manager to pull deps at runtime).
-COPY --from=buildimage /usr/lib/x86_64-linux-gnu/libsqlite3.so.0 /usr/lib/x86_64-linux-gnu/
-COPY --from=buildimage /usr/lib/x86_64-linux-gnu/libz.so.1 /usr/lib/x86_64-linux-gnu/
 COPY --from=buildimage /etc/passwd /etc/passwd
 COPY --from=buildimage /etc/group /etc/group
 COPY --from=buildimage /etc/cacert /etc/cacert
--- a/tools/build_secondary/Dockerfile
+++ b/tools/build_secondary/Dockerfile
@@ -12,6 +12,9 @@ COPY ./packages/at_persistence_spec/ ./at_persistence_spec
 COPY ./packages/at_persistence_secondary_server/ ./at_persistence_secondary_server
 COPY ./packages/at_secondary_server/ ./at_secondary_server
 RUN \
+  apt-get update ; \
+  apt-get install -y --no-install-recommends libsqlite3-0 ; \
+  rm -rf /var/lib/apt/lists/* ; \
   mkdir -p $HOMEDIR/storage ; \
   mkdir -p $HOMEDIR/config ; \
   mkdir -p /etc/cacert ; \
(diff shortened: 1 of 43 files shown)


tools/build_secondary/Dockerfile
```diff
@@ -34,6 +37,13 @@ RUN \
 # Second stage of build FROM scratch
 FROM scratch
 COPY --from=buildimage /runtime/ /
+# libsqlite3 (+ libz) for the SQLite persistence backend's FFI. Only
+# dlopen'd when persistence.backend=sqlite; the default (hive) image never
+# loads it. NOTE: the exact transitive .so set must be confirmed with a
+# Docker build before the first SQLite production deploy (this scratch
+# image has no package manager to pull deps at runtime).
+COPY --from=buildimage /usr/lib/x86_64-linux-gnu/libsqlite3.so.0 /usr/lib/x86_64-linux-gnu/
```
cpswan: This is going to fail for all the other architectures, so we need something like the case statement in the [Dart dockerfile](https://github.com/dart-lang/dart-docker/blob/main/Dockerfile-debian.template) to select the right library for the architecture it's running on.

```sh
    case "$(dpkg --print-architecture)" in \
        amd64) \
            TRIPLET="x86_64-linux-gnu" ; \
            FILES="/lib64/ld-linux-x86-64.so.2" ;; \
        armhf) \
            TRIPLET="arm-linux-gnueabihf" ; \
            FILES="/lib/ld-linux-armhf.so.3 \
                /lib/arm-linux-gnueabihf/ld-linux-armhf.so.3";; \
        arm64) \
            TRIPLET="aarch64-linux-gnu" ; \
            FILES="/lib/ld-linux-aarch64.so.1 \
                /lib/aarch64-linux-gnu/ld-linux-aarch64.so.1" ;; \
        riscv64) \
            TRIPLET="riscv64-linux-gnu" ; \
            FILES="/lib/ld-linux-riscv64-lp64d.so.1 \
                /lib/riscv64-linux-gnu/ld-linux-riscv64-lp64d.so.1" ;; \
        *) \
            echo "Unsupported architecture" ; \
            exit 5;; \
```

I'd suggest that we do that in the build stage, and copy the .so file into /runtime, so it then gets picked up by `COPY --from=buildimage /runtime/ /` above.

cpswan (changes requested):

tools/build_secondary/Dockerfile
gkc: thanks Chris will do

cpswan: Looking at this a little more... I think we just need to define the triplet, and then copy over the two libs:

```sh
    case "$(dpkg --print-architecture)" in \
        amd64) \
            TRIPLET="x86_64-linux-gnu" ;; \
        armhf) \
            TRIPLET="arm-linux-gnueabihf" ;; \
        arm64) \
            TRIPLET="aarch64-linux-gnu" ;; \
        riscv64) \
            TRIPLET="riscv64-linux-gnu" ;; \
        *) \
            echo "Unsupported architecture" ; \
            exit 5;; \
    esac; \
    FILES=/usr/lib/$TRIPLET/libsqlite3.so.0 \
          /usr/lib/$TRIPLET/libz.so.1 ; \
    for f in $FILES; do \
        dir=$(dirname "$f"); \
        mkdir -p "/runtime$dir"; \
        cp --archive --link --dereference --no-target-directory "$f" "/runtime$f"; \
    done
```

gkc: Thanks @cpswan - done. See latest commit.

cpswan (approved):

PR #2704, opened 2026-07-09
fix(secondary): enforce authorization in keys verb + harden CRAM secret handling

cconstab:
## Summary

Fixes a **authorization gap in the `keys` verb** plus the CRAM-secret handling . Found during a security review of the secondary server.

## The problem

The `keys:get` / `keys:delete` **by-name** branches performed **no authorization**. `hasManageAccess` was computed but only consulted in the list branch; the by-name branches called `keyStore.get` / `keyStore.remove` on any exact key name supplied (the `keyName` param is `\S+`, so it can contain `:`).

Impact — any **approved enrollment**, even one scoped to a single namespace with no `__manage`, could:
- Read or delete **any** key on the secondary, including other enrollments' keys and arbitrary namespace data outside its grant.
- Read the raw CRAM secret (`privatekey:at_secret`) → CRAM-authenticate → a CRAM-authed `enroll:request` is auto-approved with `__manage:rw` + `*:rw` = full account takeover.

CRAM resurrection variant: even where onboarding had deleted the CRAM secret, the same bug could delete the `privatekey:at_secret_deleted` marker; on the next restart bootstrap re-plants the secret from the `-s` arg — and since long-running servers restart without `-s`, it re-planted an **empty** secret. The CRAM handler only checked the record existed, not that it had a value, so the auth digest was computed over the literal string `null`, which any caller can reproduce from the public session id.

## The fix

- **`keys` verb** (`keys_verb_handler.dart`): by-name `get`/`delete` now enforce the same rule the list branch already applied — a caller may only touch keys tagged with its **own `enrollmentId`** (plus its own PEK/SEK). Everything else, including raw non-enrollment keys, is refused with `UnAuthorizedException`.
- **Bootstrap** (`at_secondary_impl.dart`): only plants the CRAM secret when it is not already present **and** a non-empty `shared_secret` was supplied — so a restart can neither clobber nor resurrect it.
- **CRAM handler** (`cram_verb_handler.dart`): rejects a null/empty secret, closing the empty-secret digest path.

No behaviour change for legitimate flows: keys created via `keys:put` always carry the caller's `enrollmentId`, so they remain accessible.

## Tests

- `keys_verb_authorization_test.dart` (new): a wavi-scoped enrollment is denied reading the CRAM secret, reading a foreign-namespace key, and deleting an arbitrary key — and is still allowed to read its own enrollment-tagged key.
- `cram_verb_test.dart`: new case proving an empty CRAM secret is rejected (the null-replant backdoor).
- `delete_verb_test.dart`: confirms the `privatekey:at_secret_deleted` marker cannot be deleted via the delete verb.

All `at_secondary_server` tests pass (822), `dart format` clean (80-col), `dart analyze` clean.

## Scope

Changes are confined to `packages/at_secondary_server`. No keyless RCE or atClient impersonation was found — core auth (PKAM/CRAM/pol) is sound; this closes a privilege-escalation path available to an already-enrolled app.

### DIFF
--- a/packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
+++ b/packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
@@ -839,6 +839,33 @@ class AtSecondaryServerImpl implements AtSecondaryServer {
     throw Exception("AtSecondaryServer.getMetrics() is obsolete");
   }
 
+  /// Plants the CRAM (registrar activation) secret into [keyValueStore] for
+  /// first-time activation. Deliberately conservative: the secret is planted
+  /// only when it has NOT been explicitly deleted (marker absent), is NOT
+  /// already present, and a non-empty [sharedSecret] was supplied. This
+  /// prevents (a) clobbering a real secret with a null on a restart started
+  /// without `-s`, and (b) resurrecting a deleted secret. A null/empty secret
+  /// must never be stored — it would be CRAM-authenticatable (see
+  /// [CramVerbHandler]).
+  @visibleForTesting
+  Future<void> plantCramSecretIfRequired(
+      AtKeyValueStore<String, AtData, AtMetaData?> keyValueStore,
+      String? sharedSecret) async {
+    final cramSecretDeleted =
+        await keyValueStore.exists(AtConstants.atCramSecretDeleted);
+    final cramSecretExists =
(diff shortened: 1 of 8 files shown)


gkc (commented):
LGTM but I will provide a couple of additional tests in a PR against this PR, will ping you when ready

gkc (approved):

PR #2705, opened 2026-07-09
test(secondary): cross-enrollment isolation + CRAM bootstrap coverage for #2704

gkc:
**- What I did**

Two additional tests for #2704 (LGTM from me — this is follow-on coverage, targeted at your branch to merge in). They close the two coverage gaps a review of the PR surfaced; the shipped code is correct, these just guard it against future regressions.

- **Cross-enrollment isolation** (`keys_verb_authorization_test.dart`, +2 tests) — a well-formed keys-verb JSON value tagged with **another** enrollment's `enrollmentId` must be refused for both `get` and `delete`.
- **CRAM bootstrap planting guard** (`cram_secret_bootstrap_test.dart`, new) — the startup guard plants on first activation only, and never resurrects/clobbers/empties the secret.

**- How I did it**

- The isolation tests hit the *discriminating* branch of `_isAuthorizedForKey` — `decoded is Map && decoded[enrollmentId] == enId`. The existing refuse tests use non-JSON values, so they only reach the `jsonDecode`-throws path; a gate loosened to `return decoded is Map;` (i.e. dropping the enId equality — a real auth hole) passes the whole current suite. These catch that.
- The bootstrap guard was inline in `_initializePersistentInstances`, reachable only via a full `start()`, so it had no direct test — a revert to the old unconditional planting passes the suite. I extracted the block **verbatim** into `plantCramSecretIfRequired(...)` (`@visibleForTesting`, behaviour-identical) so it's unit-testable, and covered: plant-on-first-activation, don't-plant-empty, don't-plant-absent, don't-clobber-on-restart-without-`-s`, don't-resurrect-when-marker-present.

**Heads-up:** the second commit touches your production code (the extraction). It's a pure, behaviour-neutral move — happy to reshape or drop it if you'd rather structure the seam differently; the first commit is test-only and stands alone.

**- How to verify it**

Please review commit-by-commit — commit 1 is test-only; commit 2 is the extraction + its test. Locally: `dart analyze` clean, and `dart test test/keys_verb_authorization_test.dart test/cram_secret_bootstrap_test.dart test/cram_verb_test.dart test/keys_verb_test.dart --concurrency=1` all green.

n/a — test-only (plus a behaviour-neutral `@visibleForTesting` extraction).

### DIFF
--- a/packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
+++ b/packages/at_secondary_server/lib/src/server/at_secondary_impl.dart
@@ -839,6 +839,33 @@ class AtSecondaryServerImpl implements AtSecondaryServer {
     throw Exception("AtSecondaryServer.getMetrics() is obsolete");
   }
 
+  /// Plants the CRAM (registrar activation) secret into [keyValueStore] for
+  /// first-time activation. Deliberately conservative: the secret is planted
+  /// only when it has NOT been explicitly deleted (marker absent), is NOT
+  /// already present, and a non-empty [sharedSecret] was supplied. This
+  /// prevents (a) clobbering a real secret with a null on a restart started
+  /// without `-s`, and (b) resurrecting a deleted secret. A null/empty secret
+  /// must never be stored — it would be CRAM-authenticatable (see
+  /// [CramVerbHandler]).
+  @visibleForTesting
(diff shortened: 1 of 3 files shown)


cconstab (approved):
LGTM

PR #2713, opened 2026-07-16
chore: Add more enrollment-related functional tests, clean up inconsistencies

gkc:
**- What I did**
- Enrollment authorization hardening so a broad `*` ("all namespaces") grant no longer reaches management-only surfaces unless it also has `__manage` namespace access
  - the __manage namespace (enrollment records etc) is reachable only via enrollment/otp/monitor by an explicit __manage holder
  - each enrollment's per-enrollment reserved namespace (<id>.a|r|d.__e) stays private to it (public keys excepted);
- otp:get requires __manage; enroll:fetch is scoped to the caller's own enrollment, or a __manage holder with access to all of the target's namespaces.
- added more tests to the functional pack to ensure all of the enrollment authorization gates are covered (allow and deny)
- added an APKAM enrollment overview doc (Note: This is a temporary location, I'll move it to the at_protocol repo in an upcoming PR)
- added a force-pull of the VE base image in the functional `runLocal.sh` (so a stale cached copy can't be used)

**- How I did it**
See commit & diffs

**- How to verify it**
Tests pass


gkc:
@cconstab please review

cconstab (approved):

PR #2716, opened 2026-07-18
fix(migration): drop orphaned commit entries instead of migrating them

gkc:
**- What I did**

- `PersistenceMigrator` no longer copies ORPHANED commit entries into the target — a non-delete commit entry whose key is absent from the keystore.
  - Hive accumulates these (commit-log box and in-memory cache can disagree; the TTL-expiry purge is cache-driven so it can miss the box row). SQLite purges by atKey directly against storage and cannot produce one.
  - Migrating them forward would make them permanent residents of a backend that can never generate one. `PersistenceSnapshot` filters them from both sides, so the post-migration verify is structurally blind to them.
  - Migration already walks both stores, so the live-key set is free.
- Deliberately NOT treated as orphans: DELETE tombstones (they legitimately have no keystore row — dropping them would break delete propagation), keys that are expired but still present, and entries with a null atKey.
- `MigrationReport` gains a required `orphanedCommitEntries` field; drops are logged per entry plus a total.
- Corrected the 5.2.0 CHANGELOG entry that called such an orphan "sync-benign" — it wasn't, it failed the whole sync request with `AT0015` until #2715.
- Dropped the stray 5.2.1 CHANGELOG line for the `sweepStaleSource` fix; that code lives in `at_secondary_server` and is already recorded under its 3.15.0 heading.

Related: #2715 is the consumer-side guard in the sync verb handler. This PR removes the orphans themselves at migration time.

Note for review: `PersistenceSnapshot`'s public fields were already replaced in 5.2.1 (a breaking change to an exported class under a patch bump), and the new required `MigrationReport` field compounds that. The version floor may want revisiting before this ships — not changed here.

**- How I did it**

See commit & diffs

**- How to verify it**

Tests pass — new `persistence_migrator_orphan_test.dart` (4 cases, including the load-bearing "DELETE tombstone survives" negative test); full package suite 292/292 green including the `conversion_integrity` round-trip gate and `dual_write`.

### DIFF
--- a/packages/at_persistence_secondary_server/CHANGELOG.md
+++ b/packages/at_persistence_secondary_server/CHANGELOG.md
@@ -1,7 +1,23 @@
 ## 5.2.1
 
+- fix: `PersistenceMigrator` now drops ORPHANED commit entries rather than
+  copying them into the target — a non-delete commit entry whose key is absent
+  from the keystore. Hive accumulates these (its commit log's box and in-memory
+  cache can disagree, and the TTL-expiry purge is cache-driven, so it can miss
+  the box row); SQLite purges by atKey directly against storage and cannot
+  produce one. Migrating them forward would make them permanent residents of a
+  backend that can never generate one, and `PersistenceSnapshot` filters them
+  from both sides so the post-migration verify cannot see them. Migration is
+  where both stores are already walked, so reconciling costs nothing.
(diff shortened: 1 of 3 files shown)


gkc:
@cconstab This one is not urgent, just a follow-up for before we start migrations

cconstab (approved):

PR #2735, opened 2026-08-03
perf(at_persistence_secondary_server): seek skip-deletes via a partial index

gkc:
**- What I did**

Follow-up to the query-level skip-deletes shipped in 5.2.1 (#18f83203). That change stopped below-watermark DELETE rows being materialised into Dart, but its `NOT (operation = '-' ...)` predicate still makes SQLite walk the whole commit_id index in its C layer, so a client near the head of a delete-dominated log pays a full-log scan.

- Add a partial index `commit_log_live` over non-DELETE rows and express the SQLite skip-deletes traversal as index-driven range scans over it (live segment + kept-latest point lookup + above-watermark segment), instead of a filter over the full index.
  - **~253ms → ~0.2ms** on a 1M-entry log holding one live entry (measured; trunk's strategy vs this one, identical data — see the bench's `sqlite:trunk` vs `sqlite:partial-idx` rows).
- `SqliteAtCommitLog.iterate` streams via a prepared cursor instead of an eager `select()` that materialised the full result set before the first row — ordinary `sync` returning 25 entries on a 1M log **~500ms → sub-ms**.
- fix: `getSize()` on the SQLite commit-log/access-log/notification stores returned bytes; the spec and Hive stores return KB. All three now return KB (had over-reported 1024x).
- No schema `user_version` bump: the index is `CREATE INDEX IF NOT EXISTS` on open, so existing DBs acquire it on next start and rollback is unaffected.

**- How I did it**

See commit & diffs. Handler and Hive path unchanged — this is entirely inside the SQLite backend plus the shared schema.

**- How to verify it**

Tests pass (`dart test --concurrency=1`): 318 in at_persistence_secondary_server incl. 18 new equivalence/plan tests, 55 sync tests in at_secondary_server. The new `skip_deletes_pushdown_test.dart` pins pushed-vs-filtered equivalence on both backends and asserts the plan uses `commit_log_live`.

perf: SQLite sync skip-deletes now seeks a partial index instead of scanning the full commit log; iterate streams via a cursor rather than materialising all rows; getSize returns KB not bytes.

### DIFF
--- a/packages/at_persistence_secondary_server/bench/run_sync_bench.sh
+++ b/packages/at_persistence_secondary_server/bench/run_sync_bench.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+# Sweeps the sync commit-log scan bench across log sizes and writes the full
+# output to a file. See bench/sync_commit_log_bench.dart for what is measured.
+#
+#   ./bench/run_sync_bench.sh                 # 100k, 500k, 1M
+#   SIZES="100000" ./bench/run_sync_bench.sh  # override the sweep
+#
+# Output: bench/results/sync_bench_<host>.txt (full log, never piped through
+# tail — a truncated failure means re-running a multi-minute seed).
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+SIZES=${SIZES:-"100000 500000 1000000"}
+REPEAT=${REPEAT:-3}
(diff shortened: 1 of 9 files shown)


packages/at_persistence_secondary_server/bench/run_sync_bench.sh
gkc, line 1: including the benchmark script and code as it will no doubt come in handy in future

cpswan (approved):

PR #2736, opened 2026-08-03
superseded by #2739

gkc:
**- What I did**

- Added the `mldsa65` branch to PKAM signature verification. Without it a post-quantum APKAM keypair could never authenticate: an unrecognised `signingAlgo` fell through to the RSA default, so a well-formed signature over a correct public key was simply interpreted under the wrong algorithm.
- Made an APKAM-authenticated connection take its signing algorithm from the **enrollment record** rather than from whatever the client restates on each connect.
  - Hardening, not a fix — the signature is checked against the stored public key either way, so a client that misstates the algorithm only fails its own verification and cannot downgrade anything. What it closes off is cross-algorithm confusion, where one key blob parses under more than one algorithm.
- Left legacy PKAM reading the value from the wire. It has no enrollment record to be authoritative about, and the functional suite authenticates over that path with an `ecc_secp256r1` key — so pinning it to `rsa2048` would have broken working behaviour rather than tightened anything. A legacy enrollment predating the `signingAlgo` field records none and keeps the existing default.

**- How I did it**

See commit & diffs.

**- How to verify it**

Tests pass. `enroll_verb_test.dart` is green (87), and the `at_client_sdk` functional pack runs green (104) against an atServer built from this branch — every enrollment test there authenticates over the APKAM path this changes.

PKAM verification accepts `signingAlgo:mldsa65`, and an APKAM connection's signing algorithm is read from its enrollment record rather than from the wire. Legacy PKAM is unchanged.

---

Part of the PQ programme's **SS-3**; the rulings behind it are recorded in `at_client_sdk` `docs/projects/pq/decisions.md` section 21.

### DIFF
--- a/packages/at_secondary_server/CHANGELOG.md
+++ b/packages/at_secondary_server/CHANGELOG.md
@@ -1,4 +1,17 @@
 # 3.15.1
+- feat: PKAM verification accepts `signingAlgo:mldsa65`. Without the branch a
+  post-quantum APKAM keypair fell through to the RSA default and could never
+  authenticate at all — the signature was well formed, just interpreted under
+  the wrong algorithm.
+- feat: for an APKAM-authenticated connection, the signing algorithm is now
+  read from the **enrollment record** rather than restated by the client on
+  each connect. It is hardening rather than a fix: the signature is checked
+  against the stored public key either way, so a client that misstates the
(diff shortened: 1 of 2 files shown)


gkc:
Closing as Superseded

PR #2739, opened 2026-08-08
feat(pq): post-quantum enrollment — self-enrollment, ML-DSA PKAM, and expiry bounds

gkc:
**- What I did**

Post-quantum enrollment support in the atServer, and the test coverage for it.

- **PKAM can verify post-quantum signatures.** An enrollment using an ML-DSA key could never authenticate: the server did not recognise `signingAlgo:mldsa65`, fell back to the RSA default, and so checked a perfectly good signature with the wrong algorithm.
- **PKAM takes the signing algorithm from the enrollment record rather than from the client.** An APKAM connection used to restate its algorithm on every connect. The stored record is now authoritative, which rules out one key being interpreted under two different algorithms. Legacy PKAM is unchanged — it has no enrollment record to consult, and may legitimately present an ECC key.
- **An enrollment can now enrol itself a replacement.** A client already authenticated with an approved enrollment can send `enroll:request` and get a second enrollment back, auto-approved, with no OTP and nobody approving it by hand. This is how a device upgrades its own credential to post-quantum keys without going through onboarding again. The new enrollment can hold at most what the existing one holds, and the old one is kept alive for a grace period so that other devices sharing the same keyfile can still upgrade.
- **A self-enrolled enrollment expires, and cannot outlive the one that created it.** It was being written with no expiry at all, and the requesting client could name its own — including "never expires" — so a stolen keyfile that was deliberately limited to an hour could have minted itself a permanent credential. It now inherits the original's expiry and may only shorten it.
- **An enrollment request may leave out the RSA-wrapped symmetric key when it offers a key package instead.** That field is the one part of enrollment a quantum computer would break, because the client wraps a secret to the atSign's long-lived RSA key. If the approver mints that secret instead and encrypts it to the client's post-quantum key package, the client has nothing to put in the field. It stays mandatory for every other request.
- **The published signing key now says which algorithm it is.** An enrollment's signing key is published at `_apsk` as a bare RSA key. A raw ML-DSA key published the same way would be read as RSA and fail every signature, so non-RSA enrollments now publish a small tagged JSON object naming the algorithm. RSA enrollments are untouched, so existing apps keep parsing what they always did.
- **A notification the server refuses to deliver is now logged.** The authorization check dropped it with no log at any level, so a refusal on the receiving side was indistinguishable from the sender never having sent it — and got blamed on the wrong side.
- **Tests.** Functional coverage over a real connection for all of the above; a revoked enrollment is now proven unable to re-authenticate on the ordinary revoke path, not just one variant; and one existing `enroll:listns` test is fixed, because its fixture relied on an enrollment that the record-authoritative change above correctly stops from authenticating.
- **Unrelated test, carried here for convenience.** An end-to-end test that `appMetadata` survives a cross-atSign `lookup:all`. The existing `appMetadata` tests only cover same-server paths, so nothing checked that it survives the hop where one atSign's server relays another's response.
- **Build.** `at_chops` is resolved from a tag on `at_client_sdk` until the version carrying ML-DSA verification is published — by the atServer, and now by the functional test pack too, which needs to generate an ML-DSA keypair and sign with it.
- **Version.** The in-progress version becomes a minor rather than a patch, since this carries features and not only fixes.

**- How I did it**

Review the commits one by one.

**- How to verify it**

Tests pass — `at_secondary_server` unit suite and the functional suite.

Unit (`packages/at_secondary_server`):

- Self-enrollment: a subset child is auto-approved; a wildcard parent covers namespaces it never named; a retrofit may keep its own `(appName, deviceName)`; the child expires per the posture it inherited; a stated expiry wins over the inherited one.
- Refusals: a namespace the parent does not hold; broader access letters on a held namespace; `__manage` via a wildcard parent; empty namespaces; an unapproved parent.
- Expiry cap: re-arms on each retrofit; never extends past the parent's own posture; a child may not outlive its parent, may not claim "never expires" against a bounded parent, and a negative value is not honoured.
- ML-DSA: an `mldsa65` enrollment publishes the tagged `_apsk` while an RSA one stays bare; it then authenticates with a genuine ML-DSA signature through the production verify path, with a tampered signature as the control; an enrollment with no recorded algorithm keeps the RSA default even when the wire claims `mldsa65`.
- Key package: a request offering one may omit the encrypted symmetric key.
- Notification gate: a self-notification in the enrollment's own namespace is delivered; one outside it is not.

Functional (`tests/at_functional_test`), driving the real wire:

- `apkam_self_enrollment_test.dart` — the self-enrollment rules above, end to end: the child authenticates; it records its parent; it needs no encrypted symmetric key; escalation is refused on every axis, with a positive control so the refusals are not passing vacuously; every expiry clamp is checked against the stored record; the parent survives and its cap re-arms; `_apsk` is bare for RSA and tagged otherwise; and PKAM keeps the record's RSA default when the wire claims `mldsa65`.
- `mldsa_pkam_test.dart` — a real ML-DSA keypair is generated in the test, enrolled over the wire, and used to answer a real `from:` challenge, with a one-byte-corrupted signature as the control. This is the whole path rather than the in-process one; a resolved at_chops with no ML-DSA branch once sat green against tests that asserted record storage only, while the live wire died parsing a raw ML-DSA key as RSA.
- `enrollment_notification_delivery_test.dart` — a scoped enrollment's monitor receives its own namespace and not another, read over a raw socket.
- `enroll_verb_test.dart` — a revoked enrollment cannot re-authenticate on subsequent connections, asserting success *before* the revoke so the refusal after it is attributable.

End-to-end (`tests/at_end2end_test`):

- `app_metadata_test.dart` — `appMetadata` survives a cross-atSign `lookup:all`, guarded on both servers being 3.14.0 or newer since this pack runs against long-lived atSigns.

Each new test was checked against the code before the change and fails there, so it detects the behaviour rather than merely passing alongside it.

### DIFF
--- a/packages/at_secondary_server/lib/src/server/persistence_backend.dart
+++ b/packages/at_secondary_server/lib/src/server/persistence_backend.dart
@@ -59,7 +59,8 @@ class PersistenceBackendMarker {
   static PersistenceBackendMarker? read(String path) {
     final file = File(path);
     if (!file.existsSync()) return null;
-    return fromJson(jsonDecode(file.readAsStringSync()) as Map<String, dynamic>);
+    return fromJson(
+        jsonDecode(file.readAsStringSync()) as Map<String, dynamic>);
   }
 
   /// Writes the marker to [path] (creating parent dirs).
@@ -161,9 +162,9 @@ class PersistenceBackendManager {
     final currentBackend = marker?.activeBackend ?? hive;
     if (currentBackend == targetBackend) return;
 
-    _logger.warning(
-        'Persistence backend change for $atSign: $currentBackend -> '
-        '$targetBackend. Migrating (abort-on-failure)...');
+    _logger
+        .warning('Persistence backend change for $atSign: $currentBackend -> '
+            '$targetBackend. Migrating (abort-on-failure)...');
 
     final sourceFactory = factoryFor(currentBackend);
     final targetFactory = factoryFor(targetBackend);
@@ -203,8 +204,7 @@ class PersistenceBackendManager {
         migratedAt: DateTime.now(),
       ),
     );
-    _logger.warning(
-        'Persistence backend for $atSign is now $targetBackend '
+    _logger.warning('Persistence backend for $atSign is now $targetBackend '
         '(previous $currentBackend data retained).');
   }
 
--- a/packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
+++ b/packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
@@ -387,6 +387,26 @@ class EnrollVerbHandler extends AbstractVerbHandler {
         enrollmentValue.apkamKeysExpiryDuration =
             parent.apkamKeysExpiryDuration;
       }
+      // A stated posture may narrow the parent's, never widen it.
+      // `verifyNoEscalation` covers namespaces; TIME is the other axis a
+      // stolen keyfile would want to widen, and this branch is the one
+      // enrollment path with no human in the loop to notice. Zero is the
+      // keystore's "never expires" and a negative value skips the ttl write
+      // altogether, so both are ways of asking for a permanent credential —
+      // against a time-bound parent, neither is honoured.
+      final parentExpiryMs = parent.apkamKeysExpiryDuration.inMilliseconds;
+      final statedExpiryMs =
+          enrollmentValue.apkamKeysExpiryDuration.inMilliseconds;
+      if (statedExpiryMs < 0 ||
+          (parentExpiryMs > 0 &&
+              (statedExpiryMs <= 0 || statedExpiryMs > parentExpiryMs))) {
+        logger.warning(
+            'Self-enrollment under $parentEnrollmentId asked for a key-expiry '
+            'of ${statedExpiryMs}ms against a parent bound to '
+            '${parentExpiryMs}ms; using the parent\'s');
+        enrollmentValue.apkamKeysExpiryDuration =
+            parent.apkamKeysExpiryDuration;
+      }
       // May be absent: a PQ self-enrollment conveys its legacy material
       // client-side, sealed to its own new key package.
       enrollmentValue.encryptedAPKAMSymmetricKey =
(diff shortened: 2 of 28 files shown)


packages/at_secondary_server/lib/src/server/persistence_backend.dart
gkc, line 1: just dart formats here

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
gkc, line 1: This is where almost all of this PR's action is

packages/at_secondary_server/lib/src/verb/handler/enroll_verb_handler.dart
```diff
@@ -520,10 +698,25 @@ class EnrollVerbHandler extends AbstractVerbHandler {
...
+    // enrollment record's own (apkamPublicKey, signingAlgo) — the record
+    // PKAM reads stays the single source. The tagged shape is unmistakable
+    // to a bare-RSA parser: base64-decoding JSON throws, so an old consumer
+    // fails loudly rather than mis-reading the key.
+    final String value;
+    if (signingAlgo == null || signingAlgo == 'rsa2048') {
+      value = apkamPublicKey;
+    } else {
+      value = jsonEncode(
+          {'v': 1, 'signingAlgo': signingAlgo, 'publicKey': apkamPublicKey});
```
gkc, line 717: @cconstab 
The above is the only server-written thing that we are going to have to live with for a long time; everything else is driven entirely by the client side. So - it's got a version field, and in version 1 we have the signingAlgo and the publicKey 

Note that in terms of the client rollout dance, clients don't need to start using mldsa65 signing keys until they want to - and indeed they should not do so until apps which depend on _apsk for signing verification (e.g. noports srvd) are upgraded so they understand legacy bare value format and this versioned json format. The at_client release plan covers this, I'll state it here
- at_client 3.N : all the pq capability but entirely passive - SDK will behave identically to pre-PQ
- at_client 3.N+1 : all the pq capability, plus enrollments will retrofit themselves, nskeys will get published etc - but pq will still not be active for writes. And apkam signing keys will still be RSA (which is fine - remember that for pq we care about eliminating use of RSA for key exchange; signing is fine.)
- at_client 4.0 : pq active for writes by default. apkam default signing key algo flips to mldsa65.

cconstab (approved):

PR #2752, opened 2026-08-16
ci: snapshot the dual-verify DB sets from a stopped container

### LINKED ISSUE #2734: ci: functional_dual_verify snapshots a live container, causing flaky mismatches
`functional_dual_verify` copies each atSign's DBs out of the container with `docker cp` while the atServer is **still running** — `docker stop` is the last step, after verification (`.github/workflows/functional_dual_verify.yaml:103-135`). Hive files and `atsign.db` are copied sequentially, and background jobs (stats notifications, expiry sweep, compaction) keep writing in between, so the two snapshots can disagree.

Symptom: intermittent failure with a small count mismatch on the busiest atSign only.

Seen on [run 30810713050](https://github.com/atsign-foundation/at_server/actions/runs/30810713050) — 39/40 atSigns `IDENTICAL`, one off by a single key:

```
MISMATCH(1) @alice🛠  hive={keystore: 262, commitLog: 70, accessLog: 228, notifications: 74}
                     sqlite={keystore: 263, commitLog: 70, accessLog: 228, notifications: 74}
  keystore: count mismatch (A=262, B=263)
```

commitLog / accessLog / notifications matched exactly, and the job passed on re-run with no code change.

**Fix:** stop or quiesce the container before snapshotting — move the `Stop container` step above `Verify every atSign's Hive and SQLite DB sets are identical`, or copy from a stopped container.

gkc:
Fixes #2734

**- What I did**

- `functional_dual_verify` now enumerates each atSign's databases while the container is up, then **stops the container before `docker cp`s them out** — previously the snapshot was taken from a live container whose atServers were still writing, so the Hive copy and the `atsign.db` copy could disagree (intermittent single-key mismatch on the busiest atSign).
- Dropped `--rm` from the `docker run` so the container survives `docker stop`; the final cleanup step is now `docker rm -f`.
- Gave `tests/at_functional_test/runDualCompare.sh` — the local counterpart of this job — the same treatment; it had the identical live-container snapshot.

**- How I did it**

See commit & diffs.

`docker stop -t 60` is a cap, not a cost. Measured against the VE image with 40 atServers up:

| check | result |
| --- | --- |
| secondaries running before stop | 40 |
| `docker stop -t 60` elapsed | 3s |
| container exit | `exited exit=0` (supervisord graceful, no SIGKILL) |
| `docker cp` from stopped container | ok |
| atServer logs recording `stop() completed, and isRunning is false` | 40 of 40 |

So every store is closed rather than killed mid-write, and the 60s only bounds a slow shutdown.

**- How to verify it**

This workflow runs on the PR — a green `functional_dual_verify` is the verification.

### DIFF
--- a/.github/workflows/functional_dual_verify.yaml
+++ b/.github/workflows/functional_dual_verify.yaml
@@ -71,9 +71,11 @@ jobs:
           context: tools/build_virtual_environment/ve
           tags: at_virtual_env:dual
 
+      # No --rm: the DB snapshot below is taken from the *stopped* container,
+      # so it has to survive `docker stop`. The final step removes it.
       - name: Run docker container in DUAL persistence mode
         run: |
-          docker run -d --rm --name at_virtual_env_dual_cont \
+          docker run -d --name at_virtual_env_dual_cont \
             -e testingMode="true" -e httpsEnabled="true" -e persistenceBackend="dual" \
             -p 6379:6379 -p 25000-25099:25000-25099 -p 64:64 -p 443:443 \
             at_virtual_env:dual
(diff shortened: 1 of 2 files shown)


akafredperry (approved):
lgtm

PR #2753, opened 2026-08-17
fix(at_persistence_secondary_server): correct immutable guard comment

gkc:
**- What I did**

- Corrected the comment above the immutable-preservation branch in `at_metadata_builder.dart`, which claimed "this condition never occurs right now". Since #2751 that holds for two reasons, not one: a LIVE immutable record is refused by the check `update`/`update:meta` share, and an EXPIRED one is deleted before the write, so the builder is handed `existingMetaData == null` and inherits nothing.
- Recorded why the guard is not dead code: the caching, enrollment, otp and config paths call `keyStore.put`/`putMeta` directly, and that refusal is the only immutability check in the tree.
- Fixed a missing interpolation in that branch's warning — it logged the literal text `newAtMetaData?.immutable` instead of the value.
- Fixed a flaky test from #2751: `An EXPIRED immutable record may be created again` created its record with `ttl:1`, so the live-record control raced the expiry — 0/5 locally, green on CI. Now `ttl:1000` + 1500ms, the pair the functional twin already uses.

**- How I did it**

See commit & diffs.

**- How to verify it**

Tests pass. The comment's claim is pinned by `A create over an EXPIRED record inherits nothing from it`; the flaky test goes 0/5 → 5/5.

### DIFF
--- a/packages/at_persistence_secondary_server/lib/src/spec/keystore/at_metadata_builder.dart
+++ b/packages/at_persistence_secondary_server/lib/src/spec/keystore/at_metadata_builder.dart
@@ -42,13 +42,26 @@ class AtMetadataBuilder {
 
     // Ensure that if there was existing metadata and it had immutable == true
     // then that value is preserved regardless of what the new update is trying
-    // to do
-    // Note: this condition never occurs right now but we will leave it here
-    // for safety's sake
+    // to do.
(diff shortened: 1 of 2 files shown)


akafredperry (approved):
looks good

PR #2182, opened 2024-12-13
fix: Sync verb handler add limit param

murali-shris:
Fixes https://github.com/atsign-foundation/at_server/issues/2181
**- What I did**
- Use limit param set from sync verb builder on client side in sync verb handler
**- How I did it**
- read limit param in SyncProgressiveVerb handler and pass it to commitLog.getEntries(..) call
- prepare response method was using SecondaryConfig.pageLimit . Modified this method to use passed syncLimit
- added unit tests
**- How to verify it**
- tests should pass

PR #2769, opened 2026-08-24
fix(at_secondary_server): update:json moves ttl/ttb/ttr only when the request names one

gkc:
**- What I did**

- Made `update:json` move a record's `ttl`, `ttb` or `ttr` only when the request names one — matching the metadata-fragment form of the same request, and the rule the rest of this surface follows.
- The cause: commons `Metadata.fromJson` turns an absent (or explicitly null) `ttl`/`ttb`/`ttr` into `0`, and `Metadata.toJson` always writes the three, so a null makes the round trip as a `0`. An `update:json` that never mentioned an axis therefore arrived asking for `ttl:0` (**clear the record's expiry**), `ttb:0` (available with no delay) or `ttr:0` (**do not cache**). A value-only json update silently dropped a record's expiry and stopped it being cached at the receiver.
- The metadata-fragment form leaves an unmentioned relative null and the retain-merge keeps what the record already holds, so the two encodings of one request stored different things — and only the json one moved an axis it never named. `getUpdateParams` now reads the nulls back off the decoded map the DTO was built from.
- Not a regression from the recent timestamp work: the coercion and the retain-merge's `??=` both predate it (verified at `96243830^1`). That work only documented the limitation, in a test that pinned it — this replaces that test.

**- How I did it**

See commit & diffs. Fixing this in at_commons is the root fix and is a separate PR against at_client_sdk; a version of at_commons that preserves the null makes this a no-op rather than a correction, since it reads the same map to the same answer either way.

**- How to verify it**

Tests pass.

- Both new tests are mutation-proven: disabling the restoration reddens the expiry test (`expiresAt` null where the record's own instant was expected) and the ttr test (`ttr` 0 where 100000 was expected), each quoting its own reason string.
- at_secondary_server 1025/1025, functional 239/239 via `runLocal.sh`, e2e 47/47 via `runLocal.sh`, `dart analyze --fatal-warnings` clean across the workspace.

An `update:json` moves a record's ttl, ttb or ttr only when the request names one, instead of an unmentioned one arriving as an explicit 0 that cleared the record's expiry or stopped it being cached.

### DIFF
--- a/packages/at_secondary_server/CHANGELOG.md
+++ b/packages/at_secondary_server/CHANGELOG.md
@@ -80,6 +80,17 @@
   The keystore canonicalizes keys to lowercase, so two case-variants of
   one update command name the same stored record; keyed on the original
   case they took different mutexes and raced.
+- fix: an `update:json` moves a record's ttl, ttb or ttr only when the
+  request names one, matching the metadata-fragment form of the same
+  request. commons `Metadata.fromJson` turns an absent — or explicitly
+  null — ttl/ttb/ttr into 0 and `Metadata.toJson` always writes the
+  three, so what a json request never mentioned arrived as an explicit
+  `ttl:0` (clear the record's expiry), `ttb:0` (available with no delay)
+  or `ttr:0` (do not cache): a value-only json update silently dropped a
+  record's expiry and stopped it being cached at the receiver. The verb
+  layer now reads the nulls back off the decoded map the DTO was built
+  from. A future at_commons that preserves them makes this a no-op
+  rather than a correction.
 
 # 3.16.2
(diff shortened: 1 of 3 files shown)


gkc:
Root fix for the same defect, in at_commons: atsign-foundation/at_client_sdk#2182 (`Metadata.fromJson` preserves a null ttl/ttb/ttr, so the toJson/fromJson round trip stops being lossy). This PR stands on its own — at_server resolves a published at_commons, and once #2182 ships this becomes a no-op rather than a correction, since it reads the same decoded map to the same answer either way.

gkc:
@cpswan follow-up to the previous PR

cpswan (approved):

PR #2777, opened 2026-08-30
fix(ee): unbreak the EE and VE base images, and build an EE from the working tree

gkc:
Build tooling only — no package code. Three defects, one new script, all independent.

**- What I did**

**1. Repinned the debian base in `ee_base` + `ve_base`.** The pinned digest `eb37f586…` resolves to an OCI index with **zero manifests** — 90 bytes, no entries — so neither image can be built, on any architecture, by anyone. `docker build` reports it as `no match for platform in manifest`, which reads like an arch problem and isn't (`--platform linux/amd64` fails identically). Repinned to what the tag resolves to today, `04634311…`: 16 entries, `linux/amd64` and `linux/arm64/v8` among them.

**2. `createConf.sh`'s default branch substituted a name that isn't there** — it sets `DNS_FQDN`, then seds for `DNS_FQN`. The placeholder survives, so `root_server.url` is the literal string `DNS_FQDN` and, in the default configuration, an atServer cannot reach the atDirectory: no cross-atSign lookup, no `from:` handshake. Only the `DNS_FQDN`-is-set branch has ever produced a working EE, which is why the README's examples all pass it.

**3. A failed compile now fails the EE image build.** The `RUN` chain separated steps with `;`, so only the last exit status reached Docker. With no `.dockerignore`, `COPY . .` then carried the host's own build outputs into `/app` at exactly the paths the compile writes and the second stage reads — and `runLocal.sh` builds those *inside a Linux dart container*, so they're correct-architecture ELF binaries that run fine. Both were present on my host while writing this. Net effect: a compile failure shipped an image running whatever branch was functional-tested last. The ignore list is anchored to the two exact paths, since `**/root` would also exclude `ee_base/contents/atsign/root`, which the image needs.

**4. Added `buildee.sh`** — builds an EE from the working tree (previously impossible; `runee.sh` and the compose file both name the published `:latest`). It labels the image with its source commit and **reads that label back**, then checks the three artefacts carry ELF magic. `runee.sh` takes the tag as an optional third arg, defaulting to the published image, and now passes `DNS_FQDN` explicitly so it's correct against published images still carrying (2).

**CI note.** The bad digest landed 2026-08-25 (`5e0ac64e`, dependabot) and no CI run has ever exercised it: neither workflow that builds these Dockerfiles triggers on `pull_request`, `VE_base`'s weekly cron last ran 17h *before* the bump, and `ee_base` is built only by `vip_rebuild` (cert changes / manual dispatch), so it could have stayed broken indefinitely. A green trunk build is not evidence against this — `push_staging_virtual_env_images` builds `ve/Dockerfile.vip`, whose runtime stage is `FROM atsigncompany/vebase:latest`, a prebuilt image.

**Not fixed here:** `ve_base`'s `RUN` chain still uses `;`, so a failed compile there still goes green — same defect as (3), fixed only for EE. Say the word and I'll fold it in.

**- How I did it**

See commit & diffs.

**- How to verify it**

`tools/build_ephemeral_environment/buildee.sh`, green from the merged tree — revision label matches HEAD, all three artefacts ELF. That covers (1), (3) and (4): the image cannot build at all without the repin. For (2), `runee.sh <name> <base-port> at_ephemeral:local` and check a created atServer's `root_server.url` is a hostname, not the literal `DNS_FQDN`.

Fixed the EE and VE base images, which could not be built at all — the pinned debian digest resolved to an empty manifest index. Fixed the EE's default `DNS_FQDN` configuration, which left every atServer unable to reach the atDirectory. A failed compile now fails the EE image build instead of silently shipping the host's own build outputs. Added `buildee.sh` for building and verifying an EE image from a working tree.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,19 @@
+# Build outputs that `docker build` would otherwise copy into the image's
+# build stage at exactly the paths the compile step writes.
+#
+# `tests/at_functional_test/runLocal.sh` compiles the atDirectory and atServer
+# on the host (inside a Linux dart container, so they are Linux ELF binaries
+# indistinguishable from the ones the image builds) and leaves them here. Both
+# EE and VE Dockerfiles `COPY . .` into /app and then compile over them, so a
+# compile that fails leaves the host's build in place and the image ships it —
+# a stale atServer that looks entirely healthy.
+#
+# Anchored to the two exact paths: a `**/root` pattern would also exclude
+# tools/build_ephemeral_environment/ee_base/contents/atsign/root, which the
+# image genuinely needs.
+/packages/at_root_server/root
+/packages/at_secondary_server/secondary
+
+# Host tool state. Every path inside is absolute and belongs to the host, so it
+# is meaningless in the image; `dart pub get` regenerates what it needs.
+**/.dart_tool/
--- a/tools/build_ephemeral_environment/buildee.sh
+++ b/tools/build_ephemeral_environment/buildee.sh
@@ -8,9 +8,16 @@
 # the atServer or the atDirectory needs an image built from the branch they are
 # on, and needs to be able to prove that is what they got.
 #
-# Usage: buildee.sh [-t <tag>] [-q]
+# Usage: buildee.sh [-t <tag>] [-p <platform>] [-q]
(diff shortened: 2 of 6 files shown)


cconstab (approved):
Ooops...

Issue #2780, opened 2026-08-31
chore(at_server_spec): revert the local enroll:infons syntax override once at_commons publishes it


gkc:
`at_server_spec`'s `Enroll` verb currently builds its own syntax string instead of returning `VerbSyntax.enroll` unchanged, the way every other verb in the package does. This issue is to undo that once it is no longer needed.

## Why the override exists

at_commons owns the enroll operation alternation, and the published versions do not list `infons`. Without a local addition, `enroll:infons:<namespace>` is rejected as invalid syntax before it ever reaches a handler.

## What it does

`packages/at_server_spec/lib/src/verb/enroll.dart` INSERTS into at_commons' pattern rather than copying it:

```dart
const insertionPoint = '(request|';
final String upstream = VerbSyntax.enroll;
if (upstream.contains('infons')) {
  return upstream;
}
return upstream.replaceFirst(insertionPoint, '${insertionPoint}infons|');
```

Inserting rather than copying means every other upstream change to the enroll syntax still reaches this server; only the one addition is local.

It never throws. `syntax()` is read per command, so an exception here would take out **every** enroll operation — `enroll:request` included, which is unauthenticated onboarding — rather than just the one operation it is about. Both ways the insertion can fail to apply are benign and are caught by tests instead:

- **at_commons already lists `infons`** — the upstream pattern is returned whole, and a running server keeps working.
- **the anchor has moved** — `replaceFirst` is a no-op, `enroll:infons` is rejected as a syntax error, and every other enroll operation still works.

## The signal to do this work

`packages/at_secondary_server/test/enroll_verb_syntax_test.dart` ends with a tripwire that asserts `VerbSyntax.enroll` does **not** contain `infons`. It fails deliberately the day a published at_commons defines the operation, and its failure message names the removal steps. That test going red is what should bring someone here.

## The change

1. Delete `_syntaxWithInfons` and `_buildSyntax()` from `Enroll`.
2. Make `syntax()` return `VerbSyntax.enroll`.
3. Delete the tripwire test, and keep the rest of `enroll_verb_syntax_test.dart` — the other tests assert that `enroll:infons:wavi` parses, that it reuses at_commons' existing `listNamespace` capture group, that it is exactly as strict as `listns` and `list`, and that `enroll:bogusop:wavi` does not match. Those stay useful against the upstream pattern.
4. Raise the `at_commons` constraint floor in the same commit as the first use, per the workspace rule.

## Context

[...]

PR #2781, opened 2026-08-31
feat(at_secondary_server): retrofit successors, revocation cascades, and a revocation history

gkc:
## Why

**An app must be able to replace its own APKAM key material without a human step.** Today every new credential needs an OTP and an approver on another device. That is right for admitting a new app, and wrong for an app that is already trusted and needs new keys — a PQ signing-algorithm migration being the case that forces it. A fleet cannot migrate if every device needs someone to approve it.

Letting a credential mint its own replacement opens four holes, and the rest of this PR is those four being closed. Each is a goal in its own right.

**A replacement must not outlive its purpose.** If the enrollment it replaced keeps working forever, self-enrollment doubles the credentials on the atSign every time it runs. But it cannot be deleted on the spot either: sibling clones of one keyfile share an identity, and each still has to be able to upgrade.

**Self-enrollment must not become a way around revocation.** A stolen keyfile can mint a successor before the theft is noticed. If revoking the stolen enrollment leaves that successor approved, revocation is defeated by the very feature that created it.

**An atSign must never be left unable to approve a replacement.** Revocation that can strand an atSign is worse than no revocation: recovery requires a fully privileged enrollment, and if the last one goes there is no way back.

**A client must be able to find out that a peer was revoked.** A revoked holder leaves a namespace's roster, but nothing tells the other holders to re-fetch.

---

## What

**Self-enrollment — an APKAM connection mints its own successor**
- `enroll_verb_handler.dart`: APKAM branch of `enroll:request` auto-approves, no OTP, no approver.
- `enroll_verb_handler.dart`: `verifyNoEscalation` then `requireGrantsMatchPredecessor` — a successor carries its predecessor's grants exactly and may not choose its own.
- `enroll_datastore_value.dart` + `.g.dart`: `parentEnrollmentId`, the edge the cascade walks. The `.g.dart` is hand-maintained.

**Capping the enrollment that was replaced**
- `pkam_verb_handler.dart`: the cap is armed by the successor's FIRST authentication, not when the server stores the record.
- `enrollment_manager.dart`: `armRetrofitCapOnFirstAuth`, `capEnrollmentExpiry`, `retrofitCapTtlMillis`, `RetrofitCapOutcome`.
- `enrollment_manager.dart`: `_clearCapStamp` — a DECLINED cap takes its stamp back, because a decline is a judgement about state that can change.
- `enroll_datastore_value.dart`: `predecessorCapArmedAt` — a timestamp, not a flag, because the cap re-arms per sibling.
- `enrollment_manager.dart`: `declinedAtGeneration`, an in-process memo so a standing decline stops re-walking the keystore every authentication.
- Every write on this path asserts the stored `expiresAt` back, or the metadata builder restarts the credential's expiry clock.

**Revocation cascade**
- `enrollment_manager.dart`: `descendantsOf` climbs upward and fetches each link BY KEY, closing the window between a record expiring and its removal. ⚠️ Only that window: a scheduled `deleteExpiredKeys()` sweep then removes the record and severs the chain for good — see #2782, which is the remaining half and needs ancestry that outlives the record.
- `enrollment_manager.dart`: `_predecessorIdOf` — absent and undecodable end a chain; a STORE FAULT propagates rather than being read as "no predecessor".
- `enrollment_manager.dart`: `revokeAll` is two-phase; `movePerEnrollmentDataFor` makes one keystore pass for the whole cascade instead of one per record.
- `enroll_verb_handler.dart`: the cascade is written BEFORE the target, so a retry after a part-way failure can finish the job.
- `enroll_verb_handler.dart`: connections are dropped for the INTENDED cascade, not the subset this call flipped.
- `enroll_verb_handler.dart`: `_refuseIfPredecessorNotApproved` on `approve` and `unrevoke`.
- `pkam_verb_handler.dart`, `enroll_verb_handler.dart`: the enrollment id is folded to lower case where it is read off the wire.

**Refusals that stop an atSign stranding itself**
- `enroll_verb_handler.dart`: a caller inside its own cascade is refused.
- `enroll_verb_handler.dart`: revoking a fully privileged enrollment may not leave the atSign without one — asked for every such revoke, not only a self-revoke.
- `enrollment_manager.dart`: `hasRootEnrollmentAliveAfter(Set excluding, DateTime deadline)` — the set excludes the cascade, which is still `approved` while the check runs.
- `enroll_datastore_value.dart`, `abstract_verb_handler.dart`: `isRootEnrollment` moves to the value class as a getter.

**Revocation history, and discovering it**
- `enrollment_revocation_event.dart` (new): one durable record per revocation state change — enrollment, moment, grants held, who issued it, whose cascade took it. An un-revoke writes a counter-event; nothing is rewritten.
- `enrollment_manager.dart`: `recordRevocationEvents`, `revocationEvents`, `buildRevocationEventKey`, `revocationEventsRegex`. Key shape deliberately avoids `new.enrollments`, whose enumeration regex is an unanchored substring.
- `enrollment_manager.dart`: `lastRevocationForNamespace` derives from the events and nets out un-revocations, so it can move backwards.
- `enrollment_manager.dart`: `accessInNamespaces` — the namespace-match rule, askable of a bare grants map.
- `enroll_verb_handler.dart`: `_fetchNamespaceInfo` (`enroll:infons`) and `_requireNamespaceAccess`, one gate shared with `listns`.
- `enroll_verb_handler.dart`: a revoke records its event BEFORE its write, an un-revoke AFTER, so a crash errs towards reporting the namespace as revoked.
- `enroll_datastore_value.dart` + `.g.dart`: **`revokedAt` removed** — it lived on a record reaped on the credential's own expiry schedule, taking the answer with it.
- `at_server_spec/lib/src/verb/enroll.dart`: ⚠️ TEMPORARY — inserts `infons` into at_commons' operation alternation rather than copying it, and never throws. Tracked by #2780.

**`enroll:delete` authorisation**
- `enroll_verb_handler.dart`: the same gate `enroll:fetch` applies — `__manage` plus access to every namespace the target holds — with the same exemptions (self, and a connection carrying no enrollment id). Asked before the status checks, so a refused caller does not learn the target's state.
- `enroll_verb_handler.dart`: a target holding NO namespaces is refused rather than passed vacuously. Such a record is reachable — an `enroll:request` on a LEGACY-PKAM connection writes one. ⚠️ Every other per-namespace loop (approve/deny/revoke/unrevoke, and fetch) still passes it vacuously; only this gate refuses.

**Config**
- `config.yaml`, `at_secondary_config.dart`: `preserveFirstEnrollmentOnRetrofit` **removed**, `apkamSelfEnrollmentGraceHours` (default 720) added. ⚠️ Deployment-visible.

⚠️ The delete gate NARROWS a shipped capability: an enrollment holding `__manage` but not the target's namespaces could delete it before and cannot now. `at_onboarding_cli`'s delete command takes an arbitrary `--enrollment-id`, so it is affected whenever it runs as an enrollment rather than from an owner keyfile.

---

## How to verify it

Tests pass: 1190 unit, 253 `at_functional_test`, 47 `at_end2end_test`. `dart analyze --fatal-warnings` clean.

`tests/at_functional_test/test/apkam_self_enrollment_test.dart` goes 23 → 37 tests. Each pins one behaviour above against a real atServer over the wire.

**Added — self-enrollment and grants**
- `a request naming a subset of a wildcard predecessor is refused`
- `a wildcard predecessor is inherited verbatim when nothing is named`
- `an empty namespace set inherits, rather than being refused` — behaviour CHANGED; previously refused

**Added — the cap**
- `storing a successor does not cap the predecessor`
- `the cap is armed by the successor's first authentication, and re-arms for each sibling`
- `a successor re-authenticating does not push the deadline out again`
- `a CAPPED predecessor still authenticates`
- `the successor records WHEN it armed, under a stable at-rest name`
- `revoking an enrollment does not restart its expiry either`
- `an enrollment cannot postpone its own retirement with enroll:update`

**Added — cascade and revocation**
- `a successor stops authenticating when the enrollment it replaced is revoked`
- `the cascade reaches a successor's successor, not just the first`
- `un-revoking a descendant is refused while what it replaced stays revoked`
- `a connection already open on a cascaded enrollment is dropped`
- `an ordinary revoke response carries no cascade field`

**Added — the revocation history over the wire**
- `the enrollment record carries no revocation time; the history does`
- `enroll:infons reports the namespace's last revocation over the wire` — also pins that an un-revoke returns the namespace to its prior answer
- `enroll:infons is refused on a connection that is not APKAM`

**Amended — vocabulary only, same behaviour** (parent/child → predecessor/successor): `an approved predecessor is replaced by a successor that can authenticate`, `the successor records what it replaced, which is what the cascade walks`, `a self-enrolled successor needs no encrypted APKAM symmetric key`, `a namespace the predecessor does not hold is refused`, `a wildcard predecessor cannot mint __manage`, `a successor inherits the predecessor's key-expiry posture`, `a successor may state a SHORTER expiry than its predecessor`, `a successor may not state an expiry that outlives its predecessor`, `a successor may not state "never expires" against a bounded predecessor`.

**Unit-side, for reviewers looking for the format contracts.** Three files are raw-literal pins rather than round trips, so their diffs show what moved on disk or on the wire without opening a writer: `enroll_revocation_event_test.dart`, `enroll_data_store_value_test.dart`, `enroll_verb_syntax_test.dart`. `enrollment_fault_injection_test.dart` is new and covers three guards no Hive-backed fixture can reach — a store that throws on a key the enumeration just returned, and one that fails mid-walk.

## Description for the changelog

A retrofit replaces its predecessor rather than descending from it, carrying exactly its grants; the predecessor is capped rather than deleted, the cap armed by the successor's first authentication and re-armed per sibling. Revoking an enrollment revokes everything that replaced it, transitively, and drops those connections. Three refusals prevent an atSign being left unable to approve a replacement. Revocations and un-revocations are recorded as durable events that outlive the enrollment records, and `enroll:infons:<namespace>` reports the last revocation affecting a namespace. `enroll:delete` now requires `__manage` and access to every namespace the target holds, matching `enroll:fetch`. Config: `preserveFirstEnrollmentOnRetrofit` is replaced by `apkamSelfEnrollmentGraceHours` (default 720).


cconstab (dismissed):
Migrating this to the other SDK's will need some labs!

Xlin123 (approved):

PR #2804, opened 2026-09-09
fix(at_secondary_server): plookup reports a peer's encryption public key as ttr -1

### LINKED ISSUE #2693: cache: encryption public key evicted after 24h — stray colon makes the ttr=-1 branch dead
## Summary

In `AtCacheManager.remoteLookUp` (`packages/at_secondary_server/lib/src/caching/cache_manager.dart`), the branch meant to cache another atSign's **encryption public key indefinitely** never fires, because its guard has a stray colon. As a result the cached public key is stamped with a 24-hour TTL and is evicted a day later, instead of being retained indefinitely.

## Detail

When a remote value arrives with `ttr == 0 || ttr == null` ("do not cache"), there is a carve-out for public keys:

```dart
} else if (atData.metaData!.ttr == 0 || atData.metaData!.ttr == null) {
  shouldCache = false;
  if (cachedKeyName.startsWith('cached:public:')) {
    if (cachedKeyName.startsWith('cached:public:publickey:@')) {   // <-- stray colon
      shouldCache = true;
      atData.metaData!.ttr = -1;                                   // intended: cache forever
    } else {
      shouldCache = true;
      atData.metaData!.ttl = 24 * 60 * 60 * 1000;                 // 24h backwards-compat fallback
    }
  }
}
```

An encryption public key is cached under `cached:public:publickey@<atSign>` — there is **no colon** between `publickey` and `@`. Every other reference in this file uses the colon-free form (e.g. `!cachedKeyName.startsWith('cached:public:publickey@')` in the invalidation guards and in `put`/`putCachedPublicKey`). The guard `startsWith('cached:public:publickey:@')` (with the extra `:`) therefore never matches, the `ttr = -1` branch is dead, and the encryption public key falls into the `else` branch, which stamps `ttl = 24h`.

`put()` subsequently routes the key to `putCachedPublicKey()` (matched by the colon-free `cached:public:publickey@`), which sets `ttr = -1` — but it does **not** clear the `ttl = 24h` that the `else` branch already stamped on the same `AtData`. The stored cached public key ends up carrying `ttr = -1` **and** `ttl = 24h`, so it expires after 24 hours despite the "cache indefinitely" intent, forcing a re-fetch each day (and a lookup failure if the origin is unreachable at re-fetch time).

## Fix

Remove the stray colon so the guard matches the real cached key form:

```dart
if (cachedKeyName.startsWith('cached:public:publickey@')) {
```

Then the `ttr = -1` (no ttl) branch fires for the encryption public key, matching the colon-free form used throughout the rest of the file.

gkc:
Fixes #2693.

**- What I did**

- `plookup:all:publickey@<atSign>` and `plookup:meta:publickey@<atSign>` now report `ttr` -1 with no `ttl`, instead of `ttl` 86400000 with no `ttr`. A stray colon made the carve-out guard unmatchable, so a peer's encryption public key took the 24-hour fallback meant for other public values.
- Nothing else changes. Storage was already `ttr` -1 with no expiry. `lookup` is unaffected: it builds a `cached:@<atSign>:` name, which the carve-out cannot match.

**- How I did it**

See commit & diffs

**- How to verify it**

Tests pass. Unit 1519, functional 255.

fix: `plookup:all` and `plookup:meta` report a peer's encryption public key as `ttr` -1 rather than a 24-hour `ttl`.

### DIFF
--- a/packages/at_secondary_server/lib/src/caching/cache_manager.dart
+++ b/packages/at_secondary_server/lib/src/caching/cache_manager.dart
@@ -200,7 +200,7 @@ class AtCacheManager {
           shouldCache = false;
           if (cachedKeyName.startsWith('cached:public:')) {
             // HOWEVER: publickey@atSign should be cached with ttr of -1 (cache indefinitely)
-            if (cachedKeyName.startsWith('cached:public:publickey:@')) {
+            if (cachedKeyName.startsWith('cached:public:publickey@')) {
               shouldCache = true;
               atData.metaData!.ttr = -1;
             } else {
--- a/packages/at_secondary_server/CHANGELOG.md
+++ b/packages/at_secondary_server/CHANGELOG.md
@@ -1,4 +1,13 @@
 # 3.16.5
+- fix: `plookup:all:publickey@<atSign>` and `plookup:meta:publickey@<atSign>`
+  report `ttr` -1 with no `ttl`, instead of `ttl` 86400000 with no `ttr`. The
+  carve-out in `AtCacheManager.remoteLookUp` guarded on
+  `cached:public:publickey:@`, a form no key has, so a peer's encryption
+  public key took the 24-hour fallback meant for other public values.
+
+  Storage is unchanged and was never wrong. `lookup` is unaffected: it builds
+  a `cached:@<atSign>:` name, which the carve-out cannot match.
+
 - feat: `enroll:fetch` returns `metadata`, the opaque map the enrolling client
   supplied, verbatim. A caller that may fetch an enrollment can now read its
   key package by id instead of calling `enroll:list` and filtering a whole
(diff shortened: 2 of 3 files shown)


packages/at_secondary_server/lib/src/caching/cache_manager.dart
```diff
@@ -200,7 +200,7 @@ class AtCacheManager {
           shouldCache = false;
           if (cachedKeyName.startsWith('cached:public:')) {
             // HOWEVER: publickey@atSign should be cached with ttr of -1 (cache indefinitely)
-            if (cachedKeyName.startsWith('cached:public:publickey:@')) {
+            if (cachedKeyName.startsWith('cached:public:publickey@')) {
```
gkc, line 203: Note on the above: the public:publickey will shortly be obsolete and unimportant to track but nonetheless this is a code correction worth making.

cpswan (approved):

packages/at_secondary_server/lib/src/caching/cache_manager.dart
gkc: Obsolete but will still be sticking around for quite a long time to ensure legacy code doesn't break

